PR #12, opened 2014-12-03
Add cons operators

lionelgit:
This adds cons operators to create lists easily.


hadley:
I think this would be useful, so can you please submit a new PR? (And close this one)

I wonder if we should define some of our own is functions:

```{r}
is_base_list <- function(x) !is.object(x) && is.list(x)
is_base_numeric <- function(x) !is.object(x) && is.double(x)
is_base_integer <- function(x) !is.object(x) && is.integer(x)
is_base_atomic <- function(x) !is.object(x) && is.atomic(x) && !is.null(x)
is_base_vector <- function(x) is_base_atomic(x) || is_base_list(x)
```

etc. The idea of the base prefix is to make it clear we're testing specifically base (non-S3) vectors. That might not be the best name

lionelgit:
Good idea. Maybe the prefix `bare`?

```
is_bare_list <- function(x) !is.object(x) && is.list(x)
is_bare_numeric <- function(x) !is.object(x) && is.double(x)
is_bare_integer <- function(x) !is.object(x) && is.integer(x)
is_bare_atomic <- function(x) !is.object(x) && is.atomic(x) && !is.null(x)
is_bare_vector <- function(x) is_base_atomic(x) || is_base_list(x)
```

lionelgit:
We could also have
```
is_atomic <- function(x) is.atomic(x) && !is.null(x)
```
cf. your recent tweet.

hadley:
So maybe `is_atomic()` and `is_vector()` and `is_bare_*()`?

lionelgit:
I feel like it's inconsistent to have `is_bare_numeric()` check for floats while `is.numeric()` checks for integers and floats alike. So maybe also have `is_numeric <- is.double` for semantic consistency in projects using lowliner?

Another option:
```
is_bare_double <- function(x) !is.object(x) && is.double(x)
is_bare_numeric <- function(x) !is.object(x) && is.numeric(x)
```

I would also be tempted to add underscore synonyms of all type predicates but that would be gratuitous :-)

lionelgit:
Actually
```
is.vector(NULL)
[1] FALSE
```
so `is_vector` is not needed.

PR #24, opened 2014-12-15
Implement `cross()`

lionelgit:
This is a trimmed down version of `expand.grid()` that takes a list as input and return a list.

Example:

```
data <- list(                                
  id = c("John", "Jane"),                    
  greeting = c("Hello.", "Bonjour."),        
  sep = c("! ", "... ")                      
)                                            
args <- cross(data)                          
                                             
out <- vector("list", length(args[[1]]))     
for (i in seq_along(out))                    
  out[[i]] <- map(args, i) %>% splat(paste)()
out

## [[1]]
## [1] "John! Hello."
## 
## [[2]]
## [1] "Jane! Hello."
## 
## [[3]]
## [1] "John! Bonjour."
## 
## [[4]]
## [1] "Jane! Bonjour."
## 
## [[5]]
## [1] "John... Hello."
## 
## [[6]]
## [1] "Jane... Hello."
## 
## [[7]]
## [1] "John... Bonjour."
## 
## [[8]]
## [1] "Jane... Bonjour."
```


hadley:
It would return a special object that then other functions would need to understand. It's something I've been thinking about a bit in the content of non-equi joins for dplyr. Often the first thing you do after generating a Cartesian product is filter it in some way, so it's nice to be able to save some time/memory.

lionelgit:
On the other hand, since R usually makes shallow copies, I'm not sure a lazy version would be more efficient. There would probably be some overhead.

In what circumstances would `cross()` trigger deep copies?

hadley:
I'm overthinking this. How about something like `cross(x, y, filter)` so you could do `cross(1:10, 1:10, filter(x, y) x < y)`. The advantage of this over `cross(1:10, 1:10) %>% filter(~ .$x > .$y)` is when the full set of combinations is very large (maybe larger than memory) and you only want a small set of them.

We also need to think about the arguments - should it be `cross(x, y)`, `cross(list(x, y))` or `cross(...)`. Maybe analogous to map we should have `cross()`, `cross3()` and `cross_n()`?  Or are you imagining that cross would only work with data frames?

lionelgit:
About data frames, I should probably make `cross()` return a data frame when given one as input to be consistent with the other functions.

I was thinking `cross(list(...))` because this works well with pipes. In the use case for which I wrote this function, I built a list of things (not a data frame) and then piped it into `cross()`.

But indeed we could have
```{r}
cross(.x, .y, filter = NULL)
cross3(.x, .y, .z, filter = NULL)
cross_n(..., filter = NULL)
cross_list(.l, filter = NULL)
```

I think it would be good to also have both `map_n()` and `map_list()` as suggested in #29.

Maybe the `list` suffix is not great since these functions also accept data frames. A data frame is also a list but it kind of feels wrong to use a function named `do_list()` and get a data frame. That's probably a minor issue though.

By the way, a cosmetic question about arguments in lowliner. Should all arguments begin with a dot in all lowliner functions, or only in those which accept a `...` argument? Or should we have a mix of both as in `unzip()`? Following the mixed logic, the arguments to an hypothetical `cross3()` would be as in the code block above. Another option is to be consistent and have dots everywhere so that we don't have to think about which arguments are dotted and which are not when we call a lowliner function. I'm not sure which is best.

R/cross.R
```diff
@@ -0,0 +1,160 @@
...
+    if (.wide) {
+      for (j in seq_along(out)) {
+        out[[j]][[i]] <- x[seq][j]
+      }
+    } else {
+      out[[i]] <- x[seq]
+    }
+  }
+
+  out %>% output_hook(.l)
```
hadley: I think this should only be applied if `!.wide` - it's unlikely to give a useful output if `.wide` is TRUE

R/cross.R
```diff
@@ -0,0 +1,160 @@
...
+#' functio! that takes the same number of arguments as the number of
+#' crossed objects (2 for \code{cross()}, 3 for \code{cross3()},
+#' \code{length(.l)} for \code{cross_n()}) and returns \code{TRUE} or
+#' \code{FALSE}. The combinations where the predicate function returns
+#' \code{TRUE} will be removed from the result.
+#'
+#' @seealso expand.grid
+#' @param .x,.y,.z Lists or atomic vectors.
+#' @param .l A list of lists or atomic vectors.
+#' @param .wide If \code{TRUE}, returns a list of unique combinations.
```
hadley: I think this needs a bit more explanation, and I'm not sure if `.wide` is the best name. I almost wonder if the default should differ based on `.x` being a data frame.

R/cross.R
```diff
@@ -0,0 +1,160 @@
...
+#' If \code{FALSE}, returns a list of the same size as the number of
+#' arguments (2 for \code{cross()}, 3 for \code{cross3()},
+#' \code{length(.l)} for \code{cross_n()}).
+#' @param .filter A predicate function
+#' @return A list or a data frame.
+#' @export
+#' @examples
+#' # We build all combinations of names, greetings and separators from our
+#' # list of data and pass each one to paste()
+#' data <- list(
```
hadley: If `data` is a data.frame, you expect a data frame with three columns, `id`, `greeting` and `sep`, and then you'd work with it with `mutate()` (or `transform()`). It might be worth showing both variations.

hadley:
I think the argument names should be consistent regardless of whether `...` is present or not, i.e. they should always use the `.` prefix (and `unzip()` argument names should be tweaked).

R/cross.R
hadley: The more I think about this the more I like it. It would make `cross()` a bit different to other functions, but it would be more likely to generate the correct output by default.

hadley:
BTW I'm not that interested in performance at the moment. Once we get the API right, we can optimise, probably by writing a little Rcpp.

R/cross.R
lionelgit: yes it makes sense to have the most useful format as a default for each type. I don't like `.wide` either but I couldn't find a better name. I thought about `.transpose` but it doesn't describe the kind of output we get.

hadley: Another option would be `.format = c("zip", "unzip")` maybe? But I think I'd have to think about which one is which every time I used it.

lionelgit:
So you'd prefer that I simplify the code for now?

Using the row-by-row implementation in all cases (so we can filter) and unzip if it's needed?

hadley:
Yeah, the simpler the better for now, and use the row-by-row impl.

lionelgit:
The remaining travis complaints are the `map_d()` leftovers and because I use `smash()` and `map_call()` in the examples. Good to go.

lionelgit:
Hard coded `.wide = FALSE` for data frames because `TRUE` would result in an error due to lack of names

lionelgit:
It may make sense to also hard code `.wide = TRUE` for list. This way we get rid of this confusing `.wide` argument and of the S3 dispatching, and we can still show in the doc how to get an output similar to `expand.grid()` by unzipping the product.

hadley:
That sounds good to me

lionelgit:
One side effect of using `enlist()` in the splatting functions is that using them with one-row data frames does not work (I had to coerce with `as.list()` in the examples for `cross()`)

hadley:
That's definitely a big downside

lionelgit:
Ok, now they use `c()` and the data frame examples in `cross()` look pretty again

hadley:
Thanks!

PR #26, opened 2014-12-21
Implement probe()

lionelgit:
`probe(x, predicate)` is a shortcut meant to be the equivalent to `map_v(x, predicate, .type = logical(1))`.

The need for such a shortcut is somewhat compensated by all lowliner functions taking a predicate as argument: `keep()`, `map_if()`, ... However there are still situations where it is needed to construct a logical vector.

For example, if we use `map_if()` twice to selectively modify the list when a predicate is TRUE and an alternative modification when it's FALSE, we need to record such a logical vector because the contents of the list are altered during the first `map_if()` call. Example:
```
l <- list(a = 1, b = "2", c = "3")

is_character <- probe(l, is.character)
l <- l %>%
  map_if(is_character, as.numeric) %>%
  map_if(!is_character, as.character)
```

I think `probe()` would be a nice little touch of sugar and make the code a bit more expressive in these situations.


hadley:
I think what you're actually asking for is to rename `find_selection()` and export it. I think that's reasonable, so can you please replace the existing uses of `find_selection()` with probe?

lionelgit:
Exactly. At the time I wanted two different functions because I was hoping that lowliner functions would gain a `.depth` argument, but now we have `at_depth()`.

R/probe.R
```diff
@@ -0,0 +1,30 @@
...
+#' # we first need to record a logical vector. Using a predicate function in both
+#' # map_if() calls would not work since they change the list contents.
+#' l <- list(a = 1, b = "2", c = "3")
+#' 
+#' is_character <- probe(l, is.character)
+#' l <- l %>%
+#'   map_if(is_character, as.numeric) %>%
+#'   map_if(!is_character, as.character)
+probe <- function(.x, .p, ...) {
+  find_selection(.x, .p, ...)
```
hadley: Could you please inline `find_selection`? (and then replace existing uses with `probe()`)

lionelgit:
Travis fails with `rbenoulli.Rd` thing. `check()` passes on my machine.

hadley:
Thanks!

PR #27, opened 2014-12-26
Make each() compatible with formulaes, and implement each2() and each3()

lionelgit:
This patch allows `each()` to handle lambda formulaes, and implements multiple arguments versions `each2()` and `each3()`.

The new `recycle_args()` helper function is used by `each*()` to recycle lists of length 1. This is currently not consistent with `map*()`. So I could adjust the helper to use R's regular recycling conventions, but it may be better to modify `map2()` and `map3()` to also use `recycle_args()`. This way, lowliner's recycling rules would be consistent with dplyr's, which would be best I think.


lionelgit:
I added `each_n()`, `map_n()` and modified `map2()` and `map3()` to use `recycle_args()`.

Tell me if you prefer the old recycling behaviour, I'll adjust the PR.

hadley:
Looks good, thanks!

PR #30, opened 2015-01-08
Let splat() specify default arguments for .f

lionelgit:
This lets `splat()` accept additional arguments that are fed to `.f` as defaults.

When we pipe a list into a splatted function, it is often necessary to take care of additional arguments. That typically means adjusting the main list with additional arguments by using a temporary variable or `append()`ing  them in the middle of the pipe chain. Being able to specify these arguments directly in `splat()` hopefully leads to clearer code because the main arguments and the side arguments are specified separately (the former in the pipe chain and the latter in `splat()`).


lionelgit:
In #33 I mentioned the idea of having
> differing function signatures as a function of the type of the first argument

Along these lines, it could also make sense to have a dual behavior for `splat()` such it would behave as a functional or as a factory function depending on the type of the first argument.
```{r}
splat(fun)(x) # factory function
splat(x, fun) # functional (equivalent to do.call with x as first arg)
```

I don't like factory functions so much in pipes because the parentheses are syntax heavy. With the dual behavior we could write
```{r}
x %>%
  map(fun1) %>%
  splat(fun2)
```
instead of
```{r}
x %>%
  map(fun1) %>%
  splat(fun2)()
```

R/splat.R
```diff
@@ -1,11 +1,14 @@
...
-#' This is useful when you want to pass a function a row of data frame or
-#' list, and don't want to manually pull it apart in your function.
+#' arguments, it takes a single named list which will be interpreted
+#' as its arguments. This is useful when you want to pass a function
+#' a row of data frame or list, and don't want to manually pull it
+#' apart in your function. You can also specify default arguments for
+#' \code{.f} in the call to \code{splat()}.
 #'
 #' @param .f Function to splat.
+#' @param ... Default arguments for \code{.f}.
```
hadley: I'd add: "These will be evaluated once when `splat()` is called".

lionelgit:
Thanks for your reviews Hadley.

One last thing, do you think the point above would be a good idea? That is, making `splat()` behave as a functional when it's given a list, or a factory function when it's given a function. I think it makes sense and would result in prettier pipes.

hadley:
Again, I think that's like currying, which I don't really like for R. Functions are easier to understand when they always return the same type of thing.

lionelgit:
In this case I thought of it more as S3 polymorphism than currying (splatting a list vs splatting a function).

hadley:
Hmmmm, could you share a draft implementation?

lionelgit:
Something like

```{r}
splat <- function(x, ...) {
  UseMethod("splat")
}

#' Splat an object as arguments to a function
#'
#' This calls \code{.f} with the elements of \code{.x} as
#' arguments. This is basically a pipable version of \code{do.call}.
splat.default <- function(.x, .f, ...) {
  defaults <- list(...)
  do.call(".f", c(.x, defaults))
}

#' Splat a function
#'
#' This modifies the function \code{.f} to make it take a list of
#' arguments as argument.
splat.function <- function(.f, ...) {
  force(.f)
  defaults <- list(...)
  function(args = list()) {
    do.call(".f", c(args, defaults))
  }
}
```

```{r}
x <- list(c(1:100, NA, 1000))
splatted_mean <- splat(mean, na.rm = TRUE, trim = 0.9)

x %>% splat(mean, na.rm = TRUE, trim = 0.9)
x %>% splat(mean, na.rm = TRUE, trim = 0.9)()
x %>% splatted_mean()
```

lionelgit:
Not sure if we should use `enlist()` instead of `c()` in the `do.call()` calls.

With `enlist()`,
```
c(1:100, NA, 1000) %>% splat(mean)
```
would work as expected.

But:
```
c("01", "01", "2001") %>% splat(paste, sep = "-")
## [1] "01"   "01"   "2001"
```
```
list("01", "01", "2001") %>% splat(paste, sep = "-")
## [1] "01-01-2001"
```
With `c()` they would both work.

hadley:
That looks ok, but the arguments of the methods must match the generic, so they'd all need to use `.x` as the first argument. Documenting that is going to get nasty. How about `splat()` and `splat_map()` (or `map_splat()`?)

lionelgit:
`map_splat()` would work.

We could also `splat()` a list and `smash()` a function?

hadley:
I love the idea of a function called `smash()`, but I suspect it would be more confusing.

lionelgit:
the version for lists could simply be called `do_call()`
Or `map_call()`

lionelgit:
```
list %>%
  map_call("fun")
```
works pretty well I think

hadley:
Oh I like that

lionelgit:
Can I rename `splat()` to `smash()`?

The idea of `splat`ting arguments to a function suggests that `splat()` takes a list of arguments, not a function.

And this way we'll get to smash functions in R anytime these damn things won't take a list :-)

hadley:
Do it!

lionelgit:
`check()` complains because `map_d()` is still used in examples in some files of this branch (this is fixed in master) and because it does not find `enlist()`.

So after #36 is merged this is safe to merge.

R/smash.R
```diff
@@ -0,0 +1,58 @@
...
+#' While `smash()` wraps a function in \code{\link{do.call}()},
+#' `map_call()` is directly equivalent to \code{do.call()} except that
+#' it takes a list as first argument instead of a function. This makes
+#' `map_call()` pipable.
+#' @param .x A list or a vector. Vectors are automatically wrapped in
+#'   a list.
+#' @param .f A function or the name of a function to call with the
+#'   elements of \code{.x} as arguments.
+#' @param ... Additional arguments passed on to \code{.f}.
+#' @seealso smash do.call
```
hadley: These need to be `\code{\link{smash}}` so they turn into links in the docs

hadley:
Looks good - just one tiny niggle left in the docs. 

You can make the travis failure go away by merging master back on to this branch (but it's not a big deal either way)

lionelgit:
I encapsulated the dots in a list so that the result of `c(.x, list(...))` is always a list. This allows `map_call()` to work with atomic vectors, e.g.:
```
c("a", "b") %>% map_call("paste")
```

R/smash.R
```diff
@@ -0,0 +1,53 @@
...
+#' @seealso \code{\link{smash}()} and \code{\link{do.call}()}
+#' @export
+#' @examples
+#' # We map a list of strings to paste(), with sep = "-" and the
+#' # string "2001" as additional arguments
+#' list("01", "01") %>%
+#'   map(~ sub("^01", "10", .)) %>%
+#'   map_call("paste", "2001", sep = "-")
+map_call <- function(.x, .f, ...) {
+  do.call(.f, c(.x, list(...)))
```
hadley: `quote(.x)` would be slightly better here

hadley:
One last tiny nitpick - otherwise looks great!

lionelgit:
thanks, here you go :-)

lionelgit:
hmm with the quoted `.x`, `map_call()` does not work with atomic vectors anymore.

lionelgit:
Actually it does not work with lists either. At least not in all situtations:
```
x <- list("a", "b")

do.call("paste", list(quote(.x)))
## [1] "a" "b"

do.call("paste", .x)
## [1] "a b"
```

lionelgit:
with `fnu <- function(...) match.call()` :
```
do.call("fnu", .x)
## fnu("a", "b")
do.call("fnu", list(quote(.x)))
## fnu(.x)
```

lionelgit:
And I just realised that it makes sense to also allow dots in the closure returned by `smash()`

hadley:
Oops, yes, that makes sense.

PR #33, opened 2015-01-09
Implement at_depth()

lionelgit:
While writing the PR I noticed that `at_depth()` is basically a recursive version of `map()`, not necessarily an adverb for functions that take a list. So now I find it a bit odd that it is not called something like `map_at_depth()` (but I don't think I like this name). 

Also, the semantics is not great when the list is not piped. e.g.,
```
list %>% at_depth(3, sum)  # looks good
at_depth(list, 3, sum)     # looks odd
```

So from this point of view maybe the idea of a factory function is better:
```
list %>% at_depth(3, sum)()
at_depth(3, sum)(list)
```


lionelgit:
> So now I find it a bit odd that it is not called something like `map_at_depth()`

hmm after giving it more thought, I don't think it would be a good name at all. `x %>% map_at_depth(2, fun)` should mean to `x %>% at_depth(2, map, fun)` to be semantically consisent. But since `map(x, fun)` is actually synonym of `x %>% at_depth(1, fun)` this would effectively be a mapping at level 3, not level 2. This is not very intuitive. So from this point of view `at_depth()` is better.

About the semantic issues of using `at_depth()` with and without pipes. Would it be an option to have differing function signatures as a function of the type of the first argument (through S3 methods for example)? So that both
```{r}
at_depth(2, fun, x)
```
and
```{r}
x %>% at_depth(2, fun)
```
would be valid.

hadley:
Hmmm, I'm not a big fan of manual currying (which it sounds like the s3 option would be equivalent to). Currying is great if it's built into the language, but I think it's an awkward fit with R argument dispatch rules (ie. I think you can have currying or named arguments not both)

lionelgit:
The point about named arguments made me realized that we can just do
```
at_depth(3, fun, .x = list)
```
which is way simpler. Problem solved :-)

PR #34, opened 2015-01-10
Data frame coercion in multi maps

lionelgit:
This patch fixes my PR of yesterday so that `map2()` and `map3()` coerce their outputs to a data frame they get one as input. I added some tests to check we don't break this again in the future. However I'm not sure what to do about `map_n()`. Since the result is never a named list, it can never be coerced by `as_data_frame()`. Non-exclusive possibilities to solve this:

1. Never coerce to a data frame (don't use the output hook).

2. Coerce to a data frame based on the first element of `.l`. This may be a good option because the behavior would be symmetric with `map2()` and `map3()`. This would simplify the code as well (no need of `map_multi()`).

3. Add names automatically. That's what I did in a recent commit pushed to #24. This was to make `cross()` not fail when given a data frame with `.wide = TRUE`. In retrospective I'm not sure this is a good thing, and in any case this should be handled by `as_data_frame()` not by lowliner.

4. Downcast to a list when data frame coercion is not possible (i.e., when the list is not named or when its elements have unequal lengths).

I think 2 and 4 would be nice.


lionelgit:
While we are on the subject of `output_hook()`, would you be open to the idea that when we map over an atomic vector, this vector be used to give names to the resulting list? The vector that is mapped over is often useful metadata and I often find myself doing something like
```{r}
vec <- c("a", "b", "c")
map(vec, my_fun) %>% set_names(vec)
```
It would be nice if this was done automatically and it would only require one conditional branch in `output_hook()`. Of course many times the metadata will not be needed and sometimes will look weird when encoded as names (duplicates, numerics), but even in these cases I don't think it would hurt to get a named list.

R/map.R
```diff
@@ -145,9 +150,9 @@ each3 <- function(.x, .y, .z, .f, ...) {
 #' @rdname map2
 each_n <- function(.l, .f, ...) {
   args_list <- recycle_args(.l) %>% zip()
-  for (args in args_list) {
+  lapply(args_list, function(args) {
```
hadley: I deliberately chose a for loop here to avoid allocating the output space

hadley:
I'm still a bit torn on this behaviour. On one hand, it's not exactly hard to add an explicit `as_data_frame()` to coerce a list to a data frame. OTOH, it would be nice to do it automatically if we can. 

`map_n()` does seem much less likely to produce a data frame than the others - it's probably only likely to produce a data frame if it had a list of data frames as input. I think it's probably ok to not coerce.

R/map.R
lionelgit: oh ok, I didn't realize this loop came from the old code, I thought I introduced him mistakenly

lionelgit:
Complains because of the `rbenoulli.Rd` thing but `check()` passes on my machine.

Safe to merge.

hadley:
Thanks!

PR #36, opened 2015-01-13
Add type predicates and implement enlist()

lionelgit:
Following discussion in #12.

I wrote `is_bare_numeric()` to be the bare equivalent of `is.numeric()`, and added `is_bare_double()`. Tell me if you prefer that `is_bare_numeric()` tests for doubles.


hadley:
I'd prefer that `is_bare_numeric()` tests only for "numeric" vectors, i.e. with `is.double()`.

`is_list()`, `is_bare_list()`, `is_vector()` (`is_list()` or `is_atomic()`) and `is_bare_vector()` would also be useful.

hadley:
FYI the problem with `is.vector()` is that it tests for non-name attributes, not it's behaviour with `NULL`

lionelgit:
ah yes.

And should we have a `is_numeric()` companion for `is_bare_numeric()` then?

What do you have in mind for `is_list()`?

hadley:
Yes to `is_numeric()`, and `is.list()` could just be an alias for `is.list()`

hadley:
I guess we should probably complete the set with `is_character()` , `is_bare_character()`, and  `is_logical()`, `is_bare_logical()`.

NAMESPACE
```diff
@@ -1,4 +1,4 @@
-# Generated by roxygen2 (4.1.0): do not edit by hand
+# Generated by roxygen2 (4.1.0.9001): do not edit by hand
```
hadley: Would you mind using the release roxygen2 please?

R/objects.R
```diff
@@ -0,0 +1,141 @@
...
+#'
+#' These type predicates aim to make type testing in R more
+#' consistent.
+#'
+#' Most of these predicates are simple aliases to base R functions. In
+#' addition:
+#' \itemize {
+#'   \item Unlike \code{is.atomic()}, \code{is_atomic()} does not
+#'         return \code{TRUE} for \code{NULL}.
+#'   \item Unlike \code{is.vector()}, \code{is_vector()} does not
```
hadley: Maybe `Unlike \code{is.vector()}, \code{is_vector()} test if an object is an atomic vector or a list. \code{is.vector} checks for the presence of attributes (other than name)`

hadley:
Something looks wrong with the generated docs and can you please fix the R CMD check failure?

lionelgit:
Sorry for the roxygen mess, I will fix the other PR.

The doc should be ok now. In the future I'll try to install the repo before submitting.

lionelgit:
roxygen version should be ok in all PR

hadley:
Do you use RStudio?

lionelgit:
I use Emacs

hadley:
Ah, well it's super easy in RStudio, but if you do `devtools::check()` you'll get the same local checks as travis runs

lionelgit:
hmm I do run `check()` with a keybinding... must have missed something

hadley:
I'm still seeing the same failure on travis

lionelgit:
Note to self: run `check()` on a fresh session.

lionelgit:
I'll add: don't confuse `test()` and `check()` ...

lionelgit:
Well this was embarrassing

lionelgit:
Now the problem is that yesterday I did not remove "rbenoulli.Rd" from the git index. As you can tell I'm new to all this.

I removed it from the index in #37 and `check()` does not complain on my machine anymore for this branch so I think it's ok to merge.

I'm in the process of running `check()` on all other branches now.

PR #37, opened 2015-01-13
Misc fixes and enhancements

lionelgit:
The third commit enhances the output when an atomic vector is mapped over.
`map()` uses the vector (often useful metadata) to name the elements of the returned list.

```
map(c("a", "b", "c"), partial(rep, times = 2))
$a
[1] "a" "a"

$b
[1] "b" "b"

$c
[1] "c" "c"
```

If you prefer the old behaviour, I'll revert 408d.


hadley:
I prefer to not automatically name character vector inputs, because that's what `vapply()`, `sapply()` do and I find it annoying more often than not. If you think it's useful, I'd rather have another function like `auto_name <- function(x) setNames(x, x)`

PR #38, opened 2015-01-14
Make functions follow .arg convention

lionelgit:
This replaces PR #37 where git history was messy.

I would also move `empty()` to objects.R and rename it to `is_empty()` if you're ok with that. `empty(list)` gives the impression that we are emptying a list rather than checking that it's empty.

I can also make the random generator functions and the type predicates follow the .arg convention. I felt like it would be weird for them to do so because they are completing or mimicking base R rather than being part of the functional programming tools. Naming is hard.

### DIFF
--- a/R/at-depth.R
+++ b/R/at-depth.R
@@ -59,7 +59,7 @@ at_depth <- function(.x, .depth, .f, ...) {
 
   recurse <- function(x, depth) {
     if (depth > 1) {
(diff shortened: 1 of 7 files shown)


hadley:
I like the idea of `is_empty()`.  I don't think the non-FP functions need to have `.` prefixes.

PR #40, opened 2015-01-15
Add prepend()

lionelgit:
Since we don't have operators to merge lists and vectors, this adds `prepend()` to help merging a list at the beginning of another.

I was tempted to use `output_hook()` so that the following would produce a data frame:
```
mtcars %>%
  prepend(list(index = 1:32))
```
But then `prepend()` wouldn't be symmetric with `append()`...


lionelgit:
Of course, this is not an essential addition because we already have `c()` and `append()`, but I find that in pipes `prepend()` produces clearer code than using `c()`.

Issue #9, opened 2014-12-02
Sliding and grouped


hadley:
To partition list into pieces

hadley:
No longer seems that important to me

PR #41, opened 2015-01-21
Binary mappings

lionel-:
This adds binary mapping functions.

* `map_pairs()` applies a function over combinations of `.x` generated by `cross_n()`.

* `map_neighbours()` applies a function to all sequential elements of `.x`. This is particularly useful to check that all elements are identical.

This also adds predicate versions of these functions: `probe_`, `every_` and `some_` variants.

One aspect I'm not sure you'll like is that `map_pairs()` uses attributes to record the parents of the returned elements. Otherwise there is no way to know for sure.


hadley:
This feels a bit too special case for me.  I'd rather think about how we can make it easier for people to construct this sort of function themselves.

hadley:
This is what I was thinking for #9

PR #42, opened 2015-01-24
Rename smash() to stitch() and implement smash()

lionel-:
This implements a complement to `smash()`, ie a functional that transforms a function taking a list into a variadic function.

While I was thinking about a name for that new functional, I realised we got the analogy backwards. I think `smash()` would better be called `stitch()` and the new functional should be called `smash()` instead. The smash analogy would work better with this functional because the transformed function goes from one tidy entry point to many ones.

On the other hand, if we transform a function that takes many arguments into one that takes a list, we `stitch()` its function signature into one piece :-)


lionel-:
s/functional/factory

lionel-:
I think I wasn't clear enough that smashing a function (turning a function that takes a list into a variadic one) is sometimes needed for function composition. In particular, it allows to pass a function that takes a list to `map_n()` or `map_rows()`.

e.g., the following checks that each row is monotonically increasing.
```
mtcars %>% map_n(smash(every_neighbours, `<`))
```

But I don't know if `map-pairs.R` will get merged so I didn't include that example in the doc. I'm trying to find another compelling example but I don't think there are many to find. Indeed `smash()` will only be useful in a handful of situations. But that's just as well, smashing a function should be a rare treat.

hadley:
I like the general idea, but I think the parallel between `stitch()` and `smash()` could be stronger. Maybe `stitch()` and `unstitch()` (that would parallel `zip()`/`unzip()`?  Or `unpick()` (actual sewing terms)?

It might also be clearer to document both functions in one file, so you can emphasise that they're symmetric. (And show that's the case in an example)

lionel-:
I wonder if we should have a new naming policy for functional adverbs such as `stitch()`, `at_depth()` or `by_slice()`, to prevent clashes with `.f` arguments.

For example, if I unstitch `map()` and try to give it a default function in the `unstitch()` call by referring explicitly to `.f`, this will not work as expected:
```
fnu <- unstitch(map, .f = mean)
fnu(1:4, 1:7)
```
```
## [1] NA
## Warning message:
## In mean.default(list(...), function (.x, .f, ...)  :
##   argument is not numeric or logical: returning NA
```

With another argument name:
```
unstitch2 <- function(..f, ...) {
  force(..f)
  defaults <- list(...)
  function(...) {
    do.call("..f", c(quote(list(...)), defaults))
  }
}

fnu <- unstitch2(map, .f = mean)
fnu(1:4, 1:7)
```
```
## [[1]]
## [1] 2.5
## 
## [[2]]
## [1] 4
```

At first I thought of `.fun` but this does not work because of argument matching (it took me a long time to figure out what was going on). I saw in `partial()` that you named its function arg `_f`. This is a bit constraining as it requires backticks but it's also ok since it would be rarely needed to explicitly refer to this argument. Another option is `..f` as in `unstitch2()` above.

hadley:
I like using `..f` in adverbs.

hadley:
Otherwise, looks good to merge

lionel-:
While I was experimenting with a mapping functional for #41 and #9, I stumbled unto the following problem.

Say you want to stitch `identical()`'s signature and feed it a named list of two elements (for example a two-column subset of a data frame) with the intention that those two elements get mapped to the first two arguments of `identical()`. This will not work unless the two elements of the list are named `x` and `y`, as in `identical()`'s signature.

One way out of this would be to provide `stitch()` with a `.unname` option.

```
stitch <- function (..f, ..., .unname = FALSE) {
  force(..f)
  defaults <- list(...)
  function(args = list(), ...) {
    if (.unname) {
      args <- unname(args)
    }
    do.call("..f", c(args, defaults, list(...)))
  }
}
```

This causes an error:
```
hip <- stitch(identical, .unname = FALSE)
hip(mtcars[c(1, 2)])
```
```
## Error in ..f(mpg = c(21, 21, 22.8, 21.4, 18.7, 18.1, 14.3, 24.4, 22.8,  (from stitch.R!5402NaB#8) : 
##   unused arguments (mpg = c(21, 21, 22.8, 21.4, 18.7, 18.1, 14.3, 24.4, 22.8, 19.2, 17.8, 16.4, 17.3, 15.2, 10.4, 10.4, 14.7, 32.4, 30.4, 33.9, 21.5, 15.5, 15.2, 13.3, 19.2, 27.3, 26, 30.4, 15.8, 19.7, 15, 21.4), cyl = c(6, 6, 4, 6, 8, 6, 8, 4, 4, 6, 6, 8, 8, 8, 8, 8, 8, 4, 4, 4, 4, 8, 8, 8, 8, 4, 4, 4, 8, 6, 8, 4))
```

This works:
```
hop <- stitch(identical, .unname = TRUE)
hop(mtcars[c(1, 2)])
```
```
## [1] TRUE
```

Maybe `.blank_signature` or `.blank` would be a more informative arg name for this.

lionel-:
The last commit adds the argument discussed in the previous comment.
I will of course revert it if you prefer not to have it.

lionel-:
I added `unstitch_v()` to help with functions that takes atomic vectors, like `mean()`.

```{r}
unstitch(match_call)(1, 2, 3)
## ..f(list(1, 2, 3))

unstitch_v(match_call)(1, 2, 3)
## ..f(c(1, 2, 3))
```

I find the name `unstitch()` a bit confusing though. When I use it, I need to think about what it does. I think the easiest analogy is that it makes a function take dots instead of a list or vector. So I wonder if it may be better to have a function name that evokes dots, like `sprinkle()` or `scatter()`? Hmm probably not those two. `punctuate()`?

Lastly, to complete the set of functions that help composition, I think we need a factory that turns a function that takes a vector into one that takes a list (and maybe vice versa). This would make functions such as `mean()` compatible with more functionals. Would it be too much?

lionel-:
> to complete the set of functions that help composition, I think we need a factory that turns a function that takes a vector into one that takes a list

Here is one implementation using `can_simplify()`

```{r}
informative_name <- function (..f, ..., .type) {
  force(..f)
  defaults <- list(...)
  if (missing(.type)) .type <- NULL

  function(args = list(), ...) {
    if (can_simplify(args, .type)) {
      args <- unlist(args)
    } else {
      stop("Can not coerce input to a vector", call. = FALSE)
    }
    do.call("..f", c(list(args), defaults, list(...)))
  }
}

informative_name(match_call)(list(1, 2, 3))
## ..f(c(1, 2, 3))
```

lionel-:
Could be called `lift()`

```{r}
lift(mean)(list(1, 2, 3))
```

pretty clear I think :-)

We lift `mean()` to the realm of purer functions that take a list. The opposite operation could be called `shackle()`.

```{r}
shackle <- function (..f, ...) {
  force(..f)
  defaults <- list(...)
  function(.x, ...) {
    do.call("..f", c(list(as.list(.x)), defaults, list(...)))
  }
}

shackle(match_call)(c(1, 2, 3))
## ..f(list(1, 2, 3))
```

Also I think I like `punctuate()`. So what about those two pairs of composition helpers, `stitch()` / `punctuate()` and `lift()` / `shackle()`?

lionel-:
Implemented the proposed changes and also `stitch_v()` to complete the set. The following table summarises the coercions operated by the factories:

```
stitch()      : from ...       to list(...)
punctuate()   : from list(...) to ...      
                                         
stitch_v()    : from ...       to c(...)   
punctuate_v() : from c(...)    to ...      
                                         
lift()        : from c(...)    to list(...)
shackle()     : from list(...) to c(...)
```

lionel-:
I wonder how strict we should be with inputs.

For example, `shackle()` works by applying `as.list()` on the inputs, so that atomic vectors are converted to lists before being passed on to the original function. But this also makes the shackled function compatible with lists. Since the intent behind using `shackle()` is to make `..f` accept atomic vector, maybe we should have an error when it gets a list.

hadley:
We now have a table like this, right?

From/To      |  `...`           | `c(...)`     | `list(...)`
------------ | -----------------|--------------|------------
`...`        |                  | `stitch_v()` | `stitch()` 
`c(...)`     |  `punctuate_v()` |              | `lift()`   
`list(...)`  |  `puncutate()`   | `shackle()`  |            

I think it's getting sufficiently complicate that it might be worth taking the same approach as plyr and just having a consistent set of suffixes: e.g. `stitch_vl`, `stitch_ld`, with v = vector, l = list, and d = dots?

lionel-:
I think I prefer the expressivity of the proposed function names but you're right that it may be too complex.

hadley:
I think we have two options - either pick the best possible verbs for the most common cases, or use plyr style naming. I think the operations are sufficiently esoteric that it's going to be hard to find evocative verbs, so I'm leaning towards plyr style. What do you think?

lionel-:
personally the most evocative name is `punctuate()`. But it's not a very common operation.

I think you are right that people will be lost with all those names. With your scheme, it's clear that any `stitch_xy()` refers to a composition helper.

hadley:
Maybe `punctuate_xy()`? I do like the name since it's all about changing how ... works

lionel-:
I also think we can find a better name than `stitch()`.

> Maybe punctuate_xy()?

But there is no `...` involved if we transform a function taking a list to one taking a vector and vice versa. Plus, I think it'd be better to have a short name here.

How about `smash()`? We smash a function signature to transform it... Though it doesn't really evoke transformation.

lionel-:
`lift_xy()`? We lift the domain of a function from `x` to `y`

hadley:
I like `lift()`.

lionel-:
me too :)

I don't remember why but I wasn't happy with what I did with `can_simplify()` in this PR. The modifications were involved in coercing lists to vectors with `lift_lv()`. I'll try to come up with something consistent while working on `as_vector()`.

lionel-:
by the way, in Haskell they use `flip()` to change the signature of a function from `a → b → etc → res` to `b → a → etc → res`.

I was wondering if it would be interesting to have this? The thing is that we have named arguments so instead of:
```r
fun <- function(x, y) {
  paste(x, y)
}

map(names(mtcars), flip(fun), "prefix")
```
we can simply do:
```r
map(names(mtcars), fun, x = "prefix")
```

lionel-:
there you go. I've renamed according to `lift_xy()` pattern and I've added `as_vector()`.

`as_vector()` accepts both vector molds such as `double(3)` and strings describing types such as `"logical"`. This can be a problem if for instance someone tries to use `"character"` as a mold but I don't think this is likely.

hadley:
Thanks!

lionel-:
@hadley 

I'm having second doubts about the names. Globally they work well, but I always have to pause a second when I want to splat a function (which is done with `lift_dl()` now). That's a problem since it's such a common function.

Maybe it's because the dots analogy doesn't work so well when there are no dots involved strictly speaking. Possible solutions:

* Have `lift_l()` and `lift_v()` for the cases where we act on a whole function signature, instead of `lift_dl()` and `lift_dv()`.
* Maybe even `lift()` and `lift_v()` since lists are privileged in purrr.
* Or just having `lift()` as an alias for `lift_dl()`. This would keep the consistent xy naming while providing an easy-to-remember name for the most common case.

hadley:
I like `lift()` as alias for `lift_dl()`

PR #50, opened 2015-02-09
Chart

lionel-:
Here is a second attempt. The tools implemented in this PR are more general than in #41 and hopefully correspond to what you had in mind for #9.

* There is a mapping functional `map_along()` that applies a function on a sequence of subsets of a list `.x`. The sequence is controlled by `.route`, which can be a list of indices or a function that takes `.x` and returns a sequence.

  I added two such functions, `seq_window()` and `seq_pairs()`. The first can be used to generate indices for a sliding window, while the latter uses `cross()` to generate pairs of indices.

* `chart()` is like `map_along()` but it is a factory that transforms a function that takes a list to apply it on a given route (a given sequence of indices). This way we have a factory and a functional that serve the same purpose, just like `map_call()` and `stitch()`.

* I also added a type predicate `is_index()` to test if an atomic vector can be used as indices (i.e., it tests if a vector is either character or numeric).


lionel-:
Travis fails because this pull depends on #42 (it uses `stitch()` in an example).

Other than that, `check()` generates a note because I use magrittr's placeholder in `seq_pairs()`.
```
seq_pairs: no visible binding for global variable ‘.’
```

What is your policy on this? Add `globalVariables(".")` somewhere in the package (if so, where would you place it?), or avoid the placeholder altogether?

lionel-:
Since this branch has `smash.R` in the git index and `smash()` in the `NAMESPACE` file, I guess I'll have to rebase once #42 is merged.

Anyway, I also have these tools written using `map_along()`:

```{r}
#' Are the elements of a list identical or equal?
#'
#' \code{are_identical()} and \code{are_equal()} are equivalent to
#' respectively \code{identical()} and \code{all.equal()} except that
#' they take a list and operate on all consecutive pairs of
#' elements. The \code{all_identical()} and \code{all_equal()}
#' variants check that all elements are equal or identical and return
#' \code{TRUE} or \code{FALSE}.
#' @param .x A list of elements to be tested.
#' @return A list of size \code{length(.x) - 1} for
#'   \code{are_identical()} and \code{are_equal()}. The former returns a
#'   list of \code{TRUE} or \code{FALSE} while the latter also return
#'   strings describing the difference between two
#'   elements. \code{all_equal()} and \code{all_identical()} return
#'   \code{TRUE} or \code{FALSE}.
#' @name comparison
NULL

#' @rdname comparison
#' @export
are_identical <- function(.x) {
  .x %>% map_along(seq_window, identical)
}

#' @rdname comparison
#' @export
are_equal <- function(.x) {
  .x %>% map_along(seq_window, all.equal)
}

#' @rdname comparison
#' @export
all_identical <- function(.x) {
  .x %>% are_identical() %>% map_call(all)
}

#' @rdname comparison
#' @export
all_equal <- function(.x) {
  res <- .x %>% are_equal()
  every(res, is_logical) && map_call(res, all)
}
```

I'm not sure about the `are_` variants but I think `all_identical()` and `all_equal()` would be nice additions to purrr.

R/chart.R
```diff
@@ -0,0 +1,87 @@
+#' Chart the application of a function along coordinates
+#'
+#' \code{chart()} modifies a function that takes a list to make it act
```
hadley: I can understand all the words individually but I don't really get what you're trying to say :/

hadley:
I like the `all_equal()` and `all_identical()` functions

R/chart.R
lionel-: yes that's not well written and maybe the whole idea of a factory function is not necessary.

Was I clear on how `map_along()` works at least?

hadley: I think so - it's a generalised `map_pairs()` right?

lionel-: yes

Issue #45, opened 2015-01-24
cross_n() and data frames


lionel-:
While writing the example in #44, I was disappointed that I couldn't find an easy way to reproduce
```
expand.grid(mean = 1:5, sd = seq(0, 1, length = 10))
```
with `cross_n()`. This is because the elements to cross have unequal lengths so we can't make a data frame to give as input to `cross_n()`.

So maybe we do need `cross_d()` after all. This way `cross_d(list(a = 1:3, b = letters[1:5]))` would work as expected. We could also just resort to `expand.grid()` for this kind of inputs, but I think it's good to have alternatives to any df function that coerces character vectors to factors (and also to any function with dots in its name ;).

lionel-:
Fixed in #52

PR #52, opened 2015-02-20
Implement cross_d() for data frame outputs and make cross_n() only return lists

lionel-:
As proposed in #45

I think this is a more useful interface. It allows to cross elements of unequal length in long format.
With the previous implementation, it was not easy to do this:
```{r}
list(a = 1:2, b = letters) %>% cross_d()
```


sumprain:
`cross_d` is going to be a good function for list --> dataframe.

hadley:
Thanks!

PR #55, opened 2015-03-01
Fix #54

### LINKED ISSUE #54: map2 and map3 fail with vector as additional argument
E.g.
```
f <- function(x, y, zs) x + y + mean(zs)
map2(1:2, 1:2, f, zs = 1) # works
map2(1:2, 1:2, f, zs = 1:2) # wrong result: vectorizes over zs too
map2(1:2, 1:2, f, zs = 1:3) # Error: all(lengths %in% c(1, n)) is not TRUE
```

lionel-:
I added a `.first` arg in `partial()` to control the position of dots, which can be important if unnamed arguments are supplied (see second `map_n()` test).

I also renamed `_f` to `...f` in `partial()`. The idea is to have `.f` for mapping functions, `..f` for adverbs and other modifiers, and `...f` for functionals that handle both mapping functions and modifiers.


hadley:
Thanks!

Issue #56, opened 2015-03-03
New adverbs `at()` and `where()`


lionel-:
One fundamental limitation of replacement functions such as `[<-` is that they cannot replace an element with a larger one.  i.e., it's illegal to do this:
```{r}
mtcars[3] <- data.frame(a = mtcars[[3]], b = mtcars[[3]])
```

This is because it would be impossible to know what goes where in case the replacement region is not contiguous. But it makes it hard to work with functions that take a data frame column and may return a larger data frame, like `disjoin()` (see #28 for some background on this tool). We really need a way to apply this kind of function on subsets of columns.

So I wrote these two adverbs inspired by `fapply()`:
```{r}
at <- function(.x, .where, ..f, ...) {
  if (is.character(.where)) {
    if (is.null(names(.x))) {
      stop(".x is not named", call. = FALSE)
    }
    .where <- match(.where, names(.x))
  }
  if (anyNA(.where)) {
    stop("some indexes are missing", call. =FALSE)
  }

  out <- vector("list", length(.x))
  for (i in seq_along(.x)) {
    res <-
      if (i %in% .where) {
        ..f(.x[i], ...)
      } else {
        .x[i]
      }
    stopifnot(is.list(res))
    out[[i]] <- res
  }

  flatten(out) %>% purrr:::output_hook(.x)
}

where <- function(.x, .p, ..f, ...) {
  sel <- probe(.x, .p) %>% which()
  .x %>% at(sel, ..f, ...)
}
```

For example,
```{r}
diamonds %>% at("cut", disjoin, sep = " / ") %>% str()
## Classes ‘tbl_df’, ‘tbl’ and 'data.frame':	53940 obs. of  14 variables:
##  $ carat          : num  0.23 0.21 0.23 0.29 0.31 0.24 0.24 0.26 0.22 0.23 ...
##  $ cut / Fair     : num  0 0 0 0 0 0 0 0 1 0 ...
##  $ cut / Good     : num  0 0 1 0 1 0 0 0 0 0 ...
##  $ cut / Very Good: num  0 0 0 0 0 1 1 1 0 1 ...
##  $ cut / Premium  : num  0 1 0 1 0 0 0 0 0 0 ...
##  $ cut / Ideal    : num  1 0 0 0 0 0 0 0 0 0 ...
##  $ color          : Ord.factor w/ 7 levels "D"<"E"<"F"<"G"<..: 2 2 2 6 7 7 6 5 2 5 ...
[...]

Issue #28, opened 2014-12-26
Preserving map


lionel-:
It would be great if lowliner had a mapping function that uses preserving subsetting as in this gist: https://gist.github.com/hadley/8667699

One difficult aspect is to create an interface that makes it easy to work with both the list component `x[[i]]`  and the (I'll call it like that for lack of a better idea) list frame `x[i]`. In your gist, you supply the list frame to the mapped function. I experimented with an alternative approach in which I supplied the list component instead, as is usual in mapping tools, along with a new placeholder referring to the list frame.

This new placeholder is called `..`. Since the double dot evokes the action of going one level up in a file hierarchy, I figured it would work well to represent the list frame. It is implemented as an active binding so that any modification to the list component is reflected in `..`.

Example:
```
test <- list(a = list(1, 2), b = list(3, 4))

out <- remap(test, function(x) {
  if (names(..) == "a") {
    rep(.., 2) %>%
      set_names(c("a1", "a2"))
  } else {
    x <- list(5)
    names(..) <- "B"
    ..
  }
})

str(out)

## List of 3
## $ a1:List of 2
##  ..$ : num 1
##  ..$ : num 2
## $ a2:List of 2
##  ..$ : num 1
##  ..$ : num 2
## $ B :List of 1
##  ..$ : num 5

```

The code is in https://gist.github.com/lionel-/12350ba7e583e9c10163

hadley:
The use of a placeholder feels too magical to me - when reading the code, you'll need to remember where `..` comes from and how it's defined.

lionel-:
It also does not feel very functional. Magrittr's placeholder is a shortcut for an input that has an explicit definition in the function signature. But the `..` placeholder creates a new implicit input.

Plus I can envision the special evalution implementing this placeholder to break in all sorts of ways.

lionel-:
I encountered a good use case for preserving maps. It shows that it's definitely more useful to pass a list, `x[i]`, rather than an element, `x[[i]]`. It's easier to see this with functions that work on data frames.

The use case is to create disjunctive tables from factors to produce a design matrix prior to modelling. Here are some tools to do that:

```{r}
disjoin <- function(x, ...) {
  UseMethod("disjoin")
}

disjoin.default <- function(x, ...) {
  x <- as.factor(x)
  out <- lapply(levels(x), function(level) {
    as.numeric(x == level)
  })
  names(out) <- levels(x)
  dplyr::as_data_frame(out)
}

[...]

PR #57, opened 2015-03-06
list-preserving mapping functionals

lionel-:
As discussed in #28 and #56

`at_each()` may be a bit of a confusing name due to `each()`. But what about a new name for this one: `walk()`
We'd walk a function along a list, leaving a mess of side effects behind :)
```r
walk <- each
walk2 <- each2
unassign <- function(x, ...) rm(list = x, ...)

names(mtcars) %>%
  walk2(mtcars, assign, envir = globalenv())

names(mtcars) %>%
  walk(unassign, envir = globalenv())
```
I think it's a bit more expressive than `each()`


R/map.R
```diff
@@ -182,3 +182,41 @@ map_if <- function(.x, .p, .f, ...) {
   .x[sel] <- lapply(.x[sel], .f, ...)
   .x
 }
+
+#' Modify elements of a specified subset.
+#'
+#' @inheritParams map
+#' @param .at A character vector of names or a numeric vector of
```
hadley: Or a logical vector?

hadley:
I like `walk()`. `at()` seems a bit short to me

lionel-:
maybe we need a verb. That would also allow to have `verb_if()` to mirror `map_if()`. But I couldn't find any good verb :/

`do_at()`, `do_if()` and `do_each()` ?

R/map.R
lionel-: I didn't document it because this functionality overlaps with `map_if()` and I was afraid it would confuse things. Should I document it anyway? Or maybe make `map_at()` fail with logical vectors?

R/map.R
```diff
@@ -182,3 +182,41 @@ map_if <- function(.x, .p, .f, ...) {
...
+#' mtcars %>% map_at(c("cyl", "am"), as.character) %>% str()
+map_at <- function(.x, .at, .f, ...) {
+  .f <- as_function(.f)
+  sel <- inv_which(.x, .at)
+
+  .x[sel] <- lapply(.x[sel], .f, ...)
+  .x
+}
+
+inv_which <- function(x, sel) {
```
lionel-: Maybe this is useful enough to be exported. Could be named `index()` then?

R/map.R
hadley: Oh hmmmmm, I wonder if there's a way we can integrate this with `map_if`? Maybe at least they should be documented together so it's easy to see the differences/similarities (i.e. `map_if()` is all about logicals, `map_at()` is all about position)

R/map.R
hadley: Lets leave it unexported for the meantime, and see if a need crops up

lionel-:
* `map_if()` and `map_at()` are now documented together, and `map_at()` will fail with a logical vector so that the functionality is distinct.
* I renamed `each()` to `walk()`.

Should I rename `at()`, `where()` and `at_each()` to `do_at()`, `do_if()` and `do_each()`?

hadley:
I like the idea of symmetric names, but I'm not sure if `do` is the right prefix:

* it's rather different to `dplyr::do()`
* it's a bit short (given that it's a relatively uncommon operation)

Maybe `lmap` (list-map) or similar ?

lionel-:
it's true that uncommon functions don't need cute names.

I think the `l` in `lmap()` is a good mnemonic. I renamed the functions with this prefix.

hadley:
Thanks!

hadley:
BTW how are your drawing skills? I'm planning on writing a vignette, and think some visual representation of these different functions will be useful

lionel-:
Uh, well, apart from ggplot I don't have drawing skills :)

Are you thinking about a diagram?

hadley:
Yeah, I'm hoping I can make a picture that shows the difference between `map`, `walk`, `lmap`, `map_v` etc

lionel-:
and all those composition helpers, though maybe it's better they have their own graph/table.

The idea is to organise the chart according to input and output type I suppose. Or do you have other criteria in mind?

hadley:
Yeah - I just want a few table/diagrams that help to orient people to the package as a whole (and I find such tables also tend to highlight if we've missed any combinations).

hadley:
One thing I've been wondering about is if `map_v()` should become `vmap()` so that we can have `vmap2()`, `vmap3()`, `vmap_if()` etc.  But that's a little awkward because the 2 refers to input and the v refers to output, which seems the wrong way around in the function name.

lionel-:
Random idea: have a `v()` adverb that coerces a list to a vector?
```r
c(mtcars) %>% v(map, mean, .type = numeric(1))
```
or
```r
c(mtcars) %>% map(mean) %>% as_vector(numeric(1))
```

This seems more generic. And in this case we can propose additional verbs for common use case, as with `probe()` which is a shortcut for `vapply(x, f, logical(1))`.

hadley:
That would be nice. `map_v` has a small performance advantage (since you can feed directly into a vector rather than creating an intermediate list), but that's probably not important in practice.  

I prefer the `as_vector()` form.

And maybe we should have `map_lgl` (instead of `probe()`), `map_num()`, and `map_chr()`? (need better names)

lionel-:
I see you keep turning all my nice function names into programmy abbreviations :-)

but it would be nice to have those additional aliases and then it makes sense to have a common semantic structure indeed.

hadley:
Maybe because I'm only just dipping my toes into purrr development, but I keep having to really think what the functions do and how they're related. It's a bit hard to calibrate, but my sense is that people will use purrr less often than dplyr (usually either in munging tree data into data frames or during modelling), so having names that clearly show the relationship between functions might be more important. Or, in other words, purrr feels more like plyr than dplyr to me.

lionel-:
Personally I see purrr as more general than that. I see it as a "better standard lib" that offers a consistent interface for working with lists, the primary R object for programming. So purrr may be useful for data analysis, but it's especially nice for function and package development.

Now if purrr is a better standard lib, it also means it's not a DSL for lists. And I suspect many programmers don't like cute expressive names as much as I do. Brevity and internal consistency may be more important to them. So I trust you more than myself on this.

hadley:
Good points. I think the names will become obvious as we work on the documentation. It's hard to pick good names in isolation - they need to work as a whole across the entire package. But it's important to get it right now as it gets harder to change once purrr is on cran.

Issue #58, opened 2015-03-19
Should flatten(list()) return list() instead of NULL?


lionel-:
I expected that `flatten()` applied on an empty list would return an empty list rather than `NULL`. This behavior is caused by `unlist()` but I don't find it very intuitive.

I can send a PR to make it return a list instead.

lionel-:
`enlist()` without arguments also returns `NULL`. I think it should return an empty list instead.

hadley:
That seems reasonable to me

PR #60, opened 2015-03-31
Enhancements to flatten()

lionel-:
This adds a `.unname` argument to `flatten()`. The goal is to avoid parasite names such as in the following example:
```{r}
l <- list(A = list(a = 1, b = 2), B = list(c = 3, d = 4))
flatten(l) %>% str()
## List of 4
##  $ A.a: num 1
##  $ A.b: num 2
##  $ B.c: num 3
##  $ B.d: num 4
```

With `.unname = TRUE`, this becomes:
```{r}
flatten(l, .unname = TRUE) %>% str()
## List of 4
##  $ a: num 1
##  $ b: num 2
##  $ c: num 3
##  $ d: num 4
```

The unnaming behavior differs as a function of `.recursive`: with recursive flattening, `unlist()` is called with `use.names = FALSE`. In the non-recursive case, only the names of the outer list are zapped.

This also makes `flatten(list())` return an empty list instead of `NULL` as suggested in #58.

### DIFF
--- a/R/flatten.R
+++ b/R/flatten.R
@@ -3,8 +3,13 @@
 #' This is a wrapper for \code{\link{unlist}()} that defaults to
 #' \code{recursive = FALSE}.
 #'
+#' When \code{.unname} is \code{TRUE} and \code{.recursive} is
+#' \code{FALSE}, only the names of \code{.x} are removed, leaving the
+#' inner lists intact. But if \code{.recursive} is \code{TRUE}, the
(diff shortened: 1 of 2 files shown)


lionel-:
one thing that would also be really nice is the ability to specify S3 objects to be considered "atomic", i.e., that shouldn't be unlisted.

But that would require rewriting `unlist()` in Rcpp.

hadley:
Thanks!

PR #61, opened 2015-03-31
Enhancements to enlist()

lionel-:
This adds a `.objects` argument to `enlist()` that allows to control which S3 objects should be treated as bare lists. This is especially useful to treat data frames as regular lists. `.objects` can be a character vector of classes or a predicate function.

This also makes `enlist()` called without argument return an empty list instead of `NULL`, as suggested in #58.


hadley:
Can you give a bit more context on where you think this would be useful?

lionel-:
If you think this option is too general, it could be just about data frames: `enlist(x, data_frames = TRUE)`?

And I just realised `enlist()` would be better named `splice()`.

man/enlist.Rd
```diff
@@ -4,14 +4,23 @@
 \alias{enlist}
 \title{Concatenate objects in a list}
 \usage{
-enlist(...)
+enlist(..., .objects = FALSE)
 }
 \arguments{
 \item{...}{Objects to concatenate.}
+
+\item{.objects}{If \code{FALSE}, all S3 objects are encapsulated in
```
hadley: So if false, it's equivalent to `list(...)` ?

hadley:
It makes sense, and I prefer the name `splice()` but I still don't get where it's useful

man/enlist.Rd
lionel-: nope it's equivalent to the current behaviour.

lionel-:
> I still don't get where it's useful
you mean the setting proposed in this PR or the whole function?

hadley:
The setting in this PR

lionel-:
well let's forget about it then.

Issue #66, opened 2015-04-22
Turning arrays into lists


lionel-:
To allow purrr's functions to talk to arrays, they need be transformed into lists.

`plyr::alply()` is a flexible tool that makes it is easy to turn arrays into lists, but it can be slow and hard to read. I think it makes sense to offer faster and more expressive functions in purrr for the most common coercions.

I think the most common tasks are flattening dimensions into a list, and enlisting dimensions hierarchically:
```r
flatten_margin <- function(array, margin = NULL) {
  if (is.null(dim(array))) {
    dim(array) <- length(array)
  }
  if (is.null(margin)) {
    margin <- seq_along(dim(array))
  }
  apply(array, margin, list) %>% flatten()
}

enlist_margin <- function(array, margin = NULL) {
  if (is.null(margin)) {
    margin <- seq_along(dim(array))
  }
  if (length(margin) > 1) {
    new_margin <- ifelse(margin[-1] > margin[[1]], margin[-1] - 1, margin[-1])
    apply(array, margin[[1]], . %>% enlist_margin(new_margin))
  } else {
    flatten_margin(array, margin)
  }
}
```

These functions are faster than the corresponding `alply()` operations.
```r
x <- array(seq_len(1e6), c(1000, 100, 10))
microbenchmark::microbenchmark(
  plyr = x %>% plyr::alply(2, .fun = . %>% plyr::alply(2)),
  purrr = x %>% enlist_margin(2:3)
)
```
```
neval: 100
   expr median         unit
1  plyr    322 milliseconds
2 purrr     66 milliseconds
```

And scale much better:
```r
x2 <- array(seq_len(1e6), c(10, 100, 1000))
microbenchmark::microbenchmark(
  plyr = x2 %>% plyr::alply(2, .fun = . %>% plyr::alply(2)),
  purrr = x2 %>% enlist_margin(2:3)
)
```
```
neval: 100
   expr median         unit
1  plyr  11479 milliseconds
2 purrr    241 milliseconds
```

They could be rewritten in Rcpp in the future for an additional performance boost.

lionel-:
Here they are in action:

```r
x <- array(1:12, c(2, 2, 3))

flatten_margin(x) %>% str()
## List of 12
##  $ : int 1
##  $ : int 2
##  $ : int 3
##  $ : int 4
##  $ : int 5
##  $ : int 6
##  $ : int 7
##  $ : int 8
##  $ : int 9
##  $ : int 10
##  $ : int 11
[...]

PR #68, opened 2015-04-28
Implement flatten_margin() and enlist_margin()

lionel-:
As per #66.

`enlist_margin()` may need a better name to reflect the hierarchical output it produces.
Oh, what about `branch_margin()`?


hadley:
I feel like maybe `array_margin()` and `array_list()` might be more evocative names?

lionel-:
at first I had to think a bit about what they mean but now that I get it they make sense.

The downside is that they are not verbs. On the other hand many object constructors are not verbs, so we can indeed think of them not as actions on arrays but as constructors.

though I feel
```r
obj %>% array_list(c(3, 1))
```
is not as expressive as
```r
obj %>% branch_margin(c(3, 1))
```

I think I would prefer this one, and then rename `flatten_margin()` to `enlist_margin()` which makes more sense, especially with regard to the other `enlist()` function.

Of course the decision is yours.

hadley:
When I read `branch_margin()` I have no idea what it does - could you give a bit more detail about why you think `branch()` is the natural verb?

lionel-:
because it creates a tree of lists based on the margin. But maybe `branch` reflects more the idea of one fork than of a whole tree indeed

hadley:
What about `array_branch()` (instead of `enlist_array()`) and `array_tree()` instead of `branch_array()`?

lionel-:
I think that looks good :-)

lionel-:
Here you go. Love these function names!

hadley:
Thanks!

Issue #64, opened 2015-04-16
Inconsistent behavior with map ~ lapply, and map_n ~ Map


jonboiser:
In the package, `map` seems to wrap `lapply`, while `map_n` wraps `Map`, which leads to different behavior. 

For me, the thing that I noticed is that if your input to `map_n` is a character vector, your output is a named list whose names are that character vector. Whereas with plain `map`, it's an unnamed list.

```r
input <- letters[1:5]
map(input, function(x) 1) %>% names #=> NULL
map_n(list(input), function(x) 1) %>% names #=> "a", "b", "c", "d", "e"
```

Before I noticed this, I would have to insert the names using `magrittr::set_names` or `names<-`.

For my own purposes, I actually prefer having the names automatically inserted, so I would just use `map_n` by default from now on.

lionel-:
`map()` and `lapply()` have consistent behavior because they always use the names of the input list or vector to name the output. For example, this will work:
```r
letters[1:5] %>% setNames(., .) %>% map(function(x) 1)
```

On the other hand, `Map()` behavior is not so consistent because it only names the output after the input elements if the input is an unnamed character vector. So, if your input vector happens to have names, or to be any other type than a character vector, `Map()` will not recycle the elements into names.

hadley:
@lionel- can we suppress that behaviour? (i.e. never automatically name output if input is an unnamed character vector)

lionel-:
one use case for `enlist()` that I particularly like is to create functions that take some kind of objects or a list of those objects indistinctly. If your function takes atomic vectors, then it could makes sense to treat data frames as list.
```r
fun <- function(...) {
  dots <- enlist(..., .objects = "data.frame")
  fun_impl(dots)
}
```
then you can use your function in this way: `fun(id = 1:32, mtcars)`.

I figured there could be other use cases so I made the option as general as I could but I admit I have a hard time imagining other situations where this would be useful.

lionel-:
oops, the message above was intended for #61.

@hadley I guess this should do it:
```r
map_n <- function (.l, .f, ...) {
    args <- recycle_args(c(.l, list(...)))
    out <- do.call("Map", c(list(quote(.f)), args))

    if (is.character(args[[1]]) && is.null(names(args[[1]]))) {
      names(out) <- NULL
      out
    } else {
      out
    }
}
```

hadley:
@lionel- blech. I wonder if we should bite the bullet and implement our own `map` in Rcpp

[...]

PR #59, opened 2015-03-25
Update map.R: issue #54

sumprain:
Another way to solve the issue #54. earlier way is done by lionel.

PR #74, opened 2015-05-23
map_n() fixes

lionel-:
Hello,

Since #59 is stuck, I cherry-picked the corrective commit of @sumprain.
I also set `USE.NAMES = FALSE` so that `map_n()` always returns an unnamed list as discussed in #64.

However this last change made `map2()` and `map3()` incompatible with data frames because we need a named list to coerce back to data frame. This is one case where it makes sense to name the output after the names of the first argument so I adjusted `output_hook()` to take care of this.

### DIFF
--- a/R/map.R
+++ b/R/map.R
@@ -154,9 +154,8 @@ map3 <- function(.x, .y, .z, .f, ...) {
 #' @rdname map2
 map_n <- function(.l, .f, ...) {
   .f <- as_function(.f)
-  f <- partial(.f, ..., .first = FALSE)
   args <- recycle_args(.l)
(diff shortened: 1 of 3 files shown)


lionel-:
@hadley can you also have a look at this please?

hadley:
I feel like we should probably write our own rcpp versions of these functions to avoid the annoying work arounds.

lionel-:
on the other hand these functions are unlikely to evolve much. And we probably won't get any substantial performance boost from the rewrite.

hadley:
I'm more interested in the elegance than the performance. I feel like working in C might make the code easier to understand. Not high priority by any means.

PR #77, opened 2015-06-22
Presumed typo in documentation for zip -- missing "list"

peterhurford:


hadley:
Thanks! Had to fix by hand since I'd changed the docs a bit from when you'd submitted

PR #79, opened 2015-08-04
Added predicate function operators for creating predicate functions

paulhendricks:
These functions make it easy to combine functions and evaluate an option using && and ||. I'm unsure if you already have similar functions but I've found them useful. For example:

``` R
is_factor_or_character <- orify(is.factor, is.character)
is_factor_or_character(letters) # TRUE
is_factor_or_character(1:10) # FALSE
is_factor_or_character(factor(letters)) # TRUE
```

``` R
between_0_and_10 <- andify(function(x) x > 0, function(x) x < 10)
between_0_and_10 (0) # FALSE
between_0_and_10 (5) # TRUE
between_0_and_10 (10) # FALSE
```


lionel-:
Hello,

Just to note that there are potentially many more predicate factories like this (for example a factory for type predicates, `is_object <- is_a("object", "parent")`), so it'd be better to have the functions in a single file such as `factory-predicates.R`

paulhendricks:
@lionel- Sounds good, I'll throw it in there and re-submit the pull request.

hadley:
Hmmm I think the idea is ok, but:

* I don't like the names
* I'd prefer to have less duplication between the functions
* I think they'd be better in one file and documented together

paulhendricks:
@hadley Do you have any preferences/suggestions for naming that are consistent with purr's API?

hadley:
18 months later, I think we should just close this. It's not a bad idea, but it doesn't feel like a good fir for purrr to me.

PR #87, opened 2015-09-11
Rename zip() to zip_n()

lionel-:
And provide shortcuts `zip2()` and `zip3()`.


lionel-:
Added the integer coercion.

lionel-:
by the way, in objects.R we have `is_numeric()` and `is_bare_numeric()` but no `_double` version. To be consistent we should either change the predicates to doubles or change `map_dbl()` to `map_num()`. I think I prefer the latter because the suffix is pronounced as one syllable instead of two.

hadley:
I prefer double because it's less overloaded than numeric (which sometimes means int or double)

lionel-:
voilà

lionel-:
Fixed the conflict and added all the new functions to the README. We come a long way :)

lionel-:
by the way, I know I came up with the idea for purr, but I think I prefer the original name lowliner. It's catchy and it looks serious. The only downside is that it's a bit longer so namespaced calls will be a bit longer too, e.g. `lowliner::map_if(x, p, f)`. But to me this looks much better in my code than `purrr::map_if(x, p, f)`, perhaps in part because all those `rrr` look very blurry to me (but maybe that's just me). Purr is also a word that I find harder to pronounce.

lowliner could refer to the fact that this package is a "better standard library" with underscored function names. But it doesn't need to refer to anything really, I mean, underscore.js doesn't really have a justification (or does it? I'm not aware of any).

hadley:
Thanks!

I'm pretty happy with purrr as a name - I like the play on pure R.

PR #91, opened 2015-09-27
Add flatmap()

lionel-:
And allow `.type` = "numeric" in `can_simplify()`.


R/coercion.R
```diff
@@ -55,21 +55,26 @@ can_simplify <- function(x, type = NULL) {
...
-    type <- typeof(type)
   }
 
-  actual <- typeof(x[[1]])
-  safe_int_coercions <- c("integer", "double", "numeric")
-  if (actual == "integer" && type %in% safe_int_coercions) {
-    TRUE
-  } else {
-    actual == type
+  if (actual == "integer" && type %in% c("integer", "double", "numeric")) {
```
hadley: When does `typeof()` return `numeric`?

R/map-vector.R
```diff
@@ -0,0 +1,37 @@
...
+#' \dontrun{
+#' c(mtcars) %>% flatmap(var_select, .type = "character")
+#' }
+#' c(mtcars) %>% flatmap(var_select, .type = "numeric")
+flatmap <- function(.x, .f, ..., .type) {
+  if (!missing(.type)) {
+    if (!is_scalar_character(.type)) {
+      stop(".type must be a string")
+    }
+    if (!can_simplify(.x, .type)) {
```
hadley: Shouldn't that be called on the results of `map(.x, .f)`?

lionel-: oh yes, sorry

R/coercion.R
lionel-: it's to allow `.type = "numeric"` in `flatmap()` and `as_vector()`.

hadley: Oooo, I see.

R/map-vector.R
lionel-: fixed, and the test now catches this.

R/coercion.R
hadley: Could you add a comment to that effect?

lionel-: here you go

hadley:
Thanks - will submit to CRAN very soon!

lionel-:
cool :)

PR #96, opened 2015-09-29
Typos doc

lionel-:
I did a bit of doc proofreading.

Also `zip_n()` now takes a `.l` argument instead of `.x`, as with `map_n()` and `cross_n()`.


R/composition.R
```diff
@@ -170,10 +168,9 @@ lift_vd <- function(..f, ..., .type) {
 #'
 #' @export
 #' @examples
-#' \dontrun{ %spacing
+#' #
```
lionel-: do you know of a better way to ensure that Roxygen leaves some whitespace when it collates the examples together?

lionel-:
thanks for the kind mention in the announcement :)

hadley:
Could you please merge/rebase?

R/composition.R
hadley: roxygen2 5.0.0 now keeps by default :)

lionel-:
Here you go.

hadley:
Thanks!

PR #101, opened 2015-10-02
fix example: map() does not have .type argument

yutannihilation:


hadley:
Maybe that should be `map_dbl()`?

yutannihilation:
Ah, I got it!

I think this line should be just deleted, since the intention of these examples is to show 3 forms of `map`ing a function.

```r
1:10 %>%
  map(rnorm, n = 10) %>%
  map(mean, .type = numeric(1))

# Or use an anonymous function
1:10 %>%
  map(function(x) rnorm(10, x))

# Or a formula
1:10 %>%
  map(~ rnorm(10, .x))
```

First example should be below, which shows the way specifying a function and its argument:

```r
1:10 %>%
  map(rnorm, n = 10)
```

yutannihilation:
fixed example to use `map_dbl()`.

hadley:
Thanks!

PR #138, opened 2015-11-19
Rows update

lionel-:
This is a refactor of the rows-based functionals.

* Part of the functions are rewritten in C, so that we can use `Rf_eval()` and avoid the overhead of `Rcpp_eval()`. This is done by reusing the C implementations of `map()` and `pmap()`.

* `map_rows()` is renamed to `invoke_rows()`. I think this is a better analogy for that function since it works like applying `invoke()` to each row.

New features:

* Add `.to` option to change the name of the output column, as per @jennybc's request.

* Add a `.collate` option to make the output both more flexible and more predictable. By default, the results are returned in a list-column. If the results are data frames or atomic vectors, you can also opt to collate on rows or on columns.

  This default makes it a bit harder to modify data frames (by opposition to using them for analysis), but this is best done with dplyr anyway and list-collation has the advantage of working in all situations. 

  Though I do wonder if we should have "cols" by default for `invoke_rows()` since it seems popular (see [here](https://github.com/hadley/dplyr/issues/441) or [here](https://twitter.com/mikelove/status/652502458946703360)). It would make this function a more direct equivalent of `plyr::mdply()`.

Collation on cols:
```r
p <- expand.grid(mean = 1:5, sd = seq(0, 1, length = 10))
p %>% invoke_rows(rnorm, n = 5, .collate = "cols")
> Source: local data frame [50 x 7]
>
>     mean        sd    .out1    .out2    .out3     .out4    .out5
>    (int)     (dbl)    (dbl)    (dbl)    (dbl)     (dbl)    (dbl)
> 1      1 0.0000000 1.000000 1.000000 1.000000 1.0000000 1.000000
> 2      2 0.0000000 2.000000 2.000000 2.000000 2.0000000 2.000000
> 3      3 0.0000000 3.000000 3.000000 3.000000 3.0000000 3.000000
> 4      4 0.0000000 4.000000 4.000000 4.000000 4.0000000 4.000000
> 5      5 0.0000000 5.000000 5.000000 5.000000 5.0000000 5.000000
> 6      1 0.1111111 1.042892 1.069705 1.230865 0.9018663 1.131391
> 7      2 0.1111111 2.039779 1.868290 2.164415 1.9447401 1.934308
> 8      3 0.1111111 3.021731 2.801179 3.017455 2.8867998 2.877431
> 9      4 0.1111111 4.085519 4.010198 4.101562 4.0167346 3.985143
> 10     5 0.1111111 4.947370 4.799235 4.882036 5.1118432 4.946957
> ..   ...       ...      ...      ...      ...       ...      ...
```

on rows:
```r
p %>% invoke_rows(rnorm, n = 5, .collate = "rows")
> Source: local data frame [250 x 4]
>
>     mean    sd  .row  .out
>    (int) (dbl) (int) (dbl)
> 1      1     0     1     1
> 2      1     0     1     1
> 3      1     0     1     1
> 4      1     0     1     1
> 5      1     0     1     1
> 6      2     0     2     2
> 7      2     0     2     2
> 8      2     0     2     2
> 9      2     0     2     2
> 10     2     0     2     2
> ..   ...   ...   ...   ...
```

by list:
```r
p %>% invoke_rows(rnorm, n = 5, .collate = "list")
> Source: local data frame [50 x 3]
>
>     mean        sd     .out
>    (int)     (dbl)   (list)
> 1      1 0.0000000 <dbl[5]>
> 2      2 0.0000000 <dbl[5]>
> 3      3 0.0000000 <dbl[5]>
> 4      4 0.0000000 <dbl[5]>
> 5      5 0.0000000 <dbl[5]>
> 6      1 0.1111111 <dbl[5]>
> 7      2 0.1111111 <dbl[5]>
> 8      3 0.1111111 <dbl[5]>
> 9      4 0.1111111 <dbl[5]>
> 10     5 0.1111111 <dbl[5]>
> ..   ...       ...      ...
```


src/rows.cpp
```diff
@@ -120,124 +237,300 @@ SEXP process_slices(List& results, const List& slicers,
...
+  for (int i = 0; i < n_slices; ++i) {
+    out[i] = visitors.subset(indices[i], classes);
   }
 
-  return process_slices(results, data[slicers_names],
-                        data.attr("labels"), include_labels);
+  return out;
 }
 
+extern "C" SEXP by_slice_impl(SEXP env, SEXP d_name_, SEXP f_name_) {
```
lionel-: @kevinushey if you have time could you please take a look?

Is this a good strategy to avoid bypassing C++ cleaning on R exceptions? I call Rcpp functions from a C-like function, both before and after a `Rf_eval()`, because rewriting those parts would be too much work.

I replaced all `stop()` calls by `Rf_error()`, otherwise R crashes on user errors since the C++ exceptions are not caught. Is there any other gotcha to be aware of?

kevinushey: I don't think this is the right way to go. You definitely don't want to call `Rf_error()` from C++ code as that explicitly jumps over C++ destructors (rather than implicitly on error as an `Rf_eval()` could).

Do you absolutely need to use `Rf_eval()` here? If the goal is to just get `d_val` why not use something like `Rf_findVarInFrame()`?

What's the main thing you're trying to do / avoid here?

lionel-: > You definitely don't want to call Rf_error() from C++ code as that explicitly jumps over C++ destructors

ooh right.

> Do you absolutely need to use Rf_eval() here?

It'd be nice to avoid `Rcpp_eval()` because it will be called in a very tight loop. Not to get `d_val` but in the call to `map_impl()` below. These functions are used to map over grouped data frames, so at best you have j groups calls to an R function, and at worst j times n columns.

I think what would work would be to wrap the function with `BEGIN_RCPP` and `END_RCPP`. The exception handling will be bypassed on R functions errors, but that's not a big deal since there won't be any C++ classes instantiated at that point. And then we do get C++ cleanup as usual later on. The main thing is to avoid exporting the functions with Rcpp attributes since they instantiate C++ classes around the input SEXPs.

kevinushey: I think that should be okay (although in theory it's undefined behaviour, in practice as long as you don't have any C++ objects on the stack / pending destructors you should be safe).

Using `BEING_RCPP` / `END_RCPP` macros should be good (it makes sense to wrap any function block that might call Rcpp functions with that).

FWIW, RStudio uses a local `try-catch-throw` block when it needs to pass control to R and `longjmp` from within a C++ context -- see:

https://github.com/rstudio/rstudio/blob/2fb003e1ac493716ba13f7918bc18b71990c01a1/src/cpp/r/session/RSession.cpp#L729

https://github.com/rstudio/rstudio/blob/2fb003e1ac493716ba13f7918bc18b71990c01a1/src/cpp/r/session/RSession.cpp#L770-L783

Ie, all C++ allocation happens within a `try` block; if we need to pass control to R (e.g. on interrupt) we throw an `InterruptException`, that gets caught locally and then we proceed.

lionel-: Great. Thanks a lot for all your help on this problem Kevin.

hadley: Is there a reason not to use `R_tryEval` here?

kevinushey: It uses `R_ToplevelExec()` under the hood which ignores any pre-attached handlers. E.g. something like this would break:

```
withCallingHandlers(
    purrr::invoke_rows(data, function() message("Hello!"))
    message = function(m) {
        print("Caught message!")
    }
)
```

kevinushey: FWIW Rcpp was briefly using `R_tryEval()` and it broke all kinds of things in Shiny so I really don't recommend it.

hadley: @kevinushey ah, got it. That's probably not as important here as for Rcpp, but it's still important

lionel-:
I added support of sliced data frames to `map()`. This way we avoid some R evaluations. Also, since the results of `by_slice()` are now returned in a list-column by default of row-binding the data frames, this allows a shorter syntax:

```{r}
df <- mtcars %>% slice_rows("cyl")

df %>% map(mean)
df %>% by_slice(map, mean, .collate = "rows")
```

Also, the default collation for `invoke_rows()` is now "cols", which makes its output equivalent to that of `plyr::mdply()`.

lionel-:
> I added support of sliced data frames to map()

I wonder if we need a version of `map()` that doesn't handle data frames at all and just return a list in all cases. Something like `map_()`. We could use it internally, for example in `map_df()`. This way,
```r
map(list(1, 2), ~mtcars)
map_df(mtcars[1:2], ~mtcars)
```
would be equivalent.

Or should data frame mapping go into another function? Could be `df_map()`? Prefix refers to input, suffix to output. Though I think I prefer `map_()`.

R/rows.R
```diff
@@ -65,21 +74,36 @@ by_slice <- function(.d, ..f, ..., .labels = TRUE) {
...
-    calling_env <- parent.frame()
-    by_slice_impl(.d, ..f, dots, .labels, calling_env)
+    stop(".d must be a sliced data frame", call. = FALSE)
+  }
+  if (!.collate %in% c("list", "rows", "cols")) {
+    stop(".collate should be `list`, `rows` or `cols`", call. = FALSE)
   }
+
+  set_sliced_env(.d, .labels, .collate, .to, environment(), ".d")
+  .Call("by_slice_impl", environment(), ".d", "..f")
```
kevinushey: You should set the `PACKAGE` argument here; otherwise R is forced to lookup the symbol by scanning through all loaded libraries (and this can take a non-trivial amount of time if many libraries are loaded).

R/rows.R
```diff
@@ -117,26 +149,61 @@ by_slice <- function(.d, ..f, ..., .labels = TRUE) {
...
+  if (!is.data.frame(.d)) {
+    stop(".d must be a data frame", call. = FALSE)
+  }
+
+  attr(.d, "indices") <- as.list(seq_along(.d[[1]]))
+  .unique_labels <- 0
+  .labels_cols <- .d
+  .slicing_cols <- .d
+
+  .Call("by_slice_impl", environment(), ".d", "..f")
```
kevinushey: Same `PACKAGE` argument here.

R/rows.R
```diff
@@ -117,26 +149,61 @@ by_slice <- function(.d, ..f, ..., .labels = TRUE) {
...
   if (inherits(..f, "formula")) {
     ..f <- as_function(..f)
   } else if (!is.function(..f)) {
     stop("..f should be a function or a formula", call. = FALSE)
   }
 
-  dots <- substitute(list(...))
-  calling_env <- parent.frame()
-  by_row_impl(.d, ..f, dots, .labels, calling_env)
+  if (!.collate %in% c("list", "rows", "cols")) {
```
kevinushey: Can this be handled with `match.arg()`?

src/fast-copy.cpp
```diff
@@ -1,123 +1,108 @@
...
-    CTYPE* out_ptr = ACCESSOR(out);                    \
-    for (int i = 0; i < x_size; ++i) {                 \
-      for (int j = 0; j < times[i]; ++j) {             \
-        out_ptr[counter] = x_ptr[i];                   \
-        ++counter;                                     \
-      }                                                \
-    }                                                  \
-    return out;                                        \
-    break;                                             \
+#define DO_REP_EACH_N(RTYPE, CTYPE, ACCESSOR)           \
```
kevinushey: Are the changes here just whitespace? (should try to avoid whitespace-only changes since they can make pull requests harder to review)

src/rows-data.cpp
```diff
@@ -0,0 +1,121 @@
+#include <Rcpp.h>
+#include <dplyr.h>
+#include "utils.h"
+#include "rows-data.h"
+
+namespace Slices {
+
+
+CollationType hash_collate(std::string collate) {
```
kevinushey: Pass by `const std::string&`?

src/rows-data.cpp
```diff
@@ -0,0 +1,121 @@
...
+
+
+Labels::Labels(Environment execution_env_)
+    : are_unique(execution_env_[".unique_labels"]),
+      slicing_cols(execution_env_[".slicing_cols"]),
+      labels_(execution_env_[".labels_cols"]),
+      n_labels_(Rf_length(execution_env_[".labels_cols"])) {
+}
+
+void Labels::remove(std::vector<int>& index) {
```
kevinushey: Could this be `const std::vector<int>& index`? I think it's slightly preferred stylistically to use const-ref for objects that are read-only, and pass by pointer if the object will be modified.

This makes it very clear at the call site that an object could be mutated, eg:

```cpp
std::vector<int> x;
fill_vector(&x);
```

In the above call, one can infer that the `x` object is modified by `fill_vector()`; it is somewhat less obvious if that function were to accept just a plain reference.

lionel-:
Thanks a lot for your comments Kevin :)

src/rows-data.cpp
lionel-: I like the idea of making the syntax reveal mutation at the call site, but this does not seem very conventional. See for example https://isocpp.org/wiki/faq/references#refs-vs-ptrs

> Old line C programmers sometimes don’t like references since they provide reference semantics that isn’t explicit in the caller’s code. After some C++ experience, however, one quickly realizes this is a form of information hiding, which is an asset rather than a liability. E.g., programmers should write code in the language of the problem rather than the language of the machine.

Plus I feel like I would be guilty not checking for NULL pointers. And it does make for a heavier syntax with all the dereferencing.

How about using references but with this form:
```C++
obj& modify(obj&) { return obj };

void some_function() {
  obj = modify(obj);
}
```

Then it does not matter whether we have value or reference semantics, it becomes an implementation detail.

R/flatten.R
```diff
@@ -24,36 +24,36 @@
 #' x %>% map_int(1L)
 #' @useDynLib purrr flatten_impl vflatten_impl
 flatten <- function(.x) {
-  .Call(flatten_impl, .x)
+  .Call(flatten_impl, .x, PACKAGE = "purrr")
```
hadley: I think @kevinushey misread this - I'm using `flatten_impl` not "`flatten_impl`" so it doesn't need the package argument.

lionel-: hmm no, his comments were about my function calls and they do use strings. My bad.

lionel-:
I have second doubts about this column collation in `invoke_rows()`. List collation would be more standard with other functions and allows nice patterns like this:
```r
population <- strata_info %>%
  select(n = size, mean) %>%
  invoke_rows(.f = rnorm, sd = sigma) %>%
  tidyr::unnest(.out)
```
I'm finally getting the hang out of `nest()` and `unnest()` :)

hadley:
Also:

```R
p <- expand.grid(mean = 1:5, sd = seq(0, 1, length = 10))
p %>% invoke_rows(rnorm, n = 5, .collate = "list")
```

is pretty similar to:

```R
expand.grid(mean = 1:5, sd = seq(0, 1, length = 10)) %>% 
  transpose() %>% 
  invoke_map("rnorm", ., n = 5)
```

lionel-:
yes I'll change the default to "list".

About `by_row()`, I've never found a use for it until now. With `invoke_rows()` you need to subset the data frame first so that the names match the function signature. With `by_row()` you can select the parameters via the lambda placeholder, keeping all other variables as contextual information in the final result:
```r
population <- strata_info %>%
  by_row(~rnorm(.$size, .$mean, sigma), .to = "x") %>%
  tidyr::unnest(x)
```

lionel-:
Moved out all data frame features from `map()` to `dmap()`

hadley:
Thanks!

lionel-:
@hadley I was thinking of renaming `dmap()` to `by_col()` lately. But I'm not sure I like the variants `by_col_if()` or `by_col_at()` very much. So that got me thinking that we could handle conditional mapping through composition rather than verbal variants. i.e.:

```r
l %>% where(is_character, map, as_numeric)
mtcars %>% at(needs_shifting, by_col, shift_to_zero)
```

If we do that, we need quite a bit of specific handling in the `where()` and `at()` adverbs though:

- Allowing regions to grow and shrink so they can be used with `lmap()`.
- Passing relevant grouping/slicing attributes to subsets of data frames.
- Recycling within slices probably shouldn't be handled within these adverbs, so `by_col()` would need a `.recycle` argument. That would make its output a bit more predictable which is never a bad thing.

If you like the adverb idea, I think it can wait for the next version—just let me know if I should open an issue. And if you like the `by_col()` idea, I'll rename the verbs right away.

hadley:
@lionel- lets table this for a while. I think I'll have a better idea once I've been working on dplyr for a bit.

Issue #46, opened 2015-01-27
Look into implementing Haskell's scanl


Paradosso:
(Reminder for @hadley following a twitter conversation)

From scanl's [documentation](http://zvon.org/other/haskell/Outputprelude/scanl_f.html): It takes the second argument and the first item of the list and applies the function to them, then feeds the function with this result and the second argument and so on. It returns the list of intermediate and final results.

----

It is like fold except it also returns the intermediate results.

hadley:
Maybe call it `creduce()`? Because it's an accumulating reduce?

Paradosso:
Pro: clear that it's (a) cumulative (b) higher-order (can specify your own reducer). Con: the function does not really "reduce" a list/vector to one value, but that's maybe subjective.

lionel-:
How about simply `accumulate()` or maybe `acc()` for short?

PR #145, opened 2015-12-05
Implemented accumulate for recursive folding

michaelquinn32:
Following up on #46 and #11, I implemented a version of `accumulate`, which adds recursive folding to the package. 

The differences between `accumulate` and `reduce` are minor. I simply added another argument (`accumulate = TRUE`) to the call of `Reduce`. Nonetheless, this change provides a lot of utility. 

I have two examples in the documentation. The first shows flooring and capping a dataset, which is an extension of an example in [Haskell's scanl documentation](http://zvon.org/other/haskell/Outputprelude/scanl_f.html). The second example shows a simple simulation of a random walk with drift.


R/reduce.R
```diff
@@ -46,3 +46,54 @@ reduce_right <- function(.x, .f, ..., .init) {
...
+#'   map2(caps, ~ accumulate(.x, min, .init = .y[2]))
+#'
+#' # Simulating stochastic processes with drift
+#' \dontrun{
+#' library(dplyr)
+#' library(ggplot2)
+#'
+#' rerun(5, rnorm(100)) %>% setNames(paste0("sim", 1:5)) %>%
+#'   map(~ accumulate(., function(.x, .y) .05 + .x + .y)) %>%
+#'   map(~ data.frame(value = .x, step = 1:100)) %>%
```
hadley: This could now be `map_df(~ data_frame(value = .x, step = 1:100), .id = "simulation")`

R/reduce.R
```diff
@@ -46,3 +46,54 @@ reduce_right <- function(.x, .f, ..., .init) {
...
+#'   bind_rows(. , .id = "simulation") %>%
+#'   ggplot(., aes(x = step, y = value)) + geom_line(aes(color = simulation))
```
hadley: You don't need the explicit `.` here

R/reduce.R
```diff
@@ -46,3 +46,54 @@ reduce_right <- function(.x, .f, ..., .init) {
...
+#' caps <- mtcars %>% map(~ quantile(.x, probs = c(.1, .9), names = FALSE))
+#' floor_cap <- mtcars %>% map2(caps, ~ accumulate(.x, max, .init = .y[1])) %>%
+#'   map2(caps, ~ accumulate(.x, min, .init = .y[2]))
+#'
+#' # Simulating stochastic processes with drift
+#' \dontrun{
+#' library(dplyr)
+#' library(ggplot2)
+#'
+#' rerun(5, rnorm(100)) %>% setNames(paste0("sim", 1:5)) %>%
```
hadley: Can you put this on a new line and use `set_names()`?

R/reduce.R
```diff
@@ -46,3 +46,54 @@ reduce_right <- function(.x, .f, ..., .init) {
...
+#'   map(~ accumulate(., function(.x, .y) .05 + .x + .y)) %>%
+#'   map(~ data.frame(value = .x, step = 1:100)) %>%
+#'   bind_rows(. , .id = "simulation") %>%
+#'   ggplot(., aes(x = step, y = value)) + geom_line(aes(color = simulation))
+#' }
+accumulate <- function(.x, .f, ..., .init) {
+  force(.f)
```
hadley: I think you should use `as_function()` here to be consistent with the rest of purrr.

hadley:
Looks good - thanks!  Can you please respond to my super nitpicky comments and add a bullet to NEWS?

michaelquinn32:
Super nitpicky is super helpful! I didn't notice the addition of `map_df`. That's a great little tool. 

I also did a little additional formatting to the examples. Hopefully they'll be a little easier to read.

Please let me know if I missed anything.

R/reduce.R
michaelquinn32: I kept the syntax for `accumulate()` as close as possible to that of `reduce()`. Does it make sense to use `as_function()` in one but not the other? Or should I just change both?

hadley:
Looks great - thanks!

PR #155, opened 2015-12-26
test recycle_args and fix apparent bug?

davharris:
This package looks fantastic.  I decided to poke around with some of the code to learn more about functional programming in R and I found some behavior that should probably be changed.

In the master branch, the use of `||` in `recycle_args` leads R to only check the first element of `lengths`. I changed `||` to `|` and added some tests.  Hope this is useful.


hadley:
Nice - thanks!

PR #190, opened 2016-04-13
Fixed a typo in the man page for partial.RD

burchill:
Under "Examples", it gave the code as: `compact2 <- partial(discard, .f = is.null) `
but the named argument for `discard` which `partial` is filling is not `.f`, but `.p`.  
Running the function `compact2` (e.g. `compact2(c(1,2,NULL,4)`) as the code was previously produces an error.

### DIFF
--- a/R/partial.R
+++ b/R/partial.R
@@ -30,7 +30,7 @@
 #' compact1 <- function(x) discard(x, is.null)
 #'
 #' # we can write:
-#' compact2 <- partial(discard, .f = is.null)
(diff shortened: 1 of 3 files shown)


lionel-:
Thanks @burchill 

Could you update this in the .R file and update the .Rd file with roxygen? See http://r-pkgs.had.co.nz/man.html for more information.

burchill:
Oops, gotcha. Done.

lionel-:
Thanks!

PR #195, opened 2016-05-05
Pass ... to function in `accumulate`

### LINKED ISSUE #188: accumulate does not pass extra parameters
```
> x <- c(1, 2, NA, 4, 5, NA)
> x %>% accumulate(sum)
[1]  1  3 NA NA NA NA
> x %>% accumulate(sum, na.rm = TRUE)
[1]  1  3 NA NA NA NA
> mysum <- function(a, b) sum(a, b, na.rm = TRUE)
> x %>% accumulate(mysum)
[1]  1  3  3  7 12 12
```
this may be related to `Reduce` not having the ...

```
> Reduce(sum, x, accumulate = TRUE, na.rm = TRUE)
Error in Reduce(sum, x, accumulate = TRUE, na.rm = TRUE) : 
  unused argument (na.rm = TRUE)
```

vspinu:
Fixes #188

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # purrr 0.2.1.9000
 
+* `accumulate` and `reduce` are now correctly passing extra arguments to the worker function
+
 # purrr 0.2.1
 
 * `as_function()` gains a `.null` argument that for character and numeric
--- a/R/reduce.R
+++ b/R/reduce.R
@@ -29,7 +29,12 @@
 #' list() %>% reduce(`+`)
 #' list() %>% reduce(`+`, .init = 0)
 reduce <- function(.x, .f, ..., .init) {
-  f <- as_function(.f, ...)
+  .f <- as_function(.f, ...)
+
+  f <- function(x, y) {
+    .f(x, y, ...)
+  }
+
   Reduce(f, .x, init = .init)
 }
 
@@ -75,7 +80,12 @@ reduce_right <- function(.x, .f, ..., .init) {
 #'     ggtitle("Simulations of a random walk with drift")
 #' }
 accumulate <- function(.x, .f, ..., .init) {
-  f <- as_function(.f, ...)
+  .f <- as_function(.f, ...)
+
+  f <- function(x, y) {
+    .f(x, y, ...)
+  }
+
   Reduce(f, .x, init = .init, accumulate = TRUE)
 }
 
--- a/tests/testthat/test-reduce.R
+++ b/tests/testthat/test-reduce.R
@@ -1,0 +1,7 @@
+context("reduce")
+
+test_that("accumulate passes arguments to function", {
+  tt <- c("a", "b", "c")
+  expect_equal(accumulate(tt, paste, sep = "."), c("a", "a.b", "a.b.c"))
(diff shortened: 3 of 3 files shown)


lionel-:
I think @hadley had something in mind with indexing into lists when he added `...` to `as_function()` in `reduce()` and `accumulate()`. But it's not clear at all what the semantics should be here.

E.g. with this:

```{r}
x <- list(a = list(a = list(a = 1, b = 2), b = 3), NULL, NULL)
reduce(x, "a") %>% str()
accumulate(x, "a") %>% str()
```

Indexes "a" recursively, but the depth at which we index depends on the size of the outer list `x`. That doesn't seem really helpful. E.g. this will fail because this will index too deeply:

```{r}
x2 <- list(a = list(a = list(a = 1, b = 2), b = 3), NULL, NULL, NULL)
reduce(x2, "a")
accumulate(x2, "a") %>% str()
```

Also these mappers normally take binary functions but indexing is unary (we index with `[[`), so it doesn't seem fitting here. But I see that `...` is actually not used for anything in `as_function()` so it might be unfinished.

By the way, if you change `accumulate()` you should change `reduce()` as well. But let's see what Hadley had in mind with these.

tests/testthat/test-reduce.R
```diff
@@ -0,0 +1,9 @@
+context("reduce")
+
+test_that("accumulate passes arguments to function", {
+    tt <- c("a", "b", "c")
+    expect_equal(accumulate(tt, paste, sep = "."),
+                 c("a", "a.b", "a.b.c"))
```
lionel-: We use the `RStudio-` indent style ;)
Or these could go on one line

vspinu:
> Also these mappers normally take binary functions but indexing is unary (we index with [[), 

AFAIC, the semantics is very clear. Unary functions should not be used with reducers. While indexing is also technically a binary function, heterogenuous arguments and arbitrary return value is not something that you would want to compose on a list or even to reason about. 

> x <- list(a = list(a = list(a = 1, b = 2), b = 3), NULL, NULL)
> reduce(x, "a") %>% str()
> accumulate(x, "a") %>% str()

This is quite a mind bending stretch.

hadley:
Yeah, that makes sense to me.

R/reduce.R
```diff
@@ -75,14 +75,17 @@ reduce_right <- function(.x, .f, ..., .init) {
 #'     ggtitle("Simulations of a random walk with drift")
 #' }
 accumulate <- function(.x, .f, ..., .init) {
-  f <- as_function(.f, ...)
+  .f <- as_function(.f)
```
hadley: I think you should add a comment here, briefly describing the reason why `as_function` doesn't take `...` here (unlike every other function).  

Equally, you could leave it, since adding it does no harm.

NEWS.md
```diff
@@ -1,5 +1,9 @@
 # purrr 0.2.1.9000
 
+## Bug Fixes
```
hadley: Please don't add a heading in the NEWS - I do that just prior to release since it's not clear in advance whether or not there will be enough items to warrant subheadings (and what they should be)

lionel-:
> But I see that ... is actually not used for anything in as_function() so it might be unfinished.

oops, I was browsing an outdated purrr codebase, never mind!

hadley:
Thanks!

PR #202, opened 2016-06-09
Add .env argument to invoke() family

lionel-:
For hadley/modelr#9

### DIFF
--- a/R/invoke.R
+++ b/R/invoke.R
@@ -13,6 +13,10 @@
 #'   same length as \code{.f}, and will call each function with no
 #'   arguments (apart from any suppled in \code{...}.
 #' @param ... Additional arguments passed to each function.
+#' @param .env Environment in which \code{\link{do.call}()} should
+#'   evaluate a constructed expression. This only matters if you pass
+#'   as \code{.f} the name of a function rather than its value, or as
+#'   \code{.x} symbols of objects rather than their values.
 #' @inheritParams map
 #' @export
 #' @examples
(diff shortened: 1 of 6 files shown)


hadley:
Hmmmm, I think a slightly better default would be to use the environment of the promise `.f`, but I'm not sure if we want to add a dep on lazyeval.

lionel-:
It's already an indirect dependence though.

hadley:
Although we might remove the dplyr dep, right?

lionel-:
The problem is that we have things like `map_df()`. Do you want to move these in dplyr? It doesn't seem like a good fit.

hadley:
Ah true - I think in the long-long term the code for bind rows etc might move into tibble (or a general coercion package), but in the medium-term we'll definitely need to keep the dplyr dep.   So I think that implies for this case it would be better to use lazyeval.

lionel-:
Would it be ok to add `lazy_env()` to lazyeval? This way the function signature would be:

```r
function(.f, .x = NULL, ..., .env = lazyeval::lazy_env(.f))
```

Or do you prefer not having the `.env` parameter?

lionel-:
we can just have `function(.f, .x = NULL, ..., .env = NULL)` and then `.env <- .env %||% lazyeval::lazy(.f)$env` though

hadley:
1. I think in the new lazy api it would be `expr_env()`

1. I think it will need to have a second argument that should be returned if the promise has already been evaluated. 

1. If we're using the environment of the promise, I don't think env needs to be a parameter anymore?

lionel-:
> If we're using the environment of the promise, I don't think env needs to be a parameter anymore?

right it would only be useful in a handful of cases. We can always add it later if find a need. But... If the promise has been forced we actually still need the `parent.frame()` env, which complicates everything :s

I don't see how we can make this work then.

hadley:
I'm imagining you'd do `expr_env(.f, parent.frame())`

lionel-:
ah good idea. But this needs to be in signatures so that derived invoke functions like `invoke_map()` can propagate the correct env.

So in this case is it ok to merge this PR, and when we have the new lazyeval API we can change `.env = parent.frame()` to `.env = expr_env(.f, parent.frame())` in all signatures?

hadley:
Yeah, agreed

PR #210, opened 2016-06-30
document each of the output-catching adverbs

tjmahr:
Expanded documentation to say explicitly say what safely/possibly/quietly do.


hadley:
Thanks!

PR #226, opened 2016-08-18
Deprecate dataframe-based mapping functions

lionel-:
Deprecation of rows-based functions now that we have the new mapping functions in dplyr and the nest() / mutate() / map() idiom.

One use case that does not have any equivalent in dplyr is when the output is within the range `c(1, nrow(df))`. For example:

```{r}
dmap(mtcars, summary)
```

This can be worked around, for example here we'd do

```{r}
mtcars %>% map(summary) %>% as_data_frame()
```

However a new dplyr family of verbs for variable-length output may be useful. It could be called `condense()`.

- Like `summarise()` it would discard all input columns except for the grouping variables. This allows the output to have a different number of rows than the input.

- Unlike `summarise()`, it would not require length 1 results and would only check for equal length within group. Grouping columns would be recycled to these lengths.

Ungrouped data frame: check squared constraint

```{r}
mtcars %>% condense(col = 1:5, other = 5:1)
#> # A tibble: 5 x 2
#>     col other
#>   <int> <int>
#> 1     1     5
#> 2     2     4
#> 3     3     3
#> 4     4     2
#> 5     5     1

mtcars %>% condense(col = 1:5, other = 2:1)
#> Error: results must have same length
```

This gives us immediately:

```{r}
mtcars %>% condense_all(summary)
#> # A tibble: 6 x 11
#>           mpg         cyl        disp          hp        drat
#>   <S3: table> <S3: table> <S3: table> <S3: table> <S3: table>
#> 1        10.4        4.00        71.1        52.0        2.76
#> 2        15.4        4.00       121.0        96.5        3.08
#> 3        19.2        6.00       196.0       123.0        3.70
#> 4        20.1        6.19       231.0       147.0        3.60
#> 5        22.8        8.00       326.0       180.0        3.92
#> 6        33.9        8.00       472.0       335.0        4.93
#> # ... with 6 more variables: wt <S3: table>, qsec <S3: table>, vs <S3:
#> #   table>, am <S3: table>, gear <S3: table>, carb <S3: table>
```

For a grouped data frame, we'd check the square constrain within groups:

```{r}
grouped <- mtcars %>% group_by(am)

grouped %>%
  condense(
    col = rep(mean(cyl), times = round(mean(cyl))),
    other = rep(length(col), length(col))
  )
#> # A tibble: 12 x 3
#>       am   col other
#>    <dbl> <dbl> <dbl>
#> 1      0  6.95     7
#> 2      0  6.95     7
#> 3      0  6.95     7
#> 4      0  6.95     7
#> 5      0  6.95     7
#> 6      0  6.95     7
#> 7      0  6.95     7
#> 8      1  5.08     5
#> 9      1  5.08     5
#> 10     1  5.08     5
#> 11     1  5.08     5
#> 12     1  5.08     5

grouped %>%
  condense(
    col = rep(mean(cyl), times = round(mean(cyl))),
    other = rep(length(col), length(col) - 1)
  )
#> Error: results must have same length within groups
```

Relevant discussion: hadley/dplyr#154

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,29 @@
 # purrr 0.2.2.9000
 
+* All data-frame based mappers have been deprecated in favour of new
+  functions and idioms in the tidyverse.
+
+    * Mapping a function to each column of a data frame should now be
+      handled with the colwise mutating and summarising operations in
+      dplyr instead of `dmap()`. These are the verbs with suffix
+      `_all()`, `_at()` and `_if()`, such as `mutate_all()` or
+      `summarise_if()`. Note that this means the output of `.f` should
+      conform to the requirements of dplyr operations: same length as
+      the input for mutating operations, and length 1 for summarising
+      operations.
+
+    * Inovking a function row by row with the columns of a data frame
+      as arguments should be done with `pmap()` followed by
+      `dplyr::as_dataframe()` instead of `map_rows()`.
+
+    * Mapping rowwise slices of a data frame with `by_row()` is
+      deprecated in favour of a combination of tidyverse functions.
+      First use `tidyr::nest()` to create a list-column containing
+      groupwise data frames. Then use `dplyr::mutate()` to operate on
+      this list-column. Typically you will want to apply a function on
+      each element (nested data frame) of this list-column with
+      `purrr::map()`.
+
 * `cross_n()` has been renamed to `cross()`. The `_n` suffix was
   removed for consistency with `pmap()` (originally called `map_n()`
   at the start of the project) and `transpose()` (originally called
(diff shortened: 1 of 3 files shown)


lionel-:
Could also have `disperse()` verb that would be like `condense()` but spreads the result over numbered columns, thus resulting in 1 row per group like `summarise()`? That would be the equivalent to `.collate = "cols"` in purrr df functions.

With that verb the lengths can be different across results but must be the same across groups.

hadley:
Can you please add a bullet to news, and move the dplyr stuff to a dplyr issue?  It would also be useful to comment on `dmap(iris, summary)` - I'm not sure if `summary()` is key to your examples or not.

PR #232, opened 2016-08-29
Replace actual NULL by .null

### LINKED ISSUE #231: Allow map_* to treat NULLs as NAs
When working with sparse nested lists (like JSON), it is common to have missing keys or NULL values, which are difficult to coerce into a desired type with purrr.

For example:

```{r}
list(list("a" = 1L), list("b" = 2L)) %>% map_int("a")
#> Error: Result 2 is not a length 1 atomic vector
```

We can achieve the desired effect by mapping twice and using the `null-replace` operator

```{r}
> list(list("a" = 1L), list("b" = 2L)) %>% map("a") %>% map_int(`%||%`, NA)
#> [1]  1 NA
```

Ideally, `map_*` and `flatten_*` would treat `NULL`s as `NA`s, possibly by default or if not then by an argument. It could throw a warning when this occurs.

The options I see are:

1. Throw an error
1. Return a vector that excludes the `NULL` entries
1. Return a vector with appropriate `NA`s where `NULL`s were encountered

Currently, purrr implements option 1, which will lead to difficult to debug failures when working with JSON, and a lot of duplicated code to prevent it. I believe option 2 is very dangerous (this is what `unlist` does). Hence the suggestion for option 3.

lionel-:
Closes #231 

Not sure if that was your intent not to replace actual NULLs @hadley

### DIFF
--- a/src/extract.c
+++ b/src/extract.c
@@ -103,5 +103,8 @@ SEXP extract_impl(SEXP x, SEXP index, SEXP missing) {
     }
   }
 
-  return x;
+  if (TYPEOF(x) == NILSXP)
+    return missing;
+  else
+    return x;
 }
--- a/tests/testthat/test-as-function.R
+++ b/tests/testthat/test-as-function.R
@@ -11,6 +11,15 @@ test_that(".null replaces absent values", {
   expect_equal(map_dbl(x, "b", .null = NA), c(2, NA, NA))
 })
 
+test_that(".null replaces actual NULL values", {
+  x <- list(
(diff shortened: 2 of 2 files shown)


hadley:
Ooops, yes!

hadley:
Ooops, can you please add a bullet to NEWS too?

lionel-:
oh yes, now done.

I keep forgetting about changing the NEWS files. Could Codecov or some Travis bot give a warning when the NEWS files is unchanged in a PR?

PR #247, opened 2016-09-20
Remove LinkingTo: dependency from dplyr

krlmlr:
@lionel-: Looks like you contributed the slicing code. I suspect that the R version might run slightly slower (but not too much). We could implement a lightweight version of row subsetting in this package, but only if the performance is much worse now.

### DIFF
--- a/src/rows-data.cpp
+++ b/src/rows-data.cpp
@@ -30,25 +30,19 @@ Labels::Labels(Environment execution_env_)
 }
 
 void Labels::remove(const std::vector<int>& to_remove) {
-  if (to_remove.size()) {
-    List labels = labels_; // Workaround GCC -O2 crash
+  if (!to_remove.size())
+    return;
 
-    int n = Rf_length(labels[0]);
-    std::vector<int> index(n - to_remove.size());
+  // http://stackoverflow.com/a/22833346/946850
+  static Function subset("[.data.frame");
 
-    int i = 0;
-    int j = 0;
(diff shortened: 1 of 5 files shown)


lionel-:
no worries, these functions are going to be deprecated in favour of dplyr's new colwise family. I don't think the performance should be impacted much, tidyr uses a R version of this as well for nesting/unnesting IIRC.

lionel-:
Thanks!

krlmlr:
Thanks. When do you plan a CRAN release?

lionel-:
@hadley would know :)

hadley:
Not in the near future. Maybe after @lionel- has added support for progress bars and parallel computing ;)

krlmlr:
Should be before the next dplyr release, though. Also, I'd like to get rid of `inst/include` in dplyr as soon as possible, and this means that CRAN purrr can't be installed anymore if a dev version of dplyr is installed.

hadley:
Oh got it, we can do an earlier release if needed.

src/rows-data.cpp
```diff
@@ -30,25 +29,19 @@ Labels::Labels(Environment execution_env_)
...
-        ++i;
-        ++j;
-      }
-      *it = j++;
-    }
+  if (!to_remove.size())
+    return;
+
+  // http://stackoverflow.com/a/22833346/946850
+  static Function subset("[.data.frame");
```
krlmlr: Perhaps dplyr::slice() is better here.

PR #272, opened 2016-12-10
More documentation on recursive indexing; fixes #224

### LINKED ISSUE #224: Better documentation of recursive indexing
- you can use a list
- how `.null` works

jennybc:
This causes a Warning in R CMD check!

```
Documented arguments not in \usage in documentation object 'map':
  ‘.null’
```

I hope there is some magic I don't know about for inheriting a parameter from an S3 method (`as_function.character()`) in this case.

It just feels important to document `.null` in `map()`, so I did it with brute force.


hadley:
I think we should properly fix this by making it clear that `.null` is passed on to `as_function()` and  hence `extract()`. i.e. `.null` should become an explicit argument to the map functions.

hadley:
Alternatively (and this would be less work), `map_` should link to `as_function()` which can explain in more detail there.

R/map.R
```diff
@@ -26,7 +28,9 @@
 #'   \code{walk()} (invisibly) the input \code{.x}. It's called primarily for
 #'   its side effects, but this makes it easier to combine in a pipe.
 #' @seealso \code{\link{map2}()} and \code{\link{pmap}()} to map over multiple
-#'   inputs simulatenously
+#'   inputs simultaneously. \code{\link{as_function}()} for passing additional
+#'   arguments to \code{.f} via \code{...}, such as \code{.null} when \code{.f}
+#'   is a character or integer extractor.
```
jennybc: > Alternatively (and this would be less work), map_ should link to as_function() which can explain in more detail there.

Depending on what you mean, I already did that.

hadley:
I merged this manually since I converted to markdown in the meantime.

lionel-:
`as_function()` now lives in rlang as lambda creation is a general operation. Not sure yet how to articulate the extraction methods with `as_function()`. Maybe purrr should have `as_mapper()`, and forward the function methods to rlang. Also I feel like both `as_function()` and `as_mapper()` should use switchpatch rather than an exported generic.

PR #282, opened 2017-01-13
Update compact function

nabilabd:
This allows for more flexible function specification in the arguments. Otherwise, I think it was a typo to define `.f` and then not use it.


hadley:
Could you please add a bullet point to NEWS briefly describing the change?

nabilabd:
Done, in [#290](https://github.com/hadley/purrr/pull/290/commits/f5393338ad1234213aa01ec07e7ee77099230d16)

PR #288, opened 2017-02-21
Add argument to is_formula() to test one- or two-sidedness

jrnold:
Add an argument to is_formula to additionally check whether the formula is one- or two-sided.

It was a function or argument that I was expecting, but didn't find, since testing whether a formula is one-sided or two-sided is pretty common, and even though the test is easy, checking the length is rather unclear.  

I don't know if this belongs with *lazyeval* or as separate functions, but I thought I'd open up this pull request to see what you thought.

### DIFF
--- a/R/objects.R
+++ b/R/objects.R
@@ -293,8 +293,28 @@ is_empty <- function(x) length(x) == 0
 #' Is a formula?
 #'
 #' @inheritParams is_empty
+#' @param has_lhs If \code{TRUE}, tests that the formula
+#'   is two-sided. If \code{FALSE}, tests that the
+#'   formula is one sided. If \code{NULL} or \code{NA},
+#'   does not test for the sidedness of the formula.
 #' @export
 #' @examples
 #' x <- disp ~ am
 #' is_formula(x)
-is_formula <- function(x) inherits(x, "formula")
+#' is_formula(x, has_lhs = TRUE)
+#' is_formula(x, has_lhs = FALSE)
+#' y <- ~ am
+#' is_formula(y)
+#' is_formula(y, has_lhs = TRUE)
+#' is_formula(y, has_lhs = FALSE)
+is_formula <- function(x, has_lhs = NULL) {
+  f <- inherits(x, "formula")
+  if (f && !is.null(has_lhs) && !is.na(has_lhs)) {
+    f <- if (has_lhs) {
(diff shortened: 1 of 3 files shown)


hadley:
rlang would now be a better for this if @lionel- hasn't implemented it already.

lionel-:
I was thinking `is_fquote()` for checking one-sided formulas. And `is_tquote()` would check that it also has a valid environment. (e.g. formula-quotes and tidy-quotes)

jrnold:
The names `is_fquote()` and `is_tquote()` are definitely not the names for "least surprise" 😄  But I think I understand where you're coming from after thinking about it a bit and how the tidyverse as evolved. The one sided formula is best thought of as a "smarter" quoting function, because it can also keep the environment...so it's a quosure.

lionel-:
I like the name quosure, good job :)

hadley:
I like quosure too, although I'll need to sleep on it to consider if it's too cute.

jrnold:
quosure it way too cute; it probably deserves its own anime mascot.

lionel-:
it makes sense for `~` the quosure to be the tidyverse mascot

jrnold:
The mascot practically draws itself since `~` can be the mouth

lionel-:
and here is quosure: `"\u2368"`

# ⍨

jrnold:
So will there be other functions to do this, or is adding the argument too simple to be worth adding?

hadley:
It should live in rlang so can you move discussion there please?

jrnold:
I didn't even know rlang existed! I wasn't quite sure what you were talking about, but since it sounded like R language I didn't think it was a package. My bad.

lionel-:
@jrnold there is now `rlang::is_quosure()`.

I wonder if `rlang::is_formula()` should only return `TRUE` for two-sided formulas. It makes sense for a formula to be two-sided since the mathematical notion of formula is about relationships. So we'd have formulas and quosures as disjoint sets of objects.

jrnold:
@lionel- I see you ran with quosure stuff :-)

lionel-:
yes it just makes sense, thanks for the suggestion!

jrnold:
On the `is_formula` question, I think that that behavior would confuse a lot of people since "formula" is already a used term in the language.

PR #303, opened 2017-03-13
Simplify negate() implementation

lionel-:
I think the general strategy should look like this. For primitive functions, I wonder if `as_function()` should create a closure wrapper around them? E.g.

```r
as_function(`-`)
#> function(e1, e2) `-`(e1, e2)
```

This makes sure argument matching behaves properly.


lionel-:
@hadley I think `as_mapper.default()` should forward to `rlang::as_function()`. Does that sound reasonable to you? Then we'd move the primitive handling there.

R/negate.R
```diff
@@ -14,33 +14,11 @@
...
-    args <- alist(x = )
-    body <- lang_interp(extract(x, UQ(as.list(f)), default = UQ(default)))
-  } else {
-    stop("Unknown input", call. = FALSE)
-  }
-
-  new_fn(args, trans(body), f_env(quo_f))
-}
+  body(.p) <- tidy_interp(quote({
+    ! { !!! body(.p) }
```
hadley: Shouldn't this be `()`?

hadley:
Ah, this is much simpler!  I agree that `as_mapper.default()` should just call `rlang::as_function()`

R/negate.R
lionel-: oh right, the body will include the curlies if it has several statements

lionel-:
`as_mapper()` now forwards to `as_function()` for both formulas and functions. I wonder if both of these should be turned to switchpatches though?

hadley:
Switchpatch doesn't feel right to me here. `as_function()` could do switchpatch, but I don't think `as_mapper()` should.

PR #304, opened 2017-03-14
First attempt at auto_browse

hadley:


R/output.R
```diff
@@ -40,6 +40,21 @@
...
+#'   y <- 20
+#'   if (x > 5) {
+#'     stop("!")
+#'   } else {
+#'     x
+#'   }
+#' }
+#' map(1:6, auto_browse(f))
+#'
+#' # Unfortunately you can't use it with primitive functions like log.
```
lionel-: hmm now that primitive functions are wrapped in a closure I think that should work :)

R/output.R
```diff
@@ -40,6 +40,21 @@
 #' # To replace errors with a default value, use possibly().
 #' list("a", 10, 100) %>%
 #'   map_dbl(possibly(log, NA_real_))
+#'
+#' # For interactive usage, auto_browse() is useful because it automatically
+#' # starts a browesr() in the right place.
```
lionel-: browser()

R/output.R
```diff
@@ -69,6 +84,45 @@ possibly <- function(.f, otherwise, quiet = TRUE) {
...
+      f(...),
+      error = function(e) {
+        # 1: h(simpleError(msg, call))
+        # 2: .handleSimpleError(function (e)  <...>
+        # 3: stop(...)
+        frame <- eval_frame(4)
+        browse_in_frame(frame)
+      },
+      warning = function(e) {
+        if (!identical(getOption("warn"), 2)) {
```
lionel-: should this be `if (getOption("warn") >= 2)`?

R/output.R
```diff
@@ -69,6 +84,45 @@ possibly <- function(.f, otherwise, quiet = TRUE) {
...
+          frame <- eval_frame(7)
+          browse_in_frame(frame)
+        }
+      }
+    )
+  }
+}
+
+browse_in_frame <- function(frame) {
+  if (.Platform$GUI == "ESS") {
```
lionel-: Nice I didn't know about that variable

R/output.R
```diff
@@ -69,6 +84,45 @@ possibly <- function(.f, otherwise, quiet = TRUE) {
...
+    # Workaround ESS issue
+    with_env(frame$env, on.exit({
+      browser()
+      NULL
+    }))
+    return_from(frame)
+  } else {
+    eval(quote(browser()), envir = frame$env)
```
lionel-: this should be `expr_eval()`, since `eval()` is not frame consistent (it creates a new frame that has `envir` as frame environment, so frame-dependent functions get tricked when they look up the eval stack.

I think `expr_eval()` should have a shorter name and become our default way of evaluating code. `base::eval()` has much more complicated logic due to the possibility of adding auxiliary data, and creates a quirky call stack. Maybe `exec()`? Not sure if this feels right, "eval" is such an established term.

R/output.R
hadley: What does the wrapping? I tried it, and it works in some sense, but it's never going to get you a `browser()` in exactly the right place

R/output.R
hadley: I had worried about people setting invalid values, but `options(warn = "a")` errors.

R/output.R
hadley: Can you confirm that's the correct value?

R/output.R
hadley: Ah, that makes some sort of sense, and switching to `expr_eval()` does work correctly. 

Maybe `eval_raw()`? Or `eval_clean()`?  I doubt these new functions are really going sink in with me until I've updated advanced R to use them.

R/output.R
lionel-: hmm I have `"X11"`. I'll take a look on the ESS side.

R/output.R
lionel-: it lets you check inputs though

PR #305, opened 2017-03-14
WIP: Implement partial() with interpolation

lionel-:
Current issues:

- Dots semantics change as a function of `.lazy` argument. We should probably make two different functions instead. `partial_now()` or `partial_eager()`?

- Partialisation of functions with dotted signatures doesn't work well right now. We'd need to parse the code tree, look for instances of `...`, and add the new argument there. But that could be fragile. Another way would be to modify the `...` object in the call frame to forward partialised arguments in a new pairlist.

- This approach is not compatible with S3 generics, and one example in the doc is about using `partial()` with `plot()`. So we'd need some special logic to check for generic functions and dispatch manually with partialisation.

in the end I'm not sure all this is worth the trouble?

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -20,6 +20,6 @@ Suggests:
     dplyr (>= 0.4.3),
     testthat,
 Remotes: 
-    hadley/rlang
+    hadley/rlang#58
 RoxygenNote: 6.0.1
 Roxygen: list(markdown = TRUE)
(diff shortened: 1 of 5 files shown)


hadley:
Replacing the body doesn't feel like quite the right strategy to me here. I think maybe we should discuss what the generated function should look like before we go too much further down this route.

PR #311, opened 2017-04-18
rlang update

lionel-:
- Use quosures in `list_modify()`. And evaluate any symbolic arguments and take dots with explicit splicing.

- Use `as_quosure()` in `as_mapper()` default method.

@hadley would it be better to add an option to `eval_tidy()` to treat bare formulas as quosures? I thought it's probably better to be consistent.

### DIFF
--- a/R/as_mapper.R
+++ b/R/as_mapper.R
@@ -133,5 +133,5 @@ extractor <- function(i, default) {
 
 #' @export
 as_mapper.default <- function(.f, ...) {
-  rlang::as_function(.f)
+  rlang::as_closure(.f)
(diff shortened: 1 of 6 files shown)


hadley:
Now that formulas are quosures are distinct, what if you used `eval_tidy_lhs()` and `eval_tidy_rhs()` for formulas?

lionel-:
I'm not sure what's the best way forward regarding formulas. It seems appealing to be consistent and restrict evaluation to quosures. In this case we'd remove `eval_tidy_lhs()` and rhs variants.

PR #313, opened 2017-04-21
Compat dplyr 0.6.0

lionel-:
The plan is to prepare the release in https://github.com/tidyverse/purrr/tree/release-0.2.2.1 and merge that branch in master afterwards. Could you please check the revdep email to check that everything is clear @hadley?


lionel-:
yes that'd be better. Most problems come from `map_df()` as the dplyr dep does not transit though purrr anymore, but one package (GGally) requires `dmap()`.

hadley:
Looks good. 

Let's aim for release on Monday — can you apologise for the short-notice, but comment that it should only require one of two changes (i.e. importing dplyr or importing purrrlyr), and that we need it for dplyr.

hadley:
Can you please notify everyone today?

lionel-:
Will do.

PR #317, opened 2017-04-29
Update help for modify_*() family

jennybc:
The examples were still using the deprecated `map_*()` functions.

I also changed some wording a little. The combined effect of the previous actually made me wonder if these functions somehow modified *the actual input object*. I knew that was highly unlikely and it's easily tested. But I tried to make the wording less confusing.


hadley:
This is much better — thanks!

PR #319, opened 2017-05-02
Add example of transpose() with .names; closes #318

### LINKED ISSUE #318: revisit: transpose() doesn't work as expected when inner keys are disjoint
Deliberately repeating name of #205. I'm not sure #205 really was a duplicate of #164.

`transpose()` seems to work just fine with names, so this is not about position. But it only seems to consult the first object to determine inner keys.

``` r
library(purrr)

ll <- list(
  list(x = 1, y = "one"),
  list(z = "deux", x = 2)
)

## only x and y appear, but what about z?
transpose(ll)
#> $x
#> $x[[1]]
#> [1] 1
#> 
#> $x[[2]]
#> [1] 2
#> 
#> 
#> $y
#> $y[[1]]
#> [1] "one"
#> 
#> $y[[2]]
#> NULL

## I expected this
list(
  x = map(ll, "x"),
  y = map(ll, "y"),
  z = map(ll, "z")
)
#> $x
#> $x[[1]]
#> [1] 1
#> 
#> $x[[2]]
#> [1] 2
#> 
#> 
#> $y
#> $y[[1]]
#> [1] "one"
#> 
#> $y[[2]]
#> NULL
#> 
#> 
#> $z
#> $z[[1]]
#> NULL
#> 
#> $z[[2]]
#> [1] "deux"
```

jennybc:


R/transpose.R
```diff
@@ -36,6 +36,16 @@
...
 #' x %>% transpose() %>% simplify_all()
+#'
+#' # Provide explicit component names to prevent loss of those that don't
+#' # appear in first component
+#' ll <- list(
+#'   list(x = 1, y = "one"),
+#'   list(z = "deux", x = 2)
+#' )
+#' ll %>% transpose()
+#' nms <- ll %>% map(names) %>% reduce(union) %>% unique()
```
hadley: I don't think you need unique here: the names of each element should already be unique, and `union()` ensures that they stay unique.

jennybc: removed 13b0b734d2801a119ab309a07ffce1db3bb3a53e

hadley:
Thanks!

PR #326, opened 2017-05-19
Fix typo in README.Rmd and generated files

markdly:
Fixed typo in README.md and index.html as well in case it is a while before these are regenerated from README.Rmd


hadley:
Thanks!

PR #333, opened 2017-06-11
Add persistently() adverb

richierocks:
Added `persistently()`, to repeatedly try a risky function up to a limited number of attempts.

Some discussion on twitter:

https://twitter.com/richierocks/status/874003431895257088


R/output.R
```diff
@@ -9,7 +9,8 @@
 #' @param quiet Hide errors (`TRUE`, the default), or display them
 #'   as they occur?
 #' @param otherwise Default value to use when an error occurs.
-#'
+#' @param max_attempts Positive integer. \code{persistent} functions will try
+#' to run this many times before giving up.
```
hadley: Can you please match indentation?

R/output.R
```diff
@@ -92,6 +111,31 @@ possibly <- function(.f, otherwise, quiet = TRUE) {
 
 #' @export
 #' @rdname safely
+persistently <- function(.f, otherwise = NULL, quiet = TRUE, max_attempts = 5) {
+  .f <- as_mapper(.f)
+  force(max_attempts)
+  function(...) {
+    for(i in seq_len(max_attempts)) {
```
hadley: Can you please put spaces after `for` and `if`?

richierocks:
@hadley Spaces after `if` and `for` added; roxygen lines indented; wait between tries feature added.

R/output.R
```diff
@@ -92,6 +116,41 @@ possibly <- function(.f, otherwise, quiet = TRUE) {
...
+      answer <- capture_error(.f(...), quiet = quiet)
+      if (is.null(answer$error)) {
+        return(answer$result)
+      }
+      if (wait_seconds > 0) {
+        actual_wait_seconds <- wait_seconds * 2 ^ (i - 1)
+        if (!quiet) {
+          message(sprintf("Retrying in %g seconds.", actual_wait_seconds))
+        }
+        Sys.sleep(actual_wait_seconds)
```
hadley: You are missing some randomness here

hadley:
Can you please merge/rebase and add a bullet to NEWS?

richierocks:
OK, done.

ijlyttle:
I have been interested this PR, and having the "persistently" capability. Since there has been a CRAN release since the last activity, I thought I would extract the "guts" of this PR and put it into its own package - just for now: https://github.com/ijlyttle/warrenr

The thought is to take `warrenr::persistently()` out for a spin with the new purrr; my ultimate goal is to merge that into this PR so that everything would be in purrr.

I have found a couple of things:

1. There were a few fiddly CRAN-check snags here, fixed now at __warrenr__.

2. I would like to propose to remove the `otherwise` argument from `persistently()`. Here, `persistently()` acts like `possibly()`. I propose that `persistently()` return an error - in my case, I want a hard failure with an error message.  This is currently how `warrenr::persistently()` works. This way, its behavior would be orthogonal to `possibly()` and `safely()`. If you want to invoke either of these, I'm sure you could compose:

```r
possibly(persistently(do_this), otherwise = NULL)(args)
```

If all of this is agreeable, I can make a PR to this PR in the next little while - I would like a little time (maybe a week or so) to give `warrenr::persistently()` a good shakedown.

Thoughts, @richierocks, @hadley ?

richierocks:
@ijlyttle Yes, I like the idea of separating the behavior of `persistently()` and `possibly()`. Maybe add an example of `possibly(persistently(fn))` to `examples(persistently)` to show how they can be used toegether.

ijlyttle:
@richierocks: good idea - I will add such an example

lionel-:
I find @ijlyttle's arguments convincing and agree the error should be rethrown. I think the function would be better named `insistingly()` though, because persistence has other connotations in programming.

lionel-:
Perhaps `stubbornly()` then? cc @hadley @jennybc

Would you like to include your fixes from warrenr to this PR Ian?

jennybc:
I like `insistently()`.

`repeatedly()` is also very descriptive. Sorry to add another option 🙂

lionel-:
Thanks @richierocks and @ijlyttle!

PR #354, opened 2017-07-19
Update API for list_ functions

lionel-:
- Take dots with splicing instead of a list.

- `list_update()` transforms formulas to quosures with a warning, for compatibility with `update_list()`.

With this change `list_modify()` and `list_update()` both take dots. I'm not sure about the usefulness of the latter. It seems like we need a `mutate()` method for lists instead.


R/list-modify.R
```diff
@@ -5,48 +5,61 @@
...
+list_merge <- function(.x, ...) {
+  dots <- dots_list(...)
+  list_recurse(.x, dots, c)
+}
+#' @export
+#' @rdname  list_modify
+list_update <- function(.x, ...) {
+  dots <- dots_list(...)
+
+  formulas <- map_lgl(dots, is_bare_formula, lhs = FALSE, scoped = TRUE)
```
hadley: I'm not sure if we should even take quosures. The whole evaluation API seems quite odd now. Maybe we should leave as is in `update_list()` and just remove in `list_update()` until we can spend some time on it in the next version?

lionel-: Yes I think it's best.

lionel-: Are the other changes reasonable? e.g. taking dots with splicing. It seems like this should now be our default.

hadley: Yes, definitely

PR #355, opened 2017-07-21
Update pluck()

lionel-:
- Take arguments by dots with automatic list splicing
- Support accessor functions


NEWS.md
```diff
@@ -18,6 +18,25 @@ purrr and dplyr:
...
+non-linearly:
+
+```
+accessor(x[[1]])$foo
+```
+
+to the equivalent pluck:
+
+```
+pluck(x, 1, accessor, "foo")
```
hadley: Maybe include `x %>% pluck(1, accessor, "foo")` ?

hadley: It also occurs to me that we've made `pluck()` super duper general:

```R
x %>% f() %>% g()
# same as 
x %>% pluck(f, g)
```

i.e. it's feel like a full on compose operator. I don't think we should change the name or emphasise this, but it's interesting.

R/as_mapper.R
```diff
@@ -62,35 +61,77 @@ as_function <- function(...) {
 
 #' Pluck out a single an element from a vector or environment
 #'
-#' This is a generalised form of `[[` which allows you to index by
-#' name, position, or attribute.
+#' @description
+#'
+#' This is a generalised form of `[[` which allows you to index deeply
+#' and flexibly into data structures. It supports R standard accessors
+#' like integer positions and string names and accepts any accessor
```
hadley: .., and also accepts arbitrary accessor functions, functions that take an object and return some internal piece.

R/as_mapper.R
```diff
@@ -62,35 +61,77 @@ as_function <- function(...) {
...
+#' cruft. Compare: `accessor(x[[1]])$foo` to `pluck(x, 1, accessor,
+#' "foo")`.
 #'
-#' @param x A vector or environment
-#' @param index A list indexing into the object
-#' @param default Value to use if target is empty or absent.
+#' @param .x A vector or environment
+#' @param ... A list of accessors for indexing into the object. Can be
+#'   an integer position, a string name, or an accessor function.
+#'   These dots [splice lists automatically][rlang::dots_splice]. This
```
hadley: Mention that it never does partial matching?

R/as_mapper.R
```diff
@@ -62,35 +61,77 @@ as_function <- function(...) {
...
+#'
+#' # And now an accessor for these complex data structures:
+#' my_element <- function(x) x[[2]]$elt
+#'
+#' # The accessor can then be passed to pluck:
+#' pluck(x, 1, my_element)
+#' pluck(x, 2, my_element)
+#'
+#' # Even for this simple data structure, this is more readable than
+#' # the alternative form:
```
hadley: because you have to read both from right-to-left and left-to-right in different parts of the code

R/as_mapper.R
hadley: Probably should also mention that a string will index into the slots of a S4 object and the names of list.

R/as_mapper.R
```diff
@@ -62,35 +61,77 @@ as_function <- function(...) {
...
-
 #' @export
-#' @rdname as_mapper
-#' @param x A string
-get_attr <- function(x) {
-  stopifnot(is.character(x))
-  structure(x, class = "attr")
+#' @rdname pluck
+#' @param attr An attribute name as string.
+attr_getter <- function(attr) {
```
hadley: Should we `force(attr)` here?

NEWS.md
lionel-: yup it's like an indexing-oriented composition operator. It is also a bit more explicit regarding the intent of the composition (e.g. extraction).

R/as_mapper.R
lionel-: What do you mean about partial matching?

hadley: It behaves like [[ not like $

PR #356, opened 2017-07-21
Prepare release 0.2.3

lionel-:


NEWS.md
```diff
@@ -1,4 +1,16 @@
-# purrr 0.2.2.9000
+# purrr 0.2.3
+
+## Breaking changes
+
+We noticed the following issues during reverse dependencies checks:
+
+* The reduce functions now issue an error when the input is empty and
```
hadley: This needs to be written from the perspective of the symptom. Something likes:

If you see any error like `Error: `.x` is empty, and no `.init` supplied` that comes from `reduce()`, this is because `reduce()` now returns `.init` when `.x` is empty. Fix the problem by supplying an appropriate argument to `.init`, or by providing special behaviour when `length(x) = 0`

PR #361, opened 2017-08-04
fix pluck.Rd

yutannihilation:
The help page of `pluck()` is corrupted a bit.

I heard this is a known issue of roxygen2 before (https://github.com/tidyverse/dplyr/pull/2801). If this issue is not fixed yet, I believe using `\code{}` instead of backtick is the only choice...


lionel-:
Thanks!

PR #362, opened 2017-08-04
use map_dfr() instead of map_df() in documentations

yutannihilation:
As NEWS says:

> The data frame suffix `_df` has been (soft) deprecated in favour of `_dfr` to more clearly indicate that it's a row-bind.

### DIFF
--- a/R/map.R
+++ b/R/map.R
@@ -93,11 +93,11 @@
 #' mtcars %>% map_dbl(sum)
 #'
 #' # If each element of the output is a data frame, use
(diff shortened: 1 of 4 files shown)


yutannihilation:
By the way, `cross_df()` keeps `_df`. Is this intended?

lionel-:
Thanks!

We've already renamed `cross_df()` from `cross_d()` so it may be a bit too much to rename it again.

yutannihilation:
Agreed 👍

Issue #349, opened 2017-07-14
partial() fails for functions that introspect context


egnha:
Example:
```R
ls_all <- purrr::partial(ls, all.names = TRUE)

ls(all.names = TRUE)
#> [1] ".Random.seed" "ls_all"

# ought to match ls(all.names = TRUE)
ls_all()
#> [1] "..."

ls_all(.GlobalEnv)
#> [1] ".Random.seed" "ls_all" 
```
The problem is that `ls_all()` doesn't faithfully mimic calling `ls()` (with `all.names = TRUE`), because it doesn't call `ls()` from the calling environment.

Here’s a first pass at a work-around (ignoring the `.lazy`, `.env` arguments):
```R
# rename '.first' to '...first' to allow '.first' in '...'?
partial <- function(...f, ..., .first = TRUE) {
  defvals <- rlang::dots_list(...)
  if (rlang::is_empty(defvals))
    return(...f)
  ...subst_defvals <- function(call) {
    if (.first)
      fcall <- c(...f, defvals, rlang::node_cdr(call))
    else
      fcall <- c(...f, rlang::node_cdr(call), defvals)
    as.call(fcall)
  }
  
  rlang::set_attrs(
    function(...) {
      call <- ...subst_defvals(sys.call())
      rlang::eval_bare(call, parent.frame())
    },
    class = "partial_function"
  )
}
```
A minimal print method could be something like this:
```R
print.partial_function <- function(x, ...) {
  cat("<partial_function>\n")
  cat("\n* Default values:\n")
  cat(itemize_vals(environment(x)$defvals), "\n", sep = "")
  cat("\n* Original function:\n")
  print(environment(x)$...f)
  invisible(x)
}
itemize_vals <- function(vals) {
  vals <- lapply(vals, rlang::expr_label)
  paste(sprintf("%s = %s", names(vals), vals), collapse = "\n")
}
```

egnha:
Here‘s a simpler implementation of `partial()`, that has the advantage of preserving the formals of `...f`, so far as that is possible:
```R
library(rlang)

partial <- function(...f, ..., .lazy = TRUE) {
  if (.lazy) {
    env <- child_env(environment(...f))
    env_bind_exprs(.env = env, ...)    
  } else {
    env <- child_env(.parent = environment(...f), ...)
  }

  f <- as_closure(...f)
  fmls <- formals(f)
  fmls <- fmls[!names(fmls) %in% names(env)]
  
[...]

PR #365, opened 2017-08-06
tidyeval partial() and fix for #349

egnha:
A rewrite of `partial()` to enable:

- tidy evaluation of partialised arguments

- a fix for #349, by ensuring that the underlying function is itself called in the calling environment

Aside from these changes, I have tried to maintain the behavior of `partial()`, as the tests seem to affirm. The examples in the doc work as before.

The roles of `.env` and `.lazy` have been supplanted by tidyeval and quasiquotation, so an error is signaled, with a deprecation message, if one tries to invoke them.


R/partial.R
```diff
@@ -43,45 +50,67 @@
...
+    abort("`.env` is deprecated")
+  if (!missing(.lazy))
+    abort("`.lazy` is deprecated; unquote values to evalute them immediately")
 
-  if (.lazy) {
-    fcall <- substitute(...f(...))
-  } else {
-    fcall <- make_call(substitute(...f), .args = list(...))
-  }
+  f <- as_closure(...f)
```
hadley: Would `as_mapper()` be more consistent with the rest of purrr?

R/partial.R
```diff
@@ -12,18 +12,26 @@
...
-#'   [parent.frame()] and you should rarely need to modify this.
-#' @param .lazy If `TRUE` arguments evaluated lazily, if `FALSE`,
-#'   evaluated when `partial` is called.
+#'   [Unquoting and splicing][rlang::UQ()] semantics are supported, see
+#'   _Examples_.
 #' @param .first If `TRUE`, the partialized arguments are placed
 #'   to the front of the function signature. If `FALSE`, they are
 #'   moved to the back. Only useful to control position matching of
 #'   arguments when the partialized arguments are not named.
+#' @param .env,.lazy Obsolete, see _Details_.
```
hadley: I think it would be more useful to extract the info out of details and re-include here.

R/partial.R
```diff
@@ -43,45 +50,67 @@
...
-    tmp[seq(3, n + 1)] <- fcall[seq(2, n)]
-    names(tmp)[seq(3, n + 1)] <- names2(fcall)[seq(2, n)]
-    fcall <- tmp
-  } else {
-    fcall[[n + 1]] <- quote(...)
+  if (is_empty(args))
+    return(f)
+
+  fill_args <- function(call) {
+    vals <- lapply(args, eval_tidy)
```
hadley: Is it right that these should get re-evaluated on every call? Or should they be lazily evaluated once?

R/partial.R
egnha: Makes sense, they go together.

R/partial.R
egnha: For consistency, `as_mapper()` might indeed be a better choice. I hadn't considered that. I slightly prefer `as_closure()` because I think it more clearly expresses why `...f` is being transformed: we need to apply arguments by name, and that generally won't work for primitive functions.

R/partial.R
egnha: You are absolutely right. Only once. Thanks!

egnha: I think `vals` should be (tidily) eagerly evaluated, even though this diverges from R‘s usual call semantics, as I think this fits better with the notion that `partial()` “fixes” default values. Also, the original function won't necessarily understand tidyeval. So if the partialized arguments are to comprehend tidyeval, they have to be tidyeval'd before being passed to `f`. Or am I missing something?

egnha: Need to backtrack ... now I remember why I was setting `vals` to re-evaluate on every call: to indeed mimic R‘s call semantics, whereby each call (lazily) evaluates the arguments anew (except here the arguments are eagerly evaluated, for the reasons in the previous comment).

On the other hand, you can unquote to fix values. This is akin to, but of course not literally the same as, evaluating the arguments once. I believe this is consistent with the tidyverse practice of unquoting.

Example:
```R
# Evaluation of partialized arguments occurs with every call, so runs may differ in length
f <- partial(runif, n = rpois(1, 5))
f()  #> [1] 0.8655788 0.4433602 0.5719572 0.8810286 0.7277090
f()  #> [1] 0.1200088 0.9801471 0.8726577 0.2121102

# But the run length can be fixed by unquoting:
f <- partial(runif, n = !!rpois(1, 5))
f()  #> [1] 0.3450202 0.4374072 0.2143421 0.3036093
f()  #> [1] 0.5108066 0.4283015 0.7340954 0.3224990
```

hadley: Oh of course — I wish we could describe the semantics of `partial()` better so that it's more clear to us what the desired behaviour is. My main concern with making any changes to `partial()` is that we'll accidentally change the API just because we don't remember what the desired semantics are.

lionel-: I think `.lazy` should not be made defunct but soft-deprecated. The default `.lazy = TRUE` would use the new behaviour which is consistent, and a new code path for the `FALSE` case should `eval_tidy()` once and for all.

egnha: Hmm, I see your point. Maybe it is indeed better to enable quasiquotation but otherwise leave the behavior of `partial()`, and its argument signature, unchanged.

If you are willing to import the [nofrills package](https://github.com/egnha/nofrills) into purrr, then a `partial()` that satisfies these requirements can be quite succinct:
```R
partial <- function(...f, ..., .env = parent.frame(), .lazy = TRUE, .first = TRUE) {
  f <- substitute(...f)
  args <- if (.lazy) exprs(...) else lapply(quos(...), eval_tidy)
  if (.first)
    nofrills::fn(... ~ (!! f)(!!! args, ...), ..env = .env)
  else
    nofrills::fn(... ~ (!! f)(..., !!! args), ..env = .env)
}
```

lionel-: I think using quasiquotation to opt into eager behaviour is an interesting idea, just that the deprecation needs to happen incrementally. By the way if you make it an error to use an argument, this means it's defunct not deprecated as the latter only implies a warning. In general we try to soft-deprecate first without any warning (in the future testthat will send a warning for soft-deprecated functions but only when testing interactively).

egnha:
@lionel-, @hadley Thanks for reviewing this PR. Considering your critiques, I think this PR diverges too much from the current API—or at least more than necessary to meet its aims. Also, I realize now that the premise of #349 is flawed and not worth salvaging (in my opinion). So I am closing this PR.

Nonetheless, quasiquotation for partialized arguments is a worthwhile aim, so I have opened an issue for that (#457). I've suggested an implementation. The goal this time is more clearly delineated, I hope.

lionel-:
> It's incompatible with byte-code-compiled functions (the offending line is eval_bare(call, parent.frame())).

Are you sure about this? Could you make a reprex of the offending behaviour?

> Either the environment of the original function or the original function itself must be in scope when the partialised function is called; but then it is impossible for the final function call to be evaluated in the environment in which the partialized function is called.

Just inline the original function in the partialised call then evaluate the call in the caller env.

egnha:
@lionel- Thanks for prodding me to examine the issue more carefully! It looks like you correct.

I was thrown off by `str()`:

```R
foo <- function() {
  f <- compiler::cmpfun(function() NULL)
  function(...) {
    # Inline f
    call <- sys.call()
    call[[1]] <- f
    call
  }
}

str(foo()())
#> language (function ()  NULL)()
```

This makes it look like the call in `foo()()` is no byte complied. But in fact it is :-) 👍 

```R
```

(A more damning issue is that the of environment of the original function won't generally be visible from the calling environment.)

Alternatively, one could evaluate the (byte-complied) function be referencing it by name. But in order to do that, the environment of the name must be a child of the calling environment. But that would exactly defeat the purpose of #349 (to call the original function with arguments partialised in the calling frame).

The upshot is that I think `partial()` needs to (necessarily?) return a function of the form
```R
function(...) {
  # partialised function call
}
```
as `purrr::partial()` currently does.

Maybe I am missing something trivial, but I don't see way around this. Do you?

PR #369, opened 2017-08-15
Fix mapping to data frames with functions returning vectors

lionel-:
Since we had to preserve original semantics for dataframeable lists in `bind_rows()` and `bind_cols()`, the lists of vectors need to be explicitly spliced.


hadley:
@lionel- Can you please re-open this?

PR #375, opened 2017-08-22
Get purrr working on R3.1

hadley:
And ensure it stays that way by testing more R versions on travis

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -2,11 +2,11 @@ comment: false
 
 coverage:
   status:
-    patch:
-      default:
-        target: 0
-        threshold: 100
     project:
       default:
-        target: 70
-        threshold: 100
+        target: auto
+        threshold: 10%
+    patch:
+      default:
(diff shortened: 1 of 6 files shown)


codecov.yml
```diff
@@ -2,11 +2,11 @@ comment: false
 
 coverage:
   status:
-    patch:
```
lionel-: Did you switch back to verbose codecov on purpose?

hadley: Yeah, this is now the standard tidyverse spec (from usethis)

lionel-: Got it. 10% seems reasonably quiet.

I wonder if `patch` should still have a target of zero, otherwise PRs will fail if they don't meet the overall project coverage. It's not uncommon to add or change uncovered code in a PR.

hadley: Let's see how it does - this does allow a drop in coverage of 10%

PR #386, opened 2017-09-27
replaced quote by tick mark

EdwinTh:
_list-elements_ was between a tick mark and a quote. From the context it seems like it should be between tick marks.

### DIFF
--- a/R/lmap.R
+++ b/R/lmap.R
@@ -6,7 +6,7 @@
 #' \emph{and} return a list (or data frame). Thus, instead of mapping
 #' the elements of a list (as in \code{.x[[i]]}), they apply a
 #' function `.f` to each subset of size 1 of that list (as in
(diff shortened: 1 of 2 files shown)


lionel-:
Could you reoxygenise as well please?

EdwinTh:
Sure. However it seems that the documentation of a whole bunch of functions needs to be reoxygenised as well (everywhere `code{}` is replaced by tick marks). It doesn't seem appropriate to have them in the PR.

lionel-:
If you use the same roxygen version as the project uses there shouldn't be any problem.

I think purrr is using CRAN roxygen so no worries there, otherwise you can install a specific version with `install_github("klutometis/roxygen", ref = "v6.0.1")`.

EdwinTh:
Thanks, I indeed had to update my roxygen

lionel-:
Thanks!

PR #408, opened 2017-11-10
consistent output in examples

wibeasley:
the output for the `z` element is a little different (compared to the lines above it) without `str()`.  Using `str()` throughout makes it more clear what hasn't changed.


hadley:
Thanks!

PR #424, opened 2017-12-14
Added mapper support for compose

ColinFay:
This allows to use mappers inside `compose` : 

```r
clean_lm <- compose(~ arrange(.x, desc(p.value)), 
                    ~ filter(.x, p.value < 0.05),
                    tidy, 
                    summary, 
                    lm)

clean_lm(Sepal.Length ~ Sepal.Width, data = iris)
```

instead of 

```r
clean_lm <- compose(as_mapper(~ arrange(.x, desc(p.value))), 
                    as_mapper(~ filter(.x, p.value < 0.05)),
                    tidy, 
                    summary, 
                    lm)
clean_lm(Sepal.Length ~ Sepal.Width, data = iris)

```
The ` args <- modify_if(list(...), is.character, get)` is there so you can still mimic the `match.fun` behavior, which can take a character object.

### DIFF
--- a/R/compose.R
+++ b/R/compose.R
@@ -1,6 +1,6 @@
 #' Compose multiple functions
 #'
-#' @param ... n functions to apply in order from right to left.
+#' @param ... n functions to apply in order from right to left. Args can be symbols, character vector, functions or one side formula.
 #' @return A function
 #'
 #' @export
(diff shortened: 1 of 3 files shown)


hadley:
I'd rather think this through a bit more before introducing another set of semantics. Would you mind filing an issue so we can discuss?

Issue #217, opened 2016-07-11
purrr:contains and dplyr::contains is likely to cause confusion


sebastiansauer:
Both `purrr` and `dplyr` have a contains() function, but with different syntax.

For example, if I load dplyr first, and purr afterwards, it will bite me:

```
   library(dplyr)
   library(purrr)
   data(tips, package = "reshape2")
   tips %>% select(contains("total"))
   ## Error in .p(.x[[i]], ...) : argument ".y" is missing, with no default
```

May it would be helpful to disambiguate the two functions?

hadley:
@lionel- any thoughts on a better name? We might need to consider more prefixes for purrr functions since it's going to be loaded increasingly commonly

lionel-:
we could prefix with `vec_` secondary functions that operate on vectors, generic and atomic? But then what would be the prefix for atomic vectors functions?

hadley:
`atm_`?  But if you think of `contains()` as a function that operates on vectors, maybe it should go in vctrs? I think we'll need to figure out what goes where.

lionel-:
this makes me think that the tidyverse is starting to look like Emacs lisp functions ;) Elisp doesn't have namespaces so all function names have to be prefixed.

>  maybe it should go in vctrs?

I'm not sure, this is a wrapper around `some()`. We'll also have to figure out what should be the dependency graph of those packages.

> atm_?

ah nice, I was stuck with `ato_` for some reason.

mexindian:
Or just change one of the verbs to `has()`?

jtr13:
Yes, confusion caused.
> select(x100, contains("P"))
Error in .p(.x[[i]], ...) : argument ".y" is missing, with no default

hadley:
And similarly for `order_by()`

kleinschmidt:
Why did `sort_by` need to be removed as well, and what's the recommended replacement (other than "use tibbles")?

PR #431, opened 2017-12-23
Split by

t-kalinowski:
Hello,

This is a proposal for a companion to `map_dfr()` that's a little easier to work with than than `base::split()`. 

If there is interest, I would be happy to add tests and flesh out the docs a bit.


yutannihilation:
For your information, another `split_by()` had been implemented once and was removed on version 0.2.3. (The semantic seems different than yours, though)

* https://github.com/tidyverse/purrr/blob/master/NEWS.md#purrr-023
* https://www.rdocumentation.org/packages/purrr/versions/0.2.2/topics/split_by

cderv:
For reference, `split_by` has been removed as result of #217 in https://github.com/tidyverse/purrr/commit/7dc96a8dd1e257a2c16f91901912c9c0e6c2fc1c.

For information too, some functions at the intersection of `purrr` and `dplyr` has been moved to [purrrlyr](https://github.com/hadley/purrrlyr). Your idea may belong there.

* See [NEWS.md](https://github.com/tidyverse/purrr/blob/master/NEWS.md#purrr-0221) of 0.2.2.1 version
* See documentation of `purrrlyr` on [rdocumentation](https://www.rdocumentation.org/packages/purrrlyr/versions/0.0.2/)

t-kalinowski:
Thank you @yutannihilation and @cderv, I wasn't aware (or had forgotten) that a function named `split_by()` had existed at one point. What I'm proposing here seems quite different from that function though. Perhaps a different name is needed.  

Also, I might be mistaken, but the fact that `purrrlyr` is not part of the tidyverse suggests that `purrrlyr` not being actively developed anymore, in favor of `purrr` and `dplyr`.

hadley:
Yeah, this doesn't feel quite right for purrr to me.

t-kalinowski:
Ok, Thanks for considering.

As I've been using `split_by` more and more, I find it's much easier to work with than the alternatives. `do()` is confusing, clunky, and limiting. (it wasn't until I read the source that I realized that `.` is not evaluated by magrittr as a pipe, but is actually an active binding that evaluates to each subset)

 and `nest() + `mutate(... map(data, ...))` is powerful for when you need it but it's a lot of syntax and infrastructure to write and sometimes you just want a lighter-weight solution. (also, it's limiting in it's own way because you don't have straightforward access to the grouping variable within the function you're mapping, forcing you to pass it in as a separate named argument... again leading to more verbose syntax)

The best alternative then is `base::split`, which works great when splitting on one variable
``` df %>% split(.$var1) %>% map(fun)```
 but that suddenly requires a lot of syntax if you're splitting on more than one variable
`df %>% split(list(.$var1, .$var2), drop = TRUE)) %>% map(fun)`

Hence this pull request.

I agree it's not very purrr-like since it operates on bare symbols, but I do think there is a missing piece here in terms of functionality.

Issue #446, opened 2018-01-22
Should accumulate and accumulate_right keep names of the input?


AshesITR:
I was quite surprised to see that accumulate_right and accumulate do not keep the names of the input list.
For my case it makes perfect sense to do so, because I am aggregating along a hierarchy.

Maybe there should be an argument to accumulate(_right) for keeping the names?
For compatibility it might be like this:

`l %<>% accumulate(myfun)` -> no names
`l %<>% accumulate(myfun, .keep_names = TRUE)` -> leaves `names(l)` untouched

Currently I work around this with
`l %<>% {set_names(accumulate(., myfun), names(.))` wich is somewhat clunky.

I could submit a PR with the necessary changes if desired.

hadley:
A PR would be great!

PR #458, opened 2018-02-06
Add .keep_names to `accumulate()` and `accumulate_right()`

AshesITR:
Implement #446 
Maybe `.keep_names = TRUE` is a better default?
`.keep_names = FALSE` as default matches the original behaviour, but I can hardly imagine code relying on the fact that `accumulate(...)` is unnamed.


lionel-:
Maybe we should just always keep names instead of adding a new parameter.

AshesITR:
I have no hard feelings towards either way, as long as we _can_ keep the names.
As referenced in the issue I was surprised by the default behaviour, wich hints towards your suggestion.

hadley:
We should be consistent with the rest of purrr, and always keep names.

Can you please add a bullet to NEWS? It should briefly describe the change (starting with name of the function), and crediting yourself with `(@yourname, #issuenumber)`.

Please includes `Fixes #xyz` in a commit message (or the text of the PR): that way the corresponding issue will be automatically closed when the PR is merged.

R/reduce.R
```diff
@@ -170,25 +171,43 @@ seq_len2 <- function(start, end) {
...
+accumulate <- function(.x, .f, ..., .init, .keep_names = FALSE) {
   .f <- as_mapper(.f, ...)
 
   f <- function(x, y) {
     .f(x, y, ...)
   }
+  
+  # Stop early so costly Reduce is called only when everything is fine
+  if (!is.logical(.keep_names) || length(.keep_names) != 1 || is.na(.keep_names)) {
+    rlang::abort("`.keep_names` must be TRUE or FALSE")
```
lionel-: You could use `is_logical(.keep_names, 1L) && !is.na(.keep_names)`. Also we don't prefix rlang functions with a namespace.

hadley: We're getting rid of that check for this PR, so this is advice for future reference.

AshesITR: Thanks for the advice.

AshesITR:
@hadley is everything in order now?
I thought about adding some tests as well, but haven't found the time to do so:
`accumulate(set_names(1:26, letters), sum)` should be the same as `set_names(cumsum(1:26), letters)`

R/reduce.R
```diff
@@ -142,6 +142,7 @@ seq_len2 <- function(start, end) {
 #' both functions keep the intermediate results.
 #'
 #' @inheritParams reduce
+#' @param .keep_names If `TRUE`, names of `.x` will be copied to the result. Otherwise the result will be unnamed.
```
hadley: You can now remove this

AshesITR:
@hadley Ooops.
Changed it so that the docs still tell you it keeps the names.

AshesITR:
@hadley Merged most recent PR. It should be ready to pull now.
(How) can I remove the label:wip?

R/reduce.R
```diff
@@ -139,7 +139,7 @@ seq_len2 <- function(start, end) {
 #'
 #' `accumulate` applies a function recursively over a list from the left, while
 #' `accumulate_right` applies the function from the right. Unlike `reduce`
-#' both functions keep the intermediate results.
+#' both functions keep the intermediate results and the names of `.x`.
```
hadley: Rather than adding to this sentence (where it doesn't quite belong). I think it would be better to add a `@return` section which describes the result value (i.e. A vector the same length of `.x` with the same names as `.x`)

AshesITR:
@hadley Thanks for the suggestion. I haven't found a good way to say that `accumulate` might also return a list if the result doesn't fit into an atomic vector. Do you deem that comment necessary?

hadley:
Thanks!

PR #461, opened 2018-02-07
Don’t use partial matching for attr_getter()

### LINKED ISSUE #460: attr_getter() uses partial matching, which is error-prone/dangerous
Here’s a bug/misfeature is was bitten by today. Basically, I was trying to extract the `label` associated with each variable in a tibble, using (bascially) `map(df, attr_getter("label"))`.

However, for some of the variables, the value of the `labels` (note the `s`) attribute, a ‘> 1’-length vector, were extracted instead. It turned out that this happens when the `label` attribute is missing. The reason is partial matching on the attribute name, which unfortunately is the default in `attr()`.

It’s to late to change the base `attr()` function, but partial matching is **dangerous** :skull_and_crossbones:, and I see no reason that this should be the default in `attr_getter()` too!

To fix this, one only needs to add `exact = TRUE` in the call to the `attr` function in `get_attr()`.

Reprex:

```r
library(purrr)
x = c(0, 1, 1, 0)
attr(x, "labels") = c(`0` = "Female", `1` = "Male")
get_label = attr_getter("label")
get_label(x) # Expected output: NULL
#>        0        1 
#> "Female"   "Male"
```

huftis:
Changed the `attr_getter()` fuction to no longer uses partial matching.
For example, if an `x` object has a `labels` attribute but no `label`
attribute, `attr_getter("label")(x)` will no longer extract the `labels`
attribute (fixes #460).

Also added some documentation on the `attr_getter()` function.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,10 @@
 # purrr 0.2.4.9000
 
+* `attr_getter()` no longer uses partial matching. For example,
+  if an `x` object has a `labels` attribute but no `label` attribute,
+  `attr_getter("label")(x)` will no longer extract the `labels`
+  attribute (#460, @huftis).
+
 * `flatten_dfr()` and `flatten_dfc()` now aborts if dplyr is not installed. (#454)
 
 * `imap_dfr()` now works with `.id` argument is provided (#429)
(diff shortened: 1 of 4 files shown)


hadley:
Would you mind adding a unit test too please?

huftis:
OK, @hadley, I’ve now added a simple unit test that tests

1. basic functionality of `attr_getter()`
2. that `attr_getter()` uses exact (non-partial) matching

hadley:
Thanks!

PR #410, opened 2017-11-13
Add modify methods for integer, double, and character

t-kalinowski:
`modify` methods for integer, double and character, to complement `map_int` `map_dbl` and `map_chr`.

PR #463, opened 2018-02-10
More modify methods

t-kalinowski:
clean PR of #410 (one of these years I'll get the hang of git :)


R/modify.R
```diff
@@ -87,6 +89,30 @@ modify.default <- function(.x, .f, ...) {
   .x[] <- map(.x, .f, ...)
   .x
 }
+#' @rdname modify
+#' @export
+modify.integer  <- function (.x, .f, ...) {
```
hadley: I don’t think it’s necessary to document these individual methods - I think it makes the docs harder to understand.

t-kalinowski:
OK, please take a look.

hadley:
Thanks for all your hard work on this!

PR #465, opened 2018-02-13
remove pipes where unnecessary

jankislinger:
There are some uses of pipe inside function definitions where removing those doesn't affect readability. But they unnecessarily increase computation time. Not by much but once used in loops the difference might be significant.

```
require(purrr)
microbenchmark::microbenchmark(
  identity(1), 1 %>% identity(),
  identity(identity(1)), 1 %>% identity() %>% identity(),
  unit = 'us'
)
#> Unit: microseconds
#>                            expr    min     lq      mean median       uq     max neval
#>                     identity(1)  0.000  0.001   0.18351  0.001   0.3210   4.491   100
#>                1 %>% identity() 60.942 63.187  70.55381 65.111  67.6770 135.994   100
#>           identity(identity(1))  0.001  0.322   0.56531  0.642   0.6425   1.604   100
#> 1 %>% identity() %>% identity() 92.053 94.940 109.15809 97.666 103.9205 302.137   100
```

### DIFF
--- a/R/arrays.R
+++ b/R/arrays.R
@@ -45,13 +45,13 @@
 #' # margin argument:
 #' array_tree(x, c(3, 1)) %>% str()
 array_branch <- function(array, margin = NULL) {
-    dim(array) <- dim(array) %||% length(array)
-    margin <- margin %||% seq_along(dim(array))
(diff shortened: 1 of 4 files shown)


hadley:
Thanks!

PR #468, opened 2018-03-02
fixes #378 - replaces get_attr for attr_getter

### LINKED ISSUE #378: get_attr() was renamed attr_getter() but is still referenced in a number of places
During the 2.2.9 process get_attr() became attr_getter() but it still seems like there are a number of places in documentation that refer to it by the original name.
https://github.com/tidyverse/purrr/search?utf8=%E2%9C%93&q=get_attr&type=

edublancas:


hadley:
It looks like you modified the generated `.html` files, not the underlying roxygen comments.

PR #469, opened 2018-03-04
Closes #428 update pwalk to conserve expressions, add unit test, update NEWS

### LINKED ISSUE #428: walk and walk2 treat language objects differently
There is a discrepancy between how walk and walk2 process a column that has 
language objects in it.

Walk2 will evaluate the language objects before passing them into .f. Walk will leave the objects unchanged and pass them into .f as language objects.

I think that walk2 is handling the column incorrectly, but in
any case it seems that  both walk and walk2 should handle the column in the same 
way.

I haven't checked map2.

Here is an example.

``` r
# walk2 quosure issue
suppressPackageStartupMessages(library(rlang))

# make a quosure
qwalk <- rlang::quo(1 + 3)
# the rhs is an expression
# remember that 1 + 3 is 4
expr <- rlang::f_rhs(qwalk)
# put the expression into a table
tbl1 <- tibble::tribble (
    ~ id, ~ expr,
    1, expr
)
#
# notice that the type of the expr column is langauge
tbl1
#> # A tibble: 1 x 2
#>      id       expr
#>   <dbl>     <list>
#> 1     1 <language>
#
# use walk2 to walk through table values
purrr::walk2(tbl1$expr, tbl1$id, function(expr, id) {
    # just print out value of expr
    # notice that expr is 4 and its type is double
    # this means the expression has been evaluated
    print(expr)
    print(typeof(expr))
})
#> [1] 4
#> [1] "double"
#
# use walk to to walk though a table value
purrr::walk(tbl1$expr, function(expr) {
    # notice that the value of expr 3 + 1 is the inital expression
    # and the type is langauge, i.e. the type of an expression
    print(expr)
    print(typeof(expr))
    })
#> 1 + 3
#> [1] "language"
```

keithgw:


hadley:
Thanks for the fix, but I'd rather fix this further upstream (i.e. with a replacement for `do.call()`)

Issue #474, opened 2018-03-11
purrr:::transpose returns NULL on second list element


AndreMikulec:
When I try to transpose a list that has sub elements of different sizes,
NULL is returned for the 2nd list ( that is composed of different sized elements)
```r
LIST  <- list(A=list(a=1), B=list(b1=21,b2=22))
str(LIST)

  List of 2
   $ A:List of 1
    ..$ a: num 1   # one element
   $ B:List of 2   
    ..$ b1: num 21 # two elements
    ..$ b2: num 22

str(purrr::transpose(LIST))

List of 1
 $ a:List of 2    
  ..$ A: num 1    # success
  ..$ B: NULL     # lost(fail)
```
I have been working around like this.
This could be useful to me in my everyday work.
```r
NEW_LIST <- map(LIST,purrr::transpose)
str(NEW_LIST)

List of 2
 $ A:List of 1
  ..$ :List of 1
  .. ..$ a: num 1
 $ B:List of 1
  ..$ :List of 2
  .. ..$ b1: num 21 # success
  .. ..$ b2: num 22 # success
```
```r
> devtools::session_info()
Session info ------------------------------------------------------------------
 setting  value
 version  R version 3.4.3 (2017-11-30)
 system   x86_64, mingw32
 ui       RTerm
 language (EN)
 collate  English_United States.1252
 tz       America/Chicago
 date     2018-03-11

Packages ----------------------------------------------------------------------
 package    * version    date       source
 base       * 3.4.3      2017-12-06 local
 compiler     3.4.3      2017-12-06 local
 datasets   * 3.4.3      2017-12-06 local
 devtools     1.13.5     2018-02-18 CRAN (R 3.4.3)
 digest       0.6.15     2018-01-28 CRAN (R 3.4.3)
 graphics   * 3.4.3      2017-12-06 local
 grDevices  * 3.4.3      2017-12-06 local
 magrittr   * 1.5        2014-11-22 CRAN (R 3.4.0)
 memoise      1.1.0      2017-04-21 CRAN (R 3.4.0)
 methods    * 3.4.3      2017-12-06 local
 purrr      * 0.2.4.9000 2018-03-11 Github (tidyverse/purrr@84ce1ad)
 rlang        0.2.0      2018-02-20 CRAN (R 3.4.3)
 rstudioapi   0.7        2017-09-07 CRAN (R 3.4.1)
 stats      * 3.4.3      2017-12-06 local
[...]

PR #475, opened 2018-03-11
Make transpose() help a bit more clear

jennybc:
I set out to eliminate the confusing use of "usually" here:

> `.names` For efficiency, `transpose()` usually inspects the first component of `.l` to determine the structure.

and tweaked a few other things while I was at it. I did read the code itself.

Motivated by #474.


hadley:
Thanks!

PR #487, opened 2018-04-05
Clarify docs for compact()

jennybc:
Current wording suggests only `NULL`s will be dropped but `compact()` actually calls `is_empty()`.


hadley:
Thanks!

Issue #466, opened 2018-02-19
rbernoulli produces incorrect answers with scalar `n` and vector `p`


jrnold:
The `rbernoulli` function will work, and produce incorrect answers when `n` is a scalar and `prob` is a vector with length > 1. 
```{rconsole}
> purrr::rbernoulli(1, p = c(0.1, 0.2, 0.3))
[1] FALSE FALSE FALSE
```
Since `rbernoulli` is implemented as
```{r}
stats::runif(n) > (1 - p)
```
the code above will only make one draw from the uniform distribution, but then recycle `n` to compare with `p` and return a vector the same size as `p`.  This is a pretty subtle bug and can easily burn users who may assume that `rbernoulli(1, p)` will draw one sample each for each value of `p`.

Base R seems to have the following semantics for distribution sampling functions:

- if `n` has length 1, treat as the number of samples
- if `n` has length > 1, use the length of `n` as the number of samples
- if `n == 0`, return a zero-length vector.

In `rbinom`, when `length(prob) > n`, it simply uses the first `n` values of `prob`.
```{r}
> rbinom(1, size = 1, prob = c(0.1, 0.2, 0.3))
[1] 0
```

The following checks in `rbernoulli` should address these issues:
```{r}
rbernoulli <- function(n, p = 0.5) {
  # cannot let runif catch problems with n because need the value of n in order
  # to truncate prob
  if (length(n) == 0) {
    return(logical(0))
  } else if (length(n) == 1 && n == 0) {
    return(logical(0))
  } else if (length(n) > 1) {
    n <- length(n)
  }
  # truncate to at most n probs
  if (length(p) > n) {
    p <- p[seq_len(n)]
  }
  stats::runif(n) > (1 - p)
}
```
```{rconsole}
> rbernoulli(0)
logical(0)
> rbernoulli(NULL)
logical(0)
> rbernoulli(1)
[1] FALSE
> rbernoulli(5)
[1]  TRUE FALSE FALSE  TRUE  TRUE
> rbernoulli(2, c(0.1, 0.2))
[1] FALSE FALSE
> rbernoulli(c("a", "b"), c(0.1, 0.2))
[1] FALSE FALSE
> rbernoulli(1, c(0.1, 0.2))
[1] FALSE
```

hadley:
Want to do a PR for this?

jrnold:
PR #501 fixes this issue.

PR #501, opened 2018-05-07
fix rbernoulli function

jrnold:
rbernoulli was generating incorrect results due to recycling of n and p arguments. Make argument handling more consistent with other r* distribution functions.

This PR fixes issue #466.

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -73,6 +73,19 @@ names2 <- function(x) {
 #' rbernoulli(10)
 #' rbernoulli(100, 0.1)
 rbernoulli <- function(n, p = 0.5) {
+  # cannot let runif catch problems with n because need the value of n in order
+  # to truncate prob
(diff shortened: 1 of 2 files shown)


lionel-:
This needs a bit more work because we should implement the tidyverse recycling rules here (with the exception of the scalar `n` case). Empty vectors are contagious and recycling is only done for length 1 vectors. Anything else is an error. I think something like this should work:

```r
`%<-%` <- zeallot::`%<-%`

if (length(n) == 1) {
  p <- vctrs::vec_recycle(p, n)
} else {
  c(n, p) %<-% vctrs::vec_recycle_common(n, p)
}
```

We could wait until vctrs is on CRAN or you could try to implement all the rules if you'd like this merged for the next release.

jrnold:
If purrr will depend on vctrs, it makes sense to wait until vctrs is on CRAN.

PR #517, opened 2018-06-20
add condition of length 1 to map_*

### LINKED ISSUE #518: map_* expects a vector of length 1
This isn't really documented in the `map_*` that these functions expect a length 1 vector.

It is disconcerting, b/c the `map_df*` functions don't care how big the resulting objects are, they just get merged together.

Small pull request to update the documentation.

rmflight:
This makes it clearer that only a single value can be returned, which is kind of unexpected, honestly.

Fixes #518

### DIFF
--- a/R/map.R
+++ b/R/map.R
@@ -13,7 +13,7 @@
 #'   determines which elements of `.x` are transformed with `.f`.
 #'
 #' * `map_lgl()`, `map_int()`, `map_dbl()` and `map_chr()` return
-#'   vectors of the corresponding type (or die trying).
+#'   vectors of *length 1* of the corresponding type (or die trying).
 #'
 #' * `map_dfr()` and `map_dfc()` return data frames created by
 #'   row-binding and column-binding respectively. They require dplyr
--- a/R/map.R
+++ b/R/map.R
@@ -12,8 +12,13 @@
 #'   The `_if` and `_at` variants take a predicate function `.p` that
 #'   determines which elements of `.x` are transformed with `.f`.
 #'
-#' * `map_lgl()`, `map_int()`, `map_dbl()` and `map_chr()` return
-#'   vectors of *length 1* of the corresponding type (or die trying).
+#' * `map_lgl()`, `map_int()`, `map_dbl()` and `map_chr()` each return
+#'    an atomic vector of the indicated type (or die trying).
+#'
+#'    The return value of `.f` must be of length one for each element of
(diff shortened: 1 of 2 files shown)


jennybc:
How about something along these lines?

Current:

```
#' * `map_lgl()`, `map_int()`, `map_dbl()` and `map_chr()` return
#'   vectors of *length 1* of the corresponding type (or die trying).
```

Proposal:

`map_lgl()`, `map_int()`, `map_dbl()` and `map_chr()` each return an atomic vector of the indicated type (or die trying). This implies that the result of applying `.f` to each element of the input must be of length one or, if a `.default` value is given, absent or of length zero. See [as_mapper()] for more on `.default`.

lionel-:
> if a .default value is given, absent or of length zero

This is only the case when plucking I think.

jennybc:
@lionel- OK, true and you would know better than I. But I've also always felt that it's unfortunate that the availability of `.default` is so hard to learn about. It appears nowhere in the help file we are currently discussing. Hence my interest in raising its visibility. But, yes, obviously it needs to be correct.

jennybc:
Revised proposal:

`map_lgl()`, `map_int()`, `map_dbl()` and `map_chr()` each return an atomic vector of the indicated type (or die trying). This implies that the result of applying `.f` to each element of the input must be of length one. When `.f` uses the shortcuts for creating an extractor function, a `.default` value can be given to handle values that are absent or empty. See [as_mapper()] for more on `.default`.

rmflight:
That looks better to me.

Thanks @jenny!

jennybc:
Update: I see that `.default` is mentioned at the end of the `.f` section. I still think it may be worth repeating it, though.

@rmflight I'm going to make another commit here, but it will still be your PR.

rmflight:
I'm really not worried about who's PR it is, I just wanted clearer
documentation.

hadley:
If the length 1 constraint is a surprise we should emphasise that the output length is always the same as the input length.

rmflight:
I think that is covered by this sentence:

> The return value of `.f` must be of length one for each element of `.x`.

I don't think I've ever seen this **explicitly** in the various presentations I've skimmed over on `purrr`, but I've never seen a live one so it could have been emphasized, just not on the slides.

I would be willing to write a vignette that outlines this particular problem to add to the documentation if others think that would be helpful. I know I keep trying to do it, even though every time I've done it I get the error and go "d'oh".

lionel-:
Thanks!

Issue #482, opened 2018-03-29
Strict pluck


hadley:
In base R, `[[` sometimes returns an error and sometimes returns `NULL`:

| `row[[col]]` | Zero-length | OOB (int)  | OOB (chr) | Missing  |
|--------------|-------------|------------|-----------|----------|
| `NULL`       | `NULL`      | `NULL`     | `NULL`    | `NULL`   |
| Atomic       | Error       | Error      | Error     | Error    |
| List         | Error       | Error      | `NULL`    | `NULL`   |

`pluck()` eliminates this inconsistently by (effectively) always return `NULL`. We need a strict equivalent that eliminates the inconsistency in the opposite direction: by always throwing an error.

hadley:
We could call it `chuck()` or `suck()`

daniel-barnett:
Were you thinking something like the following (but without the `strict` parameter)? If so, I can clean it up and make a PR. 

```r
test.list <- list(a = 1:5, b = letters[1:5], c = list(abc = 123))
test.atom <- 1:10

> # strict_pluck()
> pluck(NULL, integer(0), strict = TRUE)
Error: `.x` must not be NULL.
> pluck(NULL, 100, strict = TRUE)
Error: `.x` must not be NULL.
> pluck(NULL, "adsaf", strict = TRUE)
Error: `.x` must not be NULL.
> pluck(NULL, NA_integer_, strict = TRUE)
Error: `.x` must not be NULL.

> pluck(test.list, integer(0), strict = TRUE)
Error: Can't find index.
> pluck(test.list, 100, strict = TRUE)
Error: Can't find index.
> pluck(test.list, "adsaf", strict = TRUE)
Error: Can't find index.
> pluck(test.list, NA_integer_, strict = TRUE)
Error: Can't find index.

> pluck(test.atom, integer(0), strict = TRUE)
Error: Can't find index.
> pluck(test.atom, 100, strict = TRUE)
Error: Can't find index.
> pluck(test.atom, "adsaf", strict = TRUE)
Error: Can't find index.
> pluck(test.atom, NA_integer_, strict = TRUE)
Error: Can't find index.
```

hadley:
Exactly!

hadley:
A PR would be great 😄

daniel-barnett:
So, I also have this working for environments and S4 objects. For environments, it errors if the object doesn't exist in the environment, and for S4 if the object doesn't have the desired slot. This would the desired behaviour, right?

[...]

Issue #480, opened 2018-03-27
pluck() converts zero-length objects to NULL


VincentGuyader:
HI, 

Hello I noticed something that seems strange to me. Dont know if it's a bug, but I dont understand the result.
`map("result") ` return somthing wrong (to me), but `map(~.x$result)` is OK

this is a minimal example : 

```
library(purrr)
> Warning: le package 'purrr' a Ã©tÃ© compilÃ© avec la version R 3.4.4
list(list(result=data.frame(),error=1)) %>% map("result") 
> [[1]]
> NULL   # WHY ?
list(list(result=data.frame(),error=1)) %>%  map(~`$`(.x,"result"))
> [[1]]
> data frame with 0 columns and 0 rows
list(list(result=data.frame(),error=1)) %>%  map(~.x$result)
> [[1]]
> data frame with 0 columns and 0 rows
```
Any idea why ?

Regards

```
> sessionInfo()
R version 3.4.3 (2017-11-30)
Platform: x86_64-w64-mingw32/x64 (64-bit)
Running under: Windows >= 8 x64 (build 9200)

Matrix products: default

locale:
[1] LC_COLLATE=French_France.1252  LC_CTYPE=French_France.1252   
[3] LC_MONETARY=French_France.1252 LC_NUMERIC=C                  
[5] LC_TIME=French_France.1252    

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
 [1] shiny_1.0.5     bindrcpp_0.2    forcats_0.3.0   stringr_1.3.0   dplyr_0.7.4    
 [6] purrr_0.2.4     readr_1.1.1     tidyr_0.8.0     tibble_1.4.2    ggplot2_2.2.1  
[11] tidyverse_1.2.1 tictoc_1.0      cleanser_0.1   

loaded via a namespace (and not attached):
 [1] Rcpp_0.12.16      lubridate_1.7.2   lattice_0.20-35   rprojroot_1.3-2  
 [5] assertthat_0.2.0  digest_0.6.15     psych_1.7.8       mime_0.5         
 [9] R6_2.2.2          cellranger_1.1.0  plyr_1.8.4        backports_1.1.2  
[13] reprex_0.1.2      evaluate_0.10.1   httr_1.3.1        pillar_1.2.1     
[17] rlang_0.2.0       settings_0.2.4    lazyeval_0.2.1    readxl_1.0.0     
[...]

PR #522, opened 2018-07-10
Strict pluck #482

daniel-barnett:
This is an implementation of strict pluck, so-called `chuck` from #482. 

Once we have agreed on everything, I'll also add some examples and descriptions to the help.

Since this is based on pluck, it seems like #480 is involved here too.


src/extract.c
```diff
@@ -66,10 +66,17 @@ int find_offset(SEXP x, SEXP index, int i) {
   }
 }
 
-SEXP extract_vector(SEXP x, SEXP index_i, int i) {
+SEXP extract_vector(SEXP x, SEXP index_i, int i, int strict) {
   int offset = find_offset(x, index_i, i);
```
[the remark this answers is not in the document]
daniel-barnett: I've adjusted all of this so it gives specific messages for each type of problem in `find_offset()`; however, for the error messages, which way do you think is better out of one or two line errors?

```{r}
chuck(1:10, Inf)
#> Error: Each accessor must be finite:
#> * Accessor 1 is Inf

chuck(1:10 Inf)
#> Error: Accessor 1 must be finite, not Inf
```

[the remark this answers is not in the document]
hadley: The second - I slightly prefer the wording, and the former suggests we might warn for any Inf, not just the first we encounter (which while nice, I don't think that would be worth the implementation complexity)

src/extract.c
```diff
@@ -5,48 +5,124 @@
...
+          "Index %i must be greater than 0, not %.0f.",
+          i + 1,
+          val + 1
+        );
+      else
+        return -1;
+    } else if (val >= n) {
+      if (strict)
+        Rf_errorcall(R_NilValue,
+          "Index %i exceeds the length of object being plucked (%.0f > %i).",
```
hadley: Why `%.0f` and not `%i`?

src/extract.c
```diff
@@ -5,48 +5,124 @@
 #include "backports.h"
 #include <string.h>
 
-int find_offset(SEXP x, SEXP index, int i) {
+int find_offset(SEXP x, SEXP index, int i, int strict) {
   if (Rf_length(index) > 1) {
-    Rf_errorcall(R_NilValue, "Index %i must have length 1", i + 1);
+    Rf_errorcall(R_NilValue,
+      "Index %i must have length 1, not %i.",
```
hadley: I think these messages are a bit easier to write/read if they all start in the same way, either with "Invalid object:" or "Invalid index (%i):".

hadley:
Feel free to start on the documentation - I think you should document `chuck()` with `pluck()` (i.e. with `@rdname chuck`, and then just tweak the existing pluck docs)

src/extract.c
daniel-barnett: Because of `double val = REAL(index)[0];`, using `%i` gives `nvalid index (1): index must be greater than 0, not 536870913.`.

hadley: Ah got it.

src/extract.c
```diff
@@ -5,48 +5,124 @@
...
+        );
+      else
+        return -1;
+    }
 
     val--;
-    if (val < 0 || val >= n)
-      return -1;
+    if (val < 0) {
+      if (strict)
```
lionel-: General comment: can you add curly braces even to single expressions please?

daniel-barnett: Did you want me to add those for the existing code that I haven't yet modified too?

hadley: @lionel- I don't think we use that style anymore, and no parens is fine for simple expressions.

lionel-: Ah I wasn' aware, Jim, Gabor and I still use it. It's consistent with our R style and easier to read since we use 2-level indents. Especially with multi-line single expressions such as here. But nevermind then.

jennybc: It can also have implications for covr, I think? That is, the presence of the brackets helps covr do its job in some weird edge cases.

hadley: I thought the edge cases were fixed and this is officially blessed @jimhester style

jennybc: Maybe I am behind the times then.

jimhester: The officially blessed jimhester style is to always use braces for single expressions.

The edge cases in covr are fixed, but I think it is still better to use braces, it is too easy to accidentally forget to add the braces in the future and write something like the following, then not realize the second if is unconditionally evaluated.
```r
if (val < 0)
  print(val)
  if (strict)
```

lionel-:
Don't worry about the latest comments @daniel-barnett, I'm going to finish the PR because I need it merged before working on other pluck bugs and features.

lionel-:
Refactored the code to put all the checking logic at the end of the file. Merged several code paths to get more consistent error checking and make it easier to maintain. Improved consistency of error messages.

@hadley Could you review the last commit which adds documentation please? I have moved `attr_getter()` in its own topic so as not to distract from the important stuff.

R/pluck.R
```diff
@@ -1,32 +1,24 @@
...
 #'
 #' @description
 #'
-#' This is a generalised form of `[[` which allows you to index deeply
-#' and flexibly into data structures. It supports R standard accessors
-#' like integer positions and string names, and also accepts arbitrary
-#' accessor functions, i.e. functions that take an object and return
-#' some internal piece.
+#' `pluck()` and `chuck()` implement a generalised form of `[[` that
+#' allow you to index deeply and flexibly into data structures. While
```
hadley: Remove "while"

R/pluck.R
```diff
@@ -1,32 +1,24 @@
...
+#' `pluck()` consistently returns `NULL` when an element does not
+#' exist, `chuck()` always throws an error in that case.
 #'
-#' `pluck()` is often more readable than a mix of operators and
-#' accessors because it reads linearly and is free of syntactic
-#' cruft. Compare: \code{accessor(x[[1]])$foo} to `pluck(x, 1,
-#' accessor, "foo")`.
+#' * You can pluck or chuck with standard accessors like integer
```
hadley: Maybe these should move down to details?

R/pluck.R
```diff
@@ -38,8 +30,16 @@
 #'   particular, if your accessors are stored in a list, you can
 #'   splice that in with `!!!`.
 #' @param .default Value to use if target is empty or absent.
-#' @keywords internal
-#' @export
+#'
+#' @details
+#'
+#' Since it handles arbitrary accessor functions, `pluck()` is a type
```
hadley: I don't understand this paragraph. Maybe it can  be removed?

lionel-:
Thanks again @daniel-barnett !

Issue #514, opened 2018-06-11
some


ISPritchin:
Documentation "Do some elements of a list satisfy a predicate?"
Why this function returns NA in this case?
```{r}
some(c(NA, -2), ~ .x < 0)  #NA
```
I read the source code.
It follows from the code that if NA is before the satisfying element of the vector, NA will be returned,
However, -2 satisfies the condition.

batpigandme:
reprex
[it seems as though it returns NA if the first element is NA, or if the condition isn't satisfied and contains NA.
``` r
library(purrr)
some(c(NA, -2), ~ .x < 0)
#> [1] NA
some(c(-2, NA), ~ .x < 0)
#> [1] TRUE
some(c("NA", -2), ~ .x < 0)
#> [1] TRUE
some(c("NA", NA), ~ .x < 0)
#> [1] NA
some(c(2, NA), ~ .x < 0)
#> [1] NA
```
Created on 2018-06-11 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).

ISPritchin:
Expected behavior:
1. If the condition is TRUE for at least one element, return TRUE.
2. If FALSE for all and has NA - return NA (We do not have information about NA)
3 Otherwise - FALSE

PR #523, opened 2018-07-10
Fixes oddities with NAs and some() #514

daniel-barnett:
Changes `some` to follow logic as seen in #514, i.e.

> If the condition is TRUE for at least one element, return TRUE.
    If FALSE for all and has NA - return NA (We do not have information about NA)
    Otherwise - FALSE

### DIFF
--- a/R/every-some.R
+++ b/R/every-some.R
@@ -26,10 +26,12 @@ every <- function(.x, .p, ...) {
 #' @rdname every
 some <- function(.x, .p, ...) {
   .p <- as_mapper(.p, ...)
+  no_na <- TRUE
   for (i in seq_along(.x)) {
     val <- .p(.x[[i]], ...)
     if (is_true(val)) return(TRUE)
-    if (anyNA(val)) return(NA)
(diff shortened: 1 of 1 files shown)


R/every-some.R
```diff
@@ -26,10 +26,12 @@ every <- function(.x, .p, ...) {
 #' @rdname every
 some <- function(.x, .p, ...) {
   .p <- as_mapper(.p, ...)
+  no_na <- TRUE
```
hadley: What if we did something like this?

```R
val <- NA
...

val <- val || .p(.x[[i]], ...)

...

return(val)
```

(let me know if I've skipped too many steps so this doesn't make sense)

hadley:
Looks good! Can you please add a bullet to NEWS? It should briefly [describe the change](http://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`.

daniel-barnett:
This should be fine now (sorry, forgot to add a comment along with my commit!)

PR #526, opened 2018-08-01
Refine safely() return value documentation (#488)

### LINKED ISSUE #488: Documentation: safely Doesn't Always Have One NULL Value
The documentation for `safely` inaccurately states that one value in the list is always `NULL` (https://github.com/tidyverse/purrr/blob/84ce1ad00a736a2e6a75acdc9872af74f6866cf0/R/output.R#L13-L14).  With the `otherwise` argument, both values may be non-NULL.

```r
> a <- safely(function(x) stop("foo"), otherwise="bar")
> a("bloo")
$result
[1] "bar"

$error
<simpleError in .f(...): foo>
```

mikmart:
Reworded the documented return value for `safely()`. This fixes #488.

Suggested new wording:

> `safely`: wrapped function instead returns a list with components `result` and `error`. If an error occurred, `error` is an `error` object and `result` has a default value (`otherwise`). Else `error` is `NULL`.

Since the default value of `otherwise` for `safely()` is `NULL`, this wording also covers the previous "one is always `NULL`" behaviour when an `otherwise` value is not explicitly given.

As to when there's no error, I think it's probably sufficient to not explicitly state that then `return` contains the return value of the wrapped function?

### DIFF
--- a/R/output.R
+++ b/R/output.R
@@ -11,7 +11,9 @@
 #' @param otherwise Default value to use when an error occurs.
 #'
 #' @return `safely`: wrapped function instead returns a list with
-#'   components `result` and `error`. One value is always `NULL`.
+#'   components `result` and `error`. If an error occurred, `error` is
+#'   an `error` object and `result` has a default value (`otherwise`).
+#'   Else `error` is `NULL`.
 #'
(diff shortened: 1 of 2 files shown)


lionel-:
Thanks!

PR #528, opened 2018-08-04
Update set_names doc

jrnold:
The details section for `set_names()` was out of date, so I removed it. `set_names()` is no longer simply a wrapper of `setNames()`, and it includes additional features.


lionel-:
Thanks!

PR #535, opened 2018-08-21
Fix librarys typo in other-langs vignette

batpigandme:
Typo fix. 

Didn't rebuild the vignette, since I figured it'll happen in the next round. But, I'm happy to do so, if you think I should.


hadley:
Thanks!

Issue #429, opened 2017-12-18
.id argument ignored in imap_dfr()/imap_dfc()?


eaurele:
https://github.com/tidyverse/purrr/blob/62b135a13a64a4dc0d7c240bb43bdc6005c95c92/R/imap.R#L50

```r
imap_dfr <- function(.x, .f, ..., .id = NULL) {
  .f <- as_mapper(.f, ...)
  map2_dfr(.x, vec_index(.x), .f, ...)
}
```

I don't seem to be able to make use of `.id` in `imap_dfr()`. Looking at the source code, the `.id` arg seems not to be handled at all. Is it the case or am I missing something?

cderv:
It seems you are right. To illustrate, here is a dummy reprex with a custom to argument fonction to use with `imap` and `map2`
``` r
library(purrr)
#> Warning: le package 'purrr' a Ã©tÃ© compilÃ© avec la version R 3.4.2
library(tibble)

as_tibble_custom <- function(value, name = NULL) {
    tab <- as_tibble(value)
    if (! is.null(name)) {
        tab$col_id = paste0("col_", name)
    }
    tab
}
# same as imap and it works
list(a = 1, b = 2, c = 3) %>%
    map2_dfr(names(.), as_tibble_custom)
#> # A tibble: 3 x 2
#>   value col_id
#>   <dbl>  <chr>
#> 1     1  col_a
#> 2     2  col_b
#> 3     3  col_c
list(a = 1, b = 2, c = 3) %>%
    map2_dfr(names(.), as_tibble_custom, .id = "name")
#> # A tibble: 3 x 3
#>    name value col_id
#>   <chr> <dbl>  <chr>
#> 1     a     1  col_a
#> 2     b     2  col_b
#> 3     c     3  col_c
# with imap it does not work
list(a = 1, b = 2, c = 3) %>%
    imap_dfr(as_tibble_custom)
#> # A tibble: 3 x 2
#>   value col_id
#>   <dbl>  <chr>
#> 1     1  col_a
#> 2     2  col_b
#> 3     3  col_c
list(a = 1, b = 2, c = 3) %>%
    imap_dfr(as_tibble_custom, .id = "name")
#> # A tibble: 3 x 2
#>   value col_id
#>   <dbl>  <chr>
#> 1     1  col_a
#> 2     2  col_b
#> 3     3  col_c
```

In the source code, `.id` is not pass to `map2_dfr` as you pointed out. 

[...]

Issue #417, opened 2017-11-22
modify() not maintaining class for atomic vectors


MrFlick:
The documentation for `modify` says that the `.x` parameter is "A list or atomic vector." and the "Value" section says that the functions will return "An object the same class as `.x`" However for atomic vectors, it seems to return a list, and not a vector of the same class as `.x`.
```
x <- c("a","b","c","d","e")
y <- c(1, 2, 3, 4, 5)
z <- c(T, F, T, F, T)

class(modify(x,function(x) "k"))
# [1] "list"      ## not class(x)

class(modify(y,function(x) 10))
# [1] "list"     ## not class(y)

class(modify_if(x, z, function(x) "k"))
# [1] "list"     ## not class(x)

class(modify_if(y, z, function(x) x+10))
# [1] "list"     ## not class(y)
```
The help page also says that `modify()` is basically the same as `x[] <- map(x, .f)` and ifs that's the case i'd expect `map()` to return a list, But even on the `?map` help page it calls out: "See the `modify()` family for versions that return an object of the same type as the input."

Tested with `purrr_0.2.4` and `R version 3.4.1 (2017-06-30)`

lionel-:
Should be fixed by https://github.com/tidyverse/purrr/pull/463

stufield:
**Question**: has this issue been resolved? I stubbed my toe on it today using the CRAN version `purrr_0.2.5` and the list conversion is still happening. I reverted to the `dev` version, which appears to be `purrr_0.2.4.9000`, and it's fixed there. My confusion arises from my assumption that `0.2.5` comes *after* `0.2.4.9000` in the version schema. Am I wrong? Or worse, was it fixed, then reverted? Or should the development version actually read `0.2.5.9000`. In which case I humbly withdraw my question ;)

yutannihilation:
It seems some commits included in `v0.2.5` are forgot to be merged into master...

https://github.com/tidyverse/purrr/compare/v0.2.4...v0.2.5

PR #542, opened 2018-08-30
Fix outdated version number

ilarischeinin:
As @cderv mentions [here](https://github.com/tidyverse/purrr/issues/429#issuecomment-407148827), an odd thing has happened with purrr's version numbers.

The current CRAN version is 0.2.5, whereas the master branch on GitHub is 0.2.4.9000. The fixes that were implemented between 0.2.4 and 0.2.4.9000 are not included in the 0.2.5 release. That one was a "maintenance release following the release of dplyr 0.7.5", which I guess came in though some other pipeline than the "normal" GitHub path.

I noticed this as I ran into the bug reported in #429 (parameter `.id` is ignored in `imap_dfr()`), for which a fix was merged on Feb 5 in #454. However, while this fix is on the current master branch (0.2.4.9000), it is not included in CRAN's 0.2.5.

The result is that I cannot make my package depend on the fixed version, as 0.2.5 > 0.2.4.9000.

This PR hence bumps the version number on GitHub from 0.2.4.9000 to 0.2.5.9000. Then it is possible to make another package depend on purrr 0.2.5.9000, which does include the needed fix.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: purrr
 Title: Functional Programming Tools
-Version: 0.2.4.9000
+Version: 0.2.5.9000
 Authors@R: c(
     person("Lionel", "Henry", , "lionel@rstudio.com", c("aut", "cre")),
     person("Hadley", "Wickham", , "hadley@rstudio.com", "aut"),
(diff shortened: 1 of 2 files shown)


stufield:
This issue is also relevant to #417, fixes between `0.2.4` and `0.2.4.9000` are not included in the `0.2.5` release, the maintenance of object structures being one of them. Thus my confusion in the comments of the issue about the ordering of the release versions.

lionel-:
Thanks!

PR #554, opened 2018-09-26
Preserve classes in pmap()

### LINKED ISSUE #358: pmap strips `Date`
As title - `purrr::pmap` seems to strip date class.

``` r
print_date = function(b) b

b = as.Date("2012-03-22")

purrr::pmap(list(b), print_date)
#> [[1]]
#> [1] 15421
```

Not sure if this is intended? Caught me out though :)

mikmart:
Following [this suggestion](https://github.com/tidyverse/purrr/issues/358#issuecomment-386840204), changed the calls constructed by `pmap()` to use two double brackets for indexing rather than a single double bracket with a vector argument. This retains class for S3 atomic vector types (whose `[[` method preserves the class) in the input. Fixes #358.

However, list-based S3 classes, e.g. `POSIXlt` still fail.

### DIFF
--- a/src/map.c
+++ b/src/map.c
@@ -174,7 +174,11 @@ SEXP pmap_impl(SEXP env, SEXP l_name_, SEXP f_name_, SEXP type_) {
   SEXP i = Rf_install("i");
   SEXP one = PROTECT(Rf_ScalarInteger(1));
 
-  // Construct call like f(.x[[c(1, i)]], .x[[c(2, i)]], ...)
+  // Construct call like f(.l[[1]][[i]], .l[[2]][[i]], ...)
+  //
+  // Currently accessing S3 vectors in a list like .l[[c(1, i)]] will not
+  // preserve the class (cf. #358).
+  //
   // We construct the call backwards because can only add to the front of a
   // linked list. That makes PROTECTion tricky because we need to update it
   // each time to point to the start of the linked list.
@@ -186,10 +190,10 @@ SEXP pmap_impl(SEXP env, SEXP l_name_, SEXP f_name_, SEXP type_) {
   for (int j = m - 1; j >= 0; --j) {
     int nj = Rf_length(VECTOR_ELT(l_val, j));
 
-    // Construct call like .l[[c(j, i)]]
+    // Construct call like .l[[j]][[i]]
     SEXP j_ = PROTECT(Rf_ScalarInteger(j + 1));
-    SEXP ji_ = PROTECT(Rf_lang3(Rf_install("c"), j_, nj == 1 ? one : i));
-    SEXP l_ji = PROTECT(Rf_lang3(R_Bracket2Symbol, l, ji_));
+    SEXP l_j = PROTECT(Rf_lang3(R_Bracket2Symbol, l, j_));
+    SEXP l_ji = PROTECT(Rf_lang3(R_Bracket2Symbol, l_j, nj == 1 ? one : i));
 
     REPROTECT(f_call = Rf_lcons(l_ji, f_call), fi);
     if (has_names && CHAR(STRING_ELT(l_names, j))[0] != '\0')
--- a/tests/testthat/test-map_n.R
+++ b/tests/testthat/test-map_n.R
@@ -64,3 +64,9 @@ test_that("pmap on data frames performs rowwise operations", {
 test_that("pmap works with empty lists", {
   expect_identical(pmap(list(), identity), list())
 })
+
+test_that("preserves S3 class of input vectors (#358)", {
+  date <- as.Date("2018-09-27")
+  expect_equal(pmap(list(date), identity), list(date))
+  expect_output(pwalk(list(date), print), format(date))
+})
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # purrr 0.2.4.9000
 
+* `pmap()` and `pwalk()` now preserve class for inputs of `factor`, `Date`, `POSIXct`
+  and other atomic S3 classes with an appropriate `[[` method (#358, @mikmart).
+
 * `compose()` now supports composition with lambdas (@ColinFay, #556)
 
 * Fixed a `pmap()` crash with empty lists on the Win32 platform (#565).
--- a/R/map.R
+++ b/R/map.R
@@ -209,9 +209,6 @@ map_dfc <- function(.x, .f, ...) {
 #' @export
 #' @rdname map
 walk <- function(.x, .f, ...) {
-  .f <- as_mapper(.f, ...)
-  for (i in seq_along(.x)) {
-    .f(.x[[i]], ...)
-  }
+  map(.x, .f, ...)
   invisible(.x)
 }
--- a/R/map2-pmap.R
+++ b/R/map2-pmap.R
@@ -147,7 +147,7 @@ map2_df <- map2_dfr
 #' @export
 #' @rdname map2
 walk2 <- function(.x, .y, .f, ...) {
-  pwalk(list(.x, .y), .f, ...)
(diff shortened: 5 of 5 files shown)


mungojam:
Does this cover `pwalk` too?

mikmart:
@mungojam not as it is, as the problem with `pwalk()` comes from the implementation of `transpose()`. But maybe this should, I'll have a look.

mikmart:
Okay I don't have any ideas on fixing `transpose()` and thus `pwalk()` directly. However, if we just rewrite `pwalk()` using `pmap()` (latest commit) it should also be fixed.

@hadley do you think it's OK to change `pwalk()` to call `pmap()`?

src/map.c
```diff
@@ -174,7 +174,7 @@ SEXP pmap_impl(SEXP env, SEXP l_name_, SEXP f_name_, SEXP type_) {
   SEXP i = Rf_install("i");
   SEXP one = PROTECT(Rf_ScalarInteger(1));
 
-  // Construct call like f(.x[[c(1, i)]], .x[[c(2, i)]], ...)
+  // Construct call like f(.x[[1]][[i]], .x[[2]][[i]], ...)
```
hadley: Can you add a comment here saying that `x[[c(1, i)]]` is buggy for S3 vectors

tests/testthat/test-map_n.R
```diff
@@ -60,3 +60,17 @@ test_that("pmap on data frames performs rowwise operations", {
   expect_is(pmap_chr(mtcars2, paste), "character")
   expect_is(pmap_raw(mtcars2, function(mpg, cyl) as.raw(cyl)), "raw")
 })
+
+test_that("preserves S3 class", {
+  fctr <- factor("a")
```
hadley: I think you only need to test a single S3 class here, since the other options don't cause the tests to exercise any more code.

hadley:
The downside of using the same code is that it makes `pwalk()` less efficient than it could be, because it has to initialise and store the results of each call, even when that isn't necessary. But I could probably be persuaded if you did a little benchmarking showing that had hardly any affect.

OTOH I think `transpose()` will need a complete rewrite, once I've fully internalised the lessons from [vctrs](https://vctrs.r-lib.org), so it's also ok to leave as is for now, and we'll fix more holistically later.

mikmart:
Oh yeah I see now. I'll do some benchmarks to see what the hit would be like.

mikmart:
@hadley here's one benchmark (with a contrived use-case):
``` R
library(purrr)

pwalk2 <- function(.l, .f, ...) {
  pmap(.l, .f, ...)
  invisible(.l)
}

res <- bench::press(
  i = 0:6,
  {
    args <- list(n = rep(10^(6 - i), 10^i))
    bench::mark(
      pwalk(args, runif),
      pwalk2(args, runif)
    )
  }
)

res
#> # A tibble: 14 x 15
#>    expression     i      min     mean   median      max `itr/sec` mem_alloc
#>  * <chr>      <int> <bch:tm> <bch:tm> <bch:tm> <bch:tm>     <dbl> <bch:byt>
#>  1 pwalk(arg~     0  29.11ms  29.77ms  29.39ms  32.86ms    33.6      7.63MB
#>  2 pwalk2(ar~     0  29.14ms  31.42ms  29.52ms  57.91ms    31.8      7.63MB
#>  3 pwalk(arg~     1  29.19ms  30.18ms  30.28ms  33.13ms    33.1      7.65MB
#>  4 pwalk2(ar~     1  28.62ms  29.35ms  29.02ms  31.57ms    34.1      7.65MB
#>  5 pwalk(arg~     2   31.6ms  32.55ms  32.02ms  36.14ms    30.7      7.88MB
#>  6 pwalk2(ar~     2  28.64ms  29.25ms  28.94ms  31.24ms    34.2      7.88MB
#>  7 pwalk(arg~     3  32.46ms  33.03ms  32.81ms  35.63ms    30.3     10.12MB
#>  8 pwalk2(ar~     3  30.42ms  31.07ms  30.93ms  32.99ms    32.2     10.12MB
#>  9 pwalk(arg~     4  61.77ms  62.77ms  62.77ms  63.77ms    15.9     32.54MB
#> 10 pwalk2(ar~     4  47.58ms  49.49ms  48.24ms  56.41ms    20.2      32.5MB
#> 11 pwalk(arg~     5 560.56ms 560.56ms 560.56ms 560.56ms     1.78   244.52MB
#> 12 pwalk2(ar~     5 233.72ms 247.71ms 252.81ms  256.6ms     4.04   244.14MB
#> 13 pwalk(arg~     6    4.74s    4.74s    4.74s    4.74s     0.211    2.39GB
#> 14 pwalk2(ar~     6    2.36s    2.36s    2.36s    2.36s     0.423    2.38GB
#> # ... with 7 more variables: n_gc <dbl>, n_itr <int>, total_time <bch:tm>,
#> #   result <list>, memory <list>, time <list>, gc <list>
```

Real use-cases would probably include I/O operations, where I would imagine the time taken to interact with the output device would dominate the results anyway; so I didn't continue with testing as it looks like the `pmap()` implementation is noticeably faster in this simple case already.

mikmart:
Okay I tried some I/O with saving plots, too. Very little difference here, either:

``` r
library(purrr)

pwalk2 <- function(.l, .f, ...) {
  pmap(.l, .f, ...)
  invisible(.l)
}

f <- function(file, n) {
  pdf(file)
  plot(runif(n))
  on.exit({dev.off(); unlink(file)})
}

bench::press(
  i = 0:3,
  {
    tf <- map_chr(seq_len(10^i), ~ tempfile())
    args <- list(file = tf, n = rep(10^(4 - i), 10^i))
    bench::mark(
      pwalk(args, f),
      pwalk2(args, f)
    )
  }
)
#> Running with:
#>       i
#> 1     0
#> 2     1
#> 3     2
#> 4     3
#> Warning: Some expressions had a GC in every iteration; so filtering is
#> disabled.
#> # A tibble: 8 x 11
#>   expression     i      min     mean   median      max `itr/sec` mem_alloc
#>   <chr>      <int> <bch:tm> <bch:tm> <bch:tm> <bch:tm>     <dbl> <bch:byt>
#> 1 pwalk(arg~     0  36.43ms   38.7ms  37.15ms   42.6ms    25.8      1.56MB
#> 2 pwalk2(ar~     0  35.28ms  37.38ms  36.03ms  45.83ms    26.8    807.95KB
#> 3 pwalk(arg~     1  54.15ms  59.34ms  59.23ms   65.8ms    16.9      1.41MB
#> 4 pwalk2(ar~     1  52.02ms   56.2ms   54.7ms  62.13ms    17.8      1.41MB
#> 5 pwalk(arg~     2 282.76ms 282.76ms 282.76ms 282.76ms     3.54     7.88MB
#> 6 pwalk2(ar~     2 277.89ms 277.89ms 277.89ms 277.89ms     3.60     7.88MB
#> 7 pwalk(arg~     3    2.45s    2.45s    2.45s    2.45s     0.409   71.37MB
#> 8 pwalk2(ar~     3    2.42s    2.42s    2.42s    2.42s     0.414   71.36MB
#> # ... with 3 more variables: n_gc <dbl>, n_itr <int>, total_time <bch:tm>
```

Created on 2018-09-28 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0.9000).

hadley:
Ok, that's compelling!

Do you want to change the `walk2()` and `walk()` to wrappers around the map functions in order to be consistent?

mikmart:
Absolutely. Should be good now.

mikmart:
Perhaps the `walk()` changes should be decoupled from this PR? I could revert this to just include the `pmap()` fix.

lionel-:
Thanks!

PR #573, opened 2018-11-14
Support tidy dots in compose()

### LINKED ISSUE #564: Add tidy dots support to compose()
This is super minor, but the current behavior for passing a list of functions to `compose()` requires the usage of `invoke()` or `lift()`. It would be nice to add tidy dots support, allowing the list to be passed with the `!!!` operator:

```r
funs <- c( exp, log10, sqrt )
invoke( compose, funs )    # Works
lift( compose )(funs)      # Also works
compose( !!!funs )         # Error in !funs : invalid argument type
```

I believe the only modification required is to change [line 14 of compose.R](https://github.com/tidyverse/purrr/blob/43c3eb329001152762f618a62beec111702734b9/R/compose.R#L14) from

    fs <- lapply(list(...), match.fun)

to

    fs <- lapply(rlang::list2(...), match.fun)

Happy to submit a pull request.

lionel-:
Closes #564

### DIFF
--- a/R/compose.R
+++ b/R/compose.R
@@ -3,8 +3,9 @@
 #' @param ... Functions to apply in order from right to left. Formulas
 #'   are converted to functions in the usual way.
 #'
-#'   These dots support [tidy dots features][rlang::list2]. In particular, you can
-#'   splice lists of functions with `!!!`.
+#'   These dots support tidy dots features. In particular, if your
+#'   functions are stored in a list, you can splice that in with
+#'   `!!!`.
 #' @return A function
 #' @export
 #' @examples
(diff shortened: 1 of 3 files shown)


R/compose.R
```diff
@@ -2,6 +2,9 @@
 #'
 #' @param ... Functions to apply in order from right to left. Formulas
 #'   are converted to functions in the usual way.
+#'
+#'   These dots support [tidy dots features][rlang::list2]. In particular, you can
+#'   splice lists of functions with `!!!`.
```
jennybc: I would re-word to clarify who's doing what.

Proposal: "These dots support tidy dots features. In particular, if you pass a list of functions, it will be spliced with `!!!`.

lionel-: Done, thanks!

PR #579, opened 2018-11-23
Review genericity of modify()

lionel-:
* Use `[[<-` instead of `[<-`.

* Update documentation.


lionel-:
hmm, looks like we still need to override behaviour for base vectors to work around the base coercion rules:

```r
# Before
modify(1:3, ~ "foo")
#> Error: Can't coerce element 1 from a character to a integer

# After
modify(1:3, ~ "foo")
#> [1] "foo" "foo" "foo"
```

hadley:
Alternatively, we could possibly just live with that behaviour? It does make `modify` a more true wrapper around `[[`.

lionel-:
I think that would make it trickier to use with base types.

hadley:
I don't think that's terribly important.

Another option would be to use `vec_subset()` for non-objects, and then consider what it should do as a safer replacement for `[[`.  Either way, this feels like a vctrs issue, so I think we could implement just using `[[` for this release, and then come back to it for the next.

lionel-:
This means we'll relax the type constraints in this release, and tighten them again at the next? Shouldn't we just come back to it next time?

lionel-:
Or I'll just keep the methods for base vectors with the `[<-` implementation.

hadley:
Oh yeah, let's do that (i.e. just change the implementation of the default method)

PR #581, opened 2018-11-25
Added test on uncovered line in possibly()

jameslamb:
As of now, this portion of `possibly()` is not covered by this project's tests:

```
if (!quiet)
    message("Error: ", e$message)
```

Please consider this PR to add a test that covers it.

### DIFF
--- a/tests/testthat/test-output.R
+++ b/tests/testthat/test-output.R
@@ -37,6 +37,12 @@ test_that("possibly returns default value on failure", {
   expect_identical(possibly(log, NA_real_)("a"), NA_real_)
 })
 
+test_that("possibly emits a message on failure if quiet = FALSE", {
+  expect_message({
+    possibly(log, NA_real_, quiet = FALSE)("a")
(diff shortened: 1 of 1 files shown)


lionel-:
Thanks! Merging as is because it's unlikely the base R message will change, but normally you should make the unit test depend as little as possible on things that are not under our control, such as content of error messages from other packages.

jameslamb:
@lionel- sure, makes sense. In general I Like to use _some_ `regexp` with `expect_error()`, `expect_warning()`, and `expect_message()` because otherwise you get silent failures.

I think this is most important with `expect_error()` but fair point that this may have been overzealous with `expect_message()`.

Cheers!

lionel-:
Ideally the thrower creates a classed error that we can then check, unfortunately that's still rare. Here we could have generated the error message ourselves as well, like this: https://github.com/tidyverse/purrr/commit/0f5c6d7

jameslamb:
oh word, ok. I like that!

Yeah I rarely see R developers using classed errors :/

PR #582, opened 2018-11-25
removed unreachable line in make_call()

jameslamb:
Thank you for this awesome package! I hope you'll consider this PR to remove this unreachable line in internal function `make_call()`

image

Because `make_call()` is only called by `partial()` and `partial()` has this:

```
stopifnot(is.function(...f))
```

It's impossible for a value of `f` that is anything other than a function to reach that `make_call()` call.


lionel-:
Thanks!

PR #583, opened 2018-11-26
Fix `accumulate()` return value and documentation when `.init` is supplied

### LINKED ISSUE #569: accumulate: the output vector is longer than the input
The documentation for accumulate says that the output has the same length as the input list:

accumulate(.x, .f, ..., .init)
Value: "A vector the same length of .x with the same names as .x" 

This does not seem to be always the case. For example the lengths are equal here:
 
    > 1:5 %>% accumulate(max)
    [1] 1 2 3 4 5

But here the output is a unit longer as it takes the initial condition:

    > 1:5 %>% accumulate(max, .init = 1)
    [1] 1 1 2 3 4 5
Should the documentation be updated?

lionel-:
* Properly handle names when `.init` is supplied. The initial value gets the name `".init"`, which is consistent with the argument name and the new convention of prepending generated names with a dot. Does that sound reasonable?

* Clearer documentation about the return value. Closes #569.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -59,6 +59,9 @@ reduce2_right(.x = letters[1:4], .y = paste2, .f = c("-", ".", "-")) # working
 
 ## Minor improvements and fixes
 
+* Fixed ordering of names returned by `accumulate_right()`
+  output. They now correspond to the order of inputs.
+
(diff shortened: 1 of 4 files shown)


batpigandme:
LGTM in terms of doc clarification. Sorry I dropped the ball on that one.

PR #584, opened 2018-11-27
Retain dimnames in array_branch and array_tree

flying-sheep:
By not resetting `dim(array)` without need, the return value retains its dimnames

### DIFF
--- a/R/arrays.R
+++ b/R/arrays.R
@@ -50,7 +50,12 @@ array_branch <- function(array, margin = NULL) {
 
   if (length(margin) == 0) {
     list(array)
-  } else if (identical(as.integer(margin), 1L) && is.null(dim(array))) {
+  } else if (is.null(dim(array))) {
+    if (!identical(as.integer(margin), 1L)) {
+      stop(
+        "The passed array is 1D and does not have the margin(s) ",
+        toString(margin), call. = FALSE)
+    }
     as.list(array)
   } else {
     flatten(apply(array, margin, list))
--- a/R/arrays.R
+++ b/R/arrays.R
@@ -50,6 +50,8 @@ array_branch <- function(array, margin = NULL) {
 
   if (length(margin) == 0) {
     list(array)
+  } else if (identical(margin, 1) && is.null(dim(array))) {
+    as.list(array)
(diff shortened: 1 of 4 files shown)


lionel-:
Could you provide a motivation for this change please? Ideally the motivation would be provided in a NEWS bullet explaining the change and would be covered by unit tests.

flying-sheep:
Sure, done!

flying-sheep:
OK, this looks good to go!

tests/testthat/setup.R
```diff
@@ -0,0 +1 @@
+Sys.setlocale("LC_MESSAGES", "C")
```
lionel-: Why did you add this setup file?

R/arrays.R
```diff
@@ -50,6 +50,8 @@ array_branch <- function(array, margin = NULL) {
 
   if (length(margin) == 0) {
     list(array)
+  } else if (identical(margin, 1) && is.null(dim(array))) {
```
lionel-: The failure this is fixing was from an example, could you capture it in the unit tests as well please?

tests/testthat/setup.R
flying-sheep: Because otherwise the tests don’t run on my localized system (the german error messages don’t match the expected english ones)

Technically that change should go into another PR, but that’s a formality, so I thought I’ll only bother if you insist.

R/arrays.R
flying-sheep: Will do!

tests/testthat/setup.R
lionel-: Interesting, I wonder if this should be set by testthat

R/arrays.R
lionel-: Do you think the branch should be `if (is.null(dim(array)))`, and then we'd check for the margin and give an informative error message if it's not `1`?

By the way, `identical(1, 1L)` is `FALSE`. I think we need `is_integerish(margin, n = 1) && margin == 1`. Could you capture the integer case in the tests as well please?

lionel-:
Thanks!

PR #588, opened 2018-11-28
Add `.rev` argument to `compose()`

### LINKED ISSUE #323: FR: and_then()
Basically

```
and_then <- function(...) {
  do.call("compose", rev(list(...)))
}
```

lionel-:
Closes #323

I thought about a `.dir` argument but it seemed a bit confusing (does `.dir = "right"` start from the right or composes from left to right?).

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -64,6 +64,9 @@ reduce2_right(.x = letters[1:4], .y = paste2, .f = c("-", ".", "-")) # working
 
 ## Minor improvements and fixes
 
+* New `.rev` argument in `compose()`. If set to `FALSE`, the functions
+  are composed from left to right rather than right to left.
+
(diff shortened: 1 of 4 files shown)


lionel-:
@hadley I now see `detect()` has a `.right` argument documented like this:

```
#' @param .right If `FALSE`, the default, starts at the beginning
#'   of the vector and move towards the end; if `TRUE`, starts at the end
#'   of the vector and moves towards the beginning.
```

Looks like we should use this for instead of introducing yet a new argument name. WDYT?

lionel-:
Or should be `.right` / `.dir` for all of those functions (reduce, detect, compose). Probably the former since `.dir = "right"` is not very suggestive of starting from the right, it suggests going to the right which isn't correct for detect and compose. Also this way we don't break detect's interface.

PR #589, opened 2018-11-28
Follow style guide in error messages

lionel-:


lionel-:
Something like:

```r
abort_type_element <- function(elt, i, expected, actual = NULL) {
  actual <- actual %||% friendly_type_of(elt)
  msg <- sprintf("Element %s must be %s, not %s", i, expected, actual)

  abort(
    msg,
    "purrr_type_error_element",
    i = i,
    expected = expected,
    actual = actual
  )
}

abort_type_element(1:3, 10, "a foobaz object")
#> Error: Element 10 must be a foobaz object, not an integer vector

abort_type_element(1:3, 10, "a foobaz object", "a thing")
#> Error: Element 10 must be a foobaz object, not a thing
```

hadley:
The vctrs convention would be to call it something like `stop_bad_type()`. 

I don't care too much about the interface (I suspect it will have to be replaced by something powered by vctrs eventually), as long as it makes the code simpler and makes the tests less fragile.

PR #591, opened 2018-11-30
Require strict predicate functions in purrr

### LINKED ISSUE #470: Should predicate functions warn when called with a function that is not a predicate?
Passing a function that returns a logical vector as the predicate argument to predicate functions can cause unexpected results, as in the example below, from [this StackOverflow question](https://stackoverflow.com/questions/49149222/purrrs-every-doesnt-work-well-with-is-na).

``` r
purrr::every(list(1:10), is.na)
#> [1] TRUE
```

Perhaps adding a warning message if the predicate evaluates to a logical vector of length > 1 would be appropriate?

lionel-:
Closes #470.

`keep()`, `map_if()`, and variants now fail with an informative error message when the predicate function dose not return `TRUE` or `FALSE`.

This is only soft-deprecated for `every()` and `some()` as they were documented to have liberal truthness. Also the unit tests check that these functions return `NA` when the predicate has returned `NA`. I think it's best to deprecate this behaviour for consistency with predicate-based functionals within purrr, and because it simplifies the type of these functions (you can count on them returning TRUE/FALSE, missing values have to be dealt beforehand to avoid an early error).

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -64,9 +64,9 @@ reduce2_right(.x = letters[1:4], .y = paste2, .f = c("-", ".", "-")) # working
 
 ## Minor improvements and fixes
 
-* Functions taking predicates (`some()`, `every()`, `keep()`, etc) now
-  fail with an informative message when the return value is not a
-  scalar logical (#470).
+* Functions taking predicates (`map_if()`, `keep()`, `some()`,
+  `every()`, `keep()`, etc) now fail with an informative message when
+  the return value is not `TRUE` or `FALSE` (#470).
 
 * `pluck()` now supports primitive functions (#404).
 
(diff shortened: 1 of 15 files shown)


NEWS.md
```diff
@@ -64,6 +64,16 @@ reduce2_right(.x = letters[1:4], .y = paste2, .f = c("-", ".", "-")) # working
 
 ## Minor improvements and fixes
 
+* Functions taking predicates (`map_if()`, `keep()`, `some()`,
```
hadley: I think `TRUE` or `FALSE` is probably more accurate and easier to grasp than "scalar logical"

R/detect.R
```diff
@@ -51,7 +51,7 @@ detect <- function(.x, .f, ..., .right = FALSE) {
 #' @export
 #' @rdname detect
 detect_index <- function(.x, .f, ..., .right = FALSE) {
-  .f <- as_mapper(.f, ...)
+  .f <- as_predicate(.f, ..., .mapper = TRUE)
 
   for (i in index(.x, .right)) {
     if (is_true(.f(.x[[i]], ...))) return(i)
```
hadley: I think you can now remove the `is_true()`?

PR #592, opened 2018-11-30
Add if-else variants

### LINKED ISSUE #324: else clause in map_if
I fell that i order for the map_if function to be really useful should have the possibility of calling an "else" function if the outcome of .predicate is false.

By default it should leave the object iterated alone, and not return anything.

lionel-:
Closes #324


lionel-:
What do you mean? `map_if()` wasn't changed.

lionel-:
Ah you mean `map_if(x, p, f, .else = f2)`? It does read better:

```r
map_if_else(iris, is.factor, as.character, as.integer)

map_if(iris, is.factor, as.character, .else = as.integer)
```

hadley:
Yeah, I was thinking just adding the `.else` argument to `map_if()`

PR #594, opened 2018-12-05
Improve formals and printing of composed functions

### LINKED ISSUE #366: Friendlier compose()
It would be nice to have a "friendlier" version of `compose()` that:
- sets the formals to that of the first called function
- is more consistent with tidyverse semantics (i.e., supports splicing and formula syntax)
- has a more informative print method
- works as a drop-in replacement

To satisfy these desiderata (the second of which I presume is planned), I have something like the following in mind, which is a minor tweaking of the current implementation:
```R
library(rlang)

compose <- function(...) {
  fs <- lapply(dots_list(...), as_function)
  n <- length(fs)
  
  last <- as_closure(fs[[n]])
  `__call_last` <- function() {
    call_last <- mut_node_car(sys.call(-1), last)
    eval_bare(call_last, parent.frame(2))
  }
  `__rest` <- rev(fs[-n])
  
  set_attrs(
    `formals<-`(
      value = formals(last),
      function() {
        out <- `__call_last`()
        for (f in `__rest`)
          out <- f(out)
        out
      }
    ),
    class = "composite_function"
  )
}

decompose <- function(x) {
  if (!inherits(x, "composite_function"))
    abort("Not a composite function")
    
  environment(x)$fs
}

print.composite_function <- function(x, ...) {
  cat("<composite_function>\n")
  cat("(Listed in calling order)\n")
  
  for (f in rev(decompose(x))) {
    cat("\n")
    print(f)
  }
  
  invisible(x)
}
```
(The method for formals setting is the same as in #349.)

Example:
```R
foo <- compose(log, ~abs(.) + 1)

foo
#> <composite_function>
#> (Listed in calling order)
#> 
#> function (..., .x = ..1, .y = ..2, . = ..1) 
#> abs(.) + 1
#> 
#> function (x, base = exp(1))  .Primitive("log")

args(foo)
#> function (..., .x = ..1, .y = ..2, . = ..1) 
#> NULL

decompose(foo)
#> [[1]]
#> function (x, base = exp(1))  .Primitive("log")
#> 
#> [[2]]
#> function (..., .x = ..1, .y = ..2, . = ..1) 
#> abs(.x) + 1
```

lionel-:
Closes #366


R/compose.R
```diff
@@ -42,11 +42,37 @@ compose <- function(..., .rev = TRUE) {
...
+
+  structure(
+    new_function(formals(first_fn), body, fn_env(first_fn)),
+    class = "purrr_composed_fn",
+    first_fn = first_fn,
+    fns = fns
+  )
+}
+
+print.purrr_composed_fn <- function(x, ...) {
```
hadley: Missing `@export`?

R/compose.R
```diff
@@ -42,11 +42,37 @@ compose <- function(..., .rev = TRUE) {
...
+  cat("<composed>\n")
+
+  first <- attr(x, "first_fn")
+  cat("\n* Function 1:\n\n")
+  print(first, ...)
+
+  fns <- attr(x, "fns")
```
hadley: This makes me think we should have a test for compiling compose with zero and one function.

lionel-: `compose()` with zero inputs now returns the identity such that `compose(fn, compose())` === `compose(fn)`

PR #598, opened 2018-12-07
Add error conditions

lionel-:
Common parameters of `stop_` functions:

* `x` as first argument, the actual faulty value. This value is used to obtain the actual length or ptype if need.

* An optional `what` argument that is used to introduce the faulty value. It should be capitalised as it starts the error message. For instance supplying `"Object"` (the default) creates an error message starting with `"Object must be"`.

* An optional `arg` argument to specify from which argument name the faulty value comes. It is used to construct the default `what` value, if not supplied. E.g. `` "`.foo` must be..." ``.

Specific parameters:

* `stop_bad_type()` takes expected and actual type as friendly type descriptors. This gives it the necessary flexibility to be a base error type for the other errors.

  ```r
  stop_bad_type(list(), "a quux", actual = "a foobaz", arg = ".foo")
  #> Error: `.foo` must be a quux, not a foobaz
  ```

* `stop_bad_length()` takes `expected_length` and `.recycle` arguments. The parameters are prefixed with `.` when they are not included in the error object. Both these parameters are used to construct appropriate error messages:

  ```r
  stop_bad_length(1:3, 10)
  #> Error: Vector must have length 10, not 3

  stop_bad_length(1:3, 10, .recycle = TRUE, arg = ".foo")
  #> Error: `.foo` must have length 1 or 10, not 3
  ```

* `stop_bad_vector()` takes `expected_ptype`, and `expected_length`. The ptype is used with `friendly_type_of()` and the new `friendly_element_type_of()` function to construct appropriate error messages:

  ```r
  stop_bad_vector(1:3, character(), 10)
  #> Error: Vector must be a character vector of length 10, not an integer vector of length 3

  stop_bad_vector(1, character(), 1)
  #> Error: Vector must be a single string, not a single double

  stop_bad_vector(factor(c("a", "b")), character(), 1)
  #> Error: Vector must be a single string, not a vector of class `factor` and of length 2
  ```

* All these throwers have "element" variants which take an `index` argument. For these variants, `what` describes the container:

  ```r
  stop_bad_element_type(1:3, 1, "a foobaz")
  #> Error: Element 1 must be a foobaz, not an integer vector

  stop_bad_element_length(1:3, 2, expected_length = 1, what = "Index")
  #> Error: Index 2 must have length 1, not 3

  stop_bad_element_vector(1:3, 3, expected_ptype = character(), expected_length = 10, arg = ".foo")
  #> Error: Element 3 of `.foo` must be a character vector of length 10, not an integer vector of length 3
  ```

### DIFF
--- a/R/conditions.R
+++ b/R/conditions.R
@@ -23,7 +23,7 @@
 #' packages might tweak the friendly representation at any time.
 #'
 #' @keywords internal
-#' @name purrr-conditions
+#' @name purrr-conditions-type
 #' @noRd
 NULL
 
@@ -99,4 +99,79 @@ what_bad_element <- function(what, arg, index) {
 
   what <- what %||% "Element"
   sprintf("%s %d%s", what, index, where)
+}
+
+
+#' Error conditions for bad lengths
+#'
+#' @inheritParams purrr-conditions-type
+#' @param actual,expected The expected and actual lengths. If `actual`
+#'   is not supplied, `length(x)` is taken as default.
+#' @param recycle Whether `x` is also allowed to have length 1.
+#'
+#' @name purrr-conditions-length
+NULL
+
+stop_bad_length <- function(x,
+                            expected,
+                            ...,
+                            actual = NULL,
+                            what = NULL,
+                            arg = NULL,
+                            message = NULL,
+                            .recycle = FALSE,
+                            .subclass = NULL) {
+  what <- what %||% what_bad_object(arg) %||% "Vector"
+  actual <- actual %||% length(x)
+
+  if (.recycle) {
+    expected <- sprintf("1 or %s", expected)
+  }
+
+  message <- message %||% sprintf("%s must have length %s, not %s", what, expected, actual)
+
+  if (!is_integerish(actual)) {
+    stop_bad_type(actual, "a single number", arg = "`actual`")
+  }
+  if (length(actual) != 1) {
+    stop_bad_length(actual, 1, arg = "`actual`")
+  }
+
+  abort(
+    message,
+    x = x,
+    expected = expected,
+    actual = actual,
+    what = what,
+    arg = arg,
+    ...,
+    .subclass = c(.subclass, "purrr_error_bad_length")
+  )
+}
+
+stop_bad_element_length <- function(x,
+                                    index,
+                                    expected,
+                                    ...,
+                                    actual = NULL,
+                                    what = NULL,
+                                    arg = NULL,
+                                    message = NULL,
+                                    .recycle = FALSE,
+                                    .subclass = NULL) {
+  stopifnot(is_integerish(index, n = 1, finite = TRUE))
+  what <- what_bad_element(what, arg, index)
+
+  stop_bad_length(
+    x,
+    expected,
+    actual = actual,
+    what = what,
+    arg = arg,
+    index = index,
+    ...,
+    message = message,
+    .recycle = .recycle,
+    .subclass = c(.subclass, "purrr_error_bad_element_length")
+  )
 }
--- a/R/conditions.R
+++ b/R/conditions.R
@@ -105,43 +105,43 @@ what_bad_element <- function(what, arg, index) {
 #' Error conditions for bad lengths
 #'
 #' @inheritParams purrr-conditions-type
-#' @param actual,expected The expected and actual lengths. If `actual`
-#'   is not supplied, `length(x)` is taken as default.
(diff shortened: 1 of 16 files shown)


R/conditions.R
```diff
@@ -0,0 +1,254 @@
...
+  what <- what %||% what_bad_object(arg) %||% "Vector"
+
+  if (.recycle) {
+    expected <- sprintf("1 or %s", expected_length)
+  } else {
+    expected <- as.character(expected_length)
+  }
+  actual <- length(x)
+
+  message <- message %||% sprintf(
```
hadley: You put this all on one line above

R/conditions.R
```diff
@@ -0,0 +1,254 @@
...
+
+stop_bad_element_vector <- function(x,
+                                    index,
+                                    expected_ptype,
+                                    expected_length,
+                                    ...,
+                                    what = NULL,
+                                    arg = NULL,
+                                    message = NULL,
+                                    .recycle = FALSE,
```
hadley: I don't understand why `recycle` gets a `.` prefix

R/utils.R
```diff
@@ -185,3 +185,26 @@ paste_classes <- function(x) {
...
+  }
+
+  switch(typeof(x),
+    logical   = if (single) "a single logical"        else "a logical element",
+    integer   = if (single) "a single integer"        else "an integer element",
+    double    = if (single) "a single double"         else "a double element",
+    complex   = if (single) "a single complex number" else "a complex number",
+    character = if (single) "a single string"         else "a character vector element",
+    raw       = if (single) "a single raw value"      else "a raw vector element",
+    list      = if (single) "a single list element"   else "a list element",
```
hadley: Is element correct here?

tests/testthat/test-conditions.R
```diff
@@ -0,0 +1,90 @@
...
+  err <- catch_cnd(stop_bad_type(NA, "`NULL`", actual = "a foobaz", arg = ".foo"))
+  expect_is(err, "purrr_error_bad_type")
+  expect_identical(err$x, NA)
+  expect_identical(err$expected, "`NULL`")
+  expect_identical(err$actual, "a foobaz")
+  expect_identical(err$arg, ".foo")
+})
+
+test_that("stop_bad_type() constructs default `what`", {
+  expect_error(
```
hadley: I think ideally these would be regression tests, but testthat can't support yet.

hadley: I'm also not sure about testing text and class together, in light of https://github.com/r-lib/testthat/issues/816

R/conditions.R
lionel-: Because it doesn't become a field of the error condition.

R/utils.R
lionel-: What do you mean? Would you prefer something like "an R object"?

tests/testthat/test-conditions.R
lionel-: I'm now testing with this helper:

```r
expect_error_cnd <- function(expr, regexp, class, ...) {
  err <- catch_cnd(expr)
  expect_true(inherits_all(err, c(class, "error", "condition")))
  expect_match(conditionMessage(err), regexp, ...)
}
```

R/utils.R
hadley: I can't quite imagine what this would look like in an actual call. Can you give an example?

tests/testthat/test-conditions.R
hadley: 👍

R/conditions.R
hadley: Oh hmmmmmmmmm. Why doesn't it?

R/utils.R
lionel-: I have simplified `friendly_type_element_of()` to always return a description for a single value:

```r
  switch(typeof(x),
    logical   = "a single logical",
    integer   = "a single integer",
    double    = "a single double",
    complex   = "a single complex number",
    character = "a single string",
    raw       = "a single raw value",
    list      = "a list of one element",
    abort("Expected a base vector type")
  )
```

Example of error messages:

```r
stop_bad_vector(1:3, character(), 1)
#> Error: Vector must be a single string, not an integer vector of length 3

stop_bad_vector(1:3, list(), 1)
#> Error: Vector must be a list of one element, not an integer vector of length 3
```

R/conditions.R
lionel-: It now does and the parameter is no longer prefixed with a dot.

Issue #599, opened 2018-12-11
accumulate() simplifies output


lionel-:
Because it was implemented with `base::Reduce()`. Should probably return a list by default and accept an optional `.ptype` argument. Or should the init value determine the result type?

lionel-:
Too late to fix this.

PR #600, opened 2018-12-11
Early termination with done sentinels

### LINKED ISSUE #253: accumulate_while

lionel-:
This feature is inspired by https://clojuredocs.org/clojure.core/reduced

* Returning a value wrapped in a `done_box()` terminates a reduction early. `done_box()` is implemented here for now but probably should move to rlang as a general purpose sentinel for early termination.

* If the done box is empty, the last value is used instead. This is not useful for `reduce()` since it is equivalent to returning the result-so-far in a done box. However that's useful for `accumulate()`, the result length is 1 element shorter when returning an empty done box.

* `accumulate()` is rewritten in terms of `reduce_impl()`. I found out that `base::Reduce()` performs a simplification step, which is reproduced for compatibility. On the other hand the new `accumulate2()` function does not simplify, as it seems the right default. We should sort this out for the vctrs milestone, I have opened an issue at https://github.com/tidyverse/purrr/issues/599.

Closes #253.
Closes #566.


lionel-:
Good idea, `done()`.

PR #604, opened 2018-12-12
Add rate helpers and slowly() adverb

### LINKED ISSUE #389: Add slowly() adverb
I recently used `safely()` to wrap `gmailr::send_message()`, which I then used with `pmap()` to iterate over rows = recipients. I ran into sporadic rate limit problems.

It would be nice to also have a `slowly()` adverb for adding a little bit of delay to each call.

lionel-:
* `rate_backoff()` and `rate_delay()` create rate objects to determine waiting time between attempts. They wrap a stateful counter and are parameterised with a max number of attempts.

* `insistently()` now takes a rate object, by default a backoff rate.

* New `slowly()` adverb to wait between each call, by default with a constant rate.

* `rate_sleep()` powers `insistently()` and `slowly()`. It takes a rate, computes a waiting time, and waits. It also throws errors if the max number of attempts is reached.

  The first time that `rate_sleep()` is called, it returns immediately after signalling an init condition. If `rate_sleep()` is called again after it has thrown a max attempt error, it then throws an expiration error suggesting to use `rate_reset()` to reset the rate. An insistent function resets the rate each time it is called.

Closes #389.


hadley:
I know it's a little inconsistent, but it just feels right to me here, especially since this is unlikely to be a future extension point, and you're unlikely to use it that often (so the prompts from autocomplete are particularly helpful)

PR #619, opened 2019-01-19
Issue608

### LINKED ISSUE #608: Use tidyselect in map_at
It would be nice if `map_at()` could incorporate tidyselect functions: 

``` r 
library(purrr)
library(tidyselect)

map_at_two <- function(.x, .at, .f, ...){
  outvars <- vars_select(.vars = names(.x), .at)
  map_at(.x, outvars, .f, ...)
}

x <- list(a = "b", b = "c", aa = "bb")

map_at_two(x, contains("a"), toupper)
#> $a
#> [1] "B"
#> 
#> $b
#> [1] "c"
#> 
#> $aa
#> [1] "BB"

map_at_two(x, 1, toupper)
#> $a
#> [1] "B"
#> 
#> $b
#> [1] "c"
#> 
#> $aa
#> [1] "bb"

map_at_two(x, "a", toupper)
#> $a
#> [1] "B"
#> 
#> $b
#> [1] "c"
#> 
#> $aa
#> [1] "bb"

x <- list("b", "c", "bb")

map_at_two(x, contains("a"), toupper)
#> Error: No tidyselect variables were registered

<sup>Created on 2019-01-02 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>
```

`{tidyselect}` has purrr as a dependencies, so it might imply that `{purrr}` functions used in `{tidyselect}` have to be rewritten.

ColinFay:
This PR closes #608 about including the `tidyselect` semantic into  `purrr` `*_at` functions

The functions for doing selection in purrr are:

+ `lmap_at()`
+ `map_at()`
+ `modify_at()` & methods 

And they can now be used with `map_at(my_list, vars(contains("a")), toupper)` so to be consistent with the behavior of `dplyr::mutate_at()`.

This implementation first check if the `.at` is a `quosure`, if so checks if `tidyselect` is installed, and if `TRUE` runs `tidyselect::vars_select`, otherwise throws an error if the package is not installed (inspired by the way `map_df()` works).


R/map.R
```diff
@@ -151,10 +153,13 @@ map_if <- function(.x, .p, .f, ..., .else = NULL) {
 
   set_names(out, names(.x))
 }
+
 #' @rdname map
 #' @export
 map_at <- function(.x, .at, .f, ...) {
-  sel <- inv_which(.x, .at)
+
+  where <- at_selection( names(.x), .at)
```
lionel-: Could you remove all the leading spaces before `names` please.

R/utils.R
```diff
@@ -23,6 +23,40 @@ check_tibble <- function() {
...
 
+check_tidyselect <- function(){
+  if (!is_installed("tidyselect")) {
+    abort("Using tidyselect in `map_at()` requires tidyselect")
+  }
+}
+
+#' Select names
+#'
+#' This helper function is intended to provide a semantic equivalent to the one from `dplyr::mutate_at()`. It can be used in the `*_at` functions of `purrr` like `map_at()`. Note that the `tidyselect` package has to be installed.
```
lionel-: Could you hard-reflow the paragraph at 80 characters please. Same for the `@param` section.

R/map.R
```diff
@@ -37,6 +37,8 @@
 #' @param .at A character vector of names, positive numeric vector of
 #'   positions to include, or a negative numeric vector of positions to
 #'   exlude. Only those elements corresponding to `.at` will be modified.
+#'   If the `tidyselect` package is installed, you can use `vars()` &
```
lionel-: & -> and

R/utils.R
```diff
@@ -23,6 +23,40 @@ check_tibble <- function() {
...
+#'}
+
+vars <- function (...) {
+  quos(...)
+}
+
+at_selection <- function(nm, .at){
+  if (is_quosures(.at)){
+    check_tidyselect()
+    .at <- tidyselect::vars_select(.vars = nm, !!!.at)
```
lionel-: The `else` statement should be removed, or you can keep it but in this case should return the variables directly from the `if`branch without assigning to `.at`.

R/utils.R
```diff
@@ -23,6 +23,40 @@ check_tibble <- function() {
...
+#'
+#' @export
+#'
+#' @examples
+#' \dontrun{
+#' library(tidyselect)
+#' x <- list(a = "b", b = "c", aa = "bb")
+#' map_at(x, vars(contains("a")), toupper)
+#'}
+
```
lionel-: The function should follow the roxygen block immediately, without a new line.

ColinFay:
@lionel- I made the changes and update the NEWS 👍

lionel-:
Awesome, thanks!

PR #620, opened 2019-01-19
Improve documentation + examples for map_dfr(), map_dfc() and variants. resolves #533

### LINKED ISSUE #533: map2-pmap: need better explanation of _dfc and _dfr functionality
I'm guessing it's an indicator of operating on rows or columns.

It's in the map2-pmap.R code...maybe put an explanation in the "Values" section?  Also, might be helpful to include an example showing what it does.
https://github.com/tidyverse/purrr/blob/master/R/map2-pmap.R

Ryo-N7:


R/map2-pmap.R
```diff
@@ -84,6 +86,18 @@
 #' pmin(df$x, df$y)
 #' map2_dbl(df$x, df$y, min)
 #' pmap_dbl(df, min)
+#'
+#' # If you want to bind the results of your function rowwise, use map2_dfr() or pmap_dfr()
+#' ex_fun <- function(arg1, arg2){
+#' col <- arg1 + arg2
+#' x <- dplyr::as_data_frame(col)
```
lionel-: `as_data_frame()` is deprecated, you would need to use `as_tibble()`.

But the examples should work without having suggested packages installed, so you'd also need to wrap the example within `if (rlang::is_installed)`. For this reason, I think it'd be simpler to use base R to create the data frame.

lionel-:
Thanks!

PR #625, opened 2019-01-24
adds insistently to the "adverbs" reference for pkgdown site

ijlyttle:
all this does is add an entry for `insistently` to the adverbs section of the function-reference


lionel-:
Thanks!

(By the way, we use Sentence case with the imperative mood in our commit messages)

PR #627, opened 2019-01-28
Fix missing topics on the reference index

yutannihilation:
It seems some topics are missing: https://purrr.tidyverse.org/reference/index.html

Actually, I see these warnings when running `pkgdown::build_reference()`. Some functions are reexported from rlang, so we can ignore. But, others are not ignorable.

``` r
Warning: In '_pkgdown.yml', topic must match a function or concept.
Problem topic: `starts_with("invoke")`
Warning: Topics missing from index: attr_getter, done, exec, pluck, rate-helpers, rate_sleep, rep_along, zap
```

This PR fixes:

- Remove `invoke` as it is now marked as `internal` due to deprecation
- Add missing topics
- Move `modify_in()` from "Map Variants" section

I'm not sure "Plucking" is the right title for the section. Any suggestions...?

### DIFF
--- a/_pkgdown.yml
+++ b/_pkgdown.yml
@@ -46,14 +46,12 @@ reference:
     `modify()` modifies "in place", returning a vector the same type as `.x`;
     `map2()` iterates over two vectors in parallel;
     `pmap()` (parallel map) iterates over a list of vectors;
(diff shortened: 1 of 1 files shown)


lionel-:
Awesome, thanks!

PR #635, opened 2019-02-06
Do not allow modify2() to recycle the first argument

### LINKED ISSUE #632: modify2() doesn't preserve 1-column data.frame
``` r
library(purrr)

# OK
modify(data.frame(x = 1), ~ identity(.x))
#>   x
#> 1 1

# wrong
imodify(data.frame(x = 1), ~ identity(.x))
#> [[1]]
#> [1] 1

# wrong
modify2(data.frame(x = 1), "x", ~ identity(.x))
#> [[1]]
#> [1] 1

# OK
imodify(data.frame(x = 1, y = 1), ~ identity(.x))
#>   x y
#> 1 1 1
```

<sup>Created on 2019-02-02 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

yutannihilation:
Fix #632 

Currently, `modify2()` recycles the first argument when it is 1-length object, so the result is no more have the original class and attributes. This PR makes `recycle_args()` more strict to prevent such unexpected results.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,9 @@
 
 # purrr 0.3.0.9000
 
+* `modify2()` and `imodify()` now don't recycle the first argument, so the
+  result always has the same type as the input object (@yutannihilation,
+  #632).
 
 # purrr 0.3.0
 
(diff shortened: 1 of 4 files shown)


hadley:
I'm going to close this because in the >3 years (!!) since you wrote this, we now have `vctrs::vec_recycle()` which we should use here.

PR #641, opened 2019-02-15
Skip dplyr-related tests if unavailable.

QuLogic:
Tests are currently skipped that need `tidyselect`, but `dplyr` is also optional and not yet skipped.

Since `dplyr` now depends on `tidyselect`, this makes it difficult to bootstrap packages and run tests (at least for non-optional things).


lionel-:
Thanks!

PR #654, opened 2019-03-03
Fix `prepend()` with empty list

### LINKED ISSUE #637: prepend: fails for empty list
`purrr::prepend` fails for empty list due to the checking of the `before` param though (partly based on the documentation) I would expect it to work as `base::append` by default. Is there a reason for this?

If you agree that it should work for empty lists as well I am happy to submit a PR.

``` r
purrr::prepend(list(), "a")
#> Error in purrr::prepend(list(), "a"): before > 0 && before <= n is not TRUE

append("a", list())
#> [[1]]
#> [1] "a"

purrr::prepend(list("b"), "a")
#> [[1]]
#> [1] "a"
#> 
#> [[2]]
#> [1] "b"
append("a", list("b"))
#> [[1]]
#> [1] "a"
#> 
#> [[2]]
#> [1] "b"
```

Created on 2019-02-10 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).

My first idea for implementation:
```
function (x, values, before = 1) 
{
    n <- length(x)
    stopifnot(n == 0 || (before > 0 && before <= n))
    if (before == 1) {
        c(values, x)
    }
    else {
        c(x[1:(before - 1)], values, x[before:n])
    }
}
```

czeildi:
- changes default value for `before` argument from 1 to `NULL`
- behaviour is identical for non-empty list
- by default now appends at the beginning for empty list instead throwing an error
- if before is not NULL, still throws error for empty list
- error message now includes NULL check for all cases

Fixes #637

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -11,6 +11,8 @@
 * `compose()` now works with generic functions again (#629, #639). Its
   set of unit tests was expanded to cover many edge cases.
 
+* `prepend()` now works with empty lists (@czeildi, #637)
+
 
 # purrr 0.3.0
 
--- a/R/prepend.R
+++ b/R/prepend.R
@@ -7,7 +7,8 @@
 #'
 #' @param x the vector to be modified.
 #' @param values to be included in the modified vector.
-#' @param before a subscript, before which the values are to be appended.
+#' @param before a subscript, before which the values are to be appended. If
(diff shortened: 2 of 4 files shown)


czeildi:
I redid this PR https://github.com/tidyverse/purrr/pull/687 to follow best practices described in https://happygitwithr.com/fork-and-clone.html, now PR from a non-master branch

PR #676, opened 2019-05-30
Replace deprecated .right with .dir in detect() documentation. Removes deprecation warning from documentation and closes #675.

### LINKED ISSUE #675: Documentation for detect() uses deprecated .right instead of preferred .dir
The official documentation page for detect() shows a deprecation warning: https://purrr.tidyverse.org/reference/detect.html

This should be fixed by updating the example code in two places in _detect.R_

Breza:


lionel-:
Thanks!

PR #682, opened 2019-06-14
Modified typo

chsafouane:
"exclude" instead of "exlude"


lionel-:
Can you also change the `.R` file please?

lionel-:
Thanks for spotting this!

For future reference, you need to change the R file instead of the Rd files. The latter contain the rendered documentation and are automatically overwritten by `devtools::document()` based on the R files.

Issue #679, opened 2019-06-11
`map2` throws empy list when `.y` argument is NULL


joelgombin:
See this reprex:

```r
library(purrr)
alist <- list("bla") 
alist %>% 
  map_chr(~ .)
#> [1] "bla"
alist %>% 
  map2(.y = NULL, ~ .x)
#> list()
alist %>% 
  map2(.y = alist, ~ .x)
#> [[1]]
#> [1] "bla"
```

This may be related to #199 ?

lionel-:
I think we'd now make it an error rather than recycling to length zero. We now generally treat `NULL` as an absence of argument rather than as an empty vector. cc @hadley

hadley:
Agreed

hadley:
Hmmm, I could argue this either way for `map2()`. In any case, I think we should just apply the standard rules (i.e. #878) and see what falls out.

Issue #633, opened 2019-02-04
map_chr implicit coercion


grayskripko:
Why it isn't so strict to throw an error?
```{r}
1:2 %>% map_chr(length)
# > [1] "1" "1"
```

lionel-:
This will change when we switch to vctrs for our coercion backend.

grayskripko:
ok. Do you need this issue open?

lionel-:
Yes we'll tackle it at the vctrs milestone.

hadley:
Now tracking in #904

Issue #472, opened 2018-03-09
map_dfr() column binds vectors


garrettgman:
`map_dfr()` does the same thing as `map_dfc()` when given a list of vectors: it _column binds_ the vectors into a data frame.

````r
list_of_vecs <- list(a = c(1,1,1, 1), b = c(2, 2, 2, 2), c = c(3, 3, 3, 3))
list_of_vecs %>% map_dfr(~.x)
## # A tibble: 4 x 3
##       a     b     c
##   <dbl> <dbl> <dbl>
## 1     1     2     3
## 2     1     2     3
## 3     1     2     3
## 4     1     2     3
````

I'd expect `map_dfr()` to row bind the vectors into a data frame with one long column, or to throw an error.

jennybc:
Past thread re: difficulty of row binding: https://github.com/tidyverse/purrr/issues/179

lionel-:
> I'd expect map_dfr() to row bind the vectors into a data frame with one long column, or to throw an error.

Perhaps taking each vector as a row would be more natural? They'd need consistent internal names.

Right now they are taken as columns because `bind_rows()` and `bind_cols()` have somewhat sloppy semantics and interpret lists as data frames. This will need to wait on the new vctrs-based tools.

jennybc:
tibble is also in a holding pattern on a related matter. I might add a discussion of this to my existing slot in Monday's group meeting.

CottonRockwood:
My expectation would be that map_dfr would r_bind the results as rows if they are vectors while map_dfc would combine the the vectors as columns.

hadley:
Should switch to `vec_rbind()` and `vec_cbind()` and see what breaks.

leungi:
Faced similar issues when trying to get a tidy tibble from a vector of file paths off `fs::dir_ls()`.

Ended up with `purrr::map()` + `tibble::enframe()` + `tidyr::unnest()` combo.

```r
> library(dplyr)
> 
> file_ls <- fs::dir_ls("./raw_data/Guidelines/",
 regexp = "*.md"
 )
> 
> file_ls %>%
 .[1:2] %>%
 purrr::map_dfr(readr::read_file) %>%
 bind_rows(.id = "doc_id")
# A tibble: 1 x 3
  doc_id `./raw_data/Guidelines/ `./raw_data/Guidelines/
  <chr>  <chr>                                   <chr>                                 
  1      **Coiled~                **Operation~
> 
> file_ls %>%
[...]

Issue #376, opened 2017-08-23
map_dfc() fails when the result contains list columns


yutannihilation:
Sometimes we have a data.frame-like list and want to apply some function and harvest the result as data.frame. `map_dfc()` is quite useful for this purpose:

``` r
not_nested <- list(
  col1 = c("Apple", "Orange"),
  col2 = c("Baseball", "Football")
)

purrr::map_dfc(not_nested, purrr::map_chr, sprintf, fmt = "I like %s")
#> # A tibble: 2 x 2
#>            col1            col2
#>           <chr>           <chr>
#> 1  I like Apple I like Baseball
#> 2 I like Orange I like Football
```

But it fails if the result contains list columns:

```r
nested <- list(
  col1 = list(
    c("Apple", "Banana"),
    c("Orange")
  ),
  col2 = list(
    c("Baseball", "Soccer"),
    c("Football")
  )
)

purrr::map_dfc(nested, purrr::map, sprintf, fmt = "I like %s")
#> Error in cbind_all(x): Argument 2 must be length 2, not 1
```

Is it possible to get the result as data.frame like bellow?

``` r
tibble::as_tibble(
  purrr::map(nested, purrr::map, sprintf, fmt = "I like %s")
)
#> # A tibble: 2 x 2
#>        col1      col2
#>      <list>    <list>
#> 1 <chr [2]> <chr [2]>
#> 2 <chr [1]> <chr [1]>
```

(As the error indicates, this seems up to `cbind_all()`. Should I file this issue to dplyr's repo?)

lionel-:
The problem is that `bind_cols()` and `bind_rows()` implicitly splice inputs:

```r
dplyr::bind_cols(y = list(3, 4))
#> # A tibble: 1 x 2
#>      V1    V2
#>   <dbl> <dbl>
#> 1     3     4

dplyr::bind_cols(list(y = 3:4, z = 4:5))
#> # A tibble: 2 x 2
#>       y     z
#>   <int> <int>
#> 1     3     4
#> 2     4     5
```

So list-columns are spliced as well unless we wrap all inputs:

```r
dplyr::bind_cols(x = 1:2, y = list(3, 4))
#> Error: Argument 2 must be length 2, not 1

dplyr::bind_cols(list(x = 1:2, y = list(3, 4)))
#> # A tibble: 2 x 2
#>       x y
#>   <int> <list>
#> 1     1 <dbl [1]>
#> 2     2 <dbl [1]>
```

[...]

PR #683, opened 2019-06-18
Implement `map_vec()`

DavisVaughan:
This PR implements `map_vec()`, a vctrs powered `map()` that attempts to simplify the output automatically by computing the common type of the result elements. If a common type can be found, the result will be simplified to that type. If not, a list is returned.

This is a draft PR for now as I'm not currently proposing that it be merged. I'd like to get feedback and see what others think about it.

- [ ] Add tests
- [ ] Add examples

This function maintains the invariant that:

```
vec_size(.x) == vec_size(map_vec(.x))
```

After glancing at the issues, it seems this PR touches: #679, #633, #472, #376

For now, here are a number of examples that demonstrate the features of `map_vec()`:

<details>

``` r
library(vctrs)
library(purrr)
library(tibble)
options(rlang_backtrace_on_error = "reminder")

# Auto simplified to an integer vector
map_vec(1:2, ~ .x)
#> [1] 1 2

# Prevent simplification with `.ptype = list()`
map_vec(1:2, ~ .x, .ptype = list())
#> [[1]]
#> [1] 1
#> 
#> [[2]]
#> [1] 2

# If a `.ptype` is specified, and casting to that type
# is not possible, an error is raised
map_vec(1:2, ~ .x, .ptype = factor())
#> Can't cast <integer> to <factor<>>

# If simplification is possible, all elements must have size 1,
# otherwise an error is raised
map_vec(1:2, ~ if (.x == 1L) 1:2 else 3)
#> Result 1 must be a single double, not an integer vector of length 2

# If you use `.ptype = list()`, this is pseudo-relaxed
# (The size of the output is still the same as the 
# size of the input (2), but each inner element can vary in size)
map_vec(1:2, ~ if (.x == 1L) 1:2 else 3, .ptype = list())
#> [[1]]
#> [1] 1 2
#> 
#> [[2]]
#> [1] 3

# The best thing about `map_vec()` is its flexibility with other
# non-atomic types, for example, simplifying to a date vector
map_vec(1:2, ~ Sys.Date() + .x)
#> [1] "2019-06-19" "2019-06-20"

# If a common type cannot be determined, a list is returned
map_vec(list(1, "x"), ~ .x)
#> [[1]]
#> [1] 1
#> 
#> [[2]]
#> [1] "x"

# Note that just because a common type isn't found doesn't mean you
# can't still forcibly cast to a certain type.
map_vec(list(1, "x"), ~ .x, .ptype = character())
#> [1] "1" "x"

# The underlying `vec_c()` engine knows how to combine 
# data frame output, but this isn't a `map_dfr()` replacement
map_vec(1:2, ~ tibble(x = .x))
#> # A tibble: 2 x 1
#>       x
#>   <int>
#> 1     1
#> 2     2

# You can enforce the structure of the data frame output with a ptype.
# This has the same result as before but coerces the integers to characters
ptype <- tibble(x = character())
map_vec(1:2, ~ tibble(x = .x), .ptype = ptype)
#> # A tibble: 2 x 1
#>   x    
#>   <chr>
#> 1 1    
#> 2 2

# And this errors
map_vec(1:2, ~ tibble(x = .x, y = 1), .ptype = ptype)
#> Lossy cast from <tbl_df<
#>   x: integer
#>   y: double
#> >> to <tbl_df<x:character>>.
#> Dropped variables: `y`

# Or you can enforce a partial structure with a partial_frame()
# This ensures that a `y` column exists
partial_ptype <- partial_frame(y = numeric())
map_vec(1:2, ~ tibble(x = .x), .ptype = partial_ptype)
#> # A tibble: 2 x 2
#>       x     y
#>   <int> <dbl>
#> 1     1    NA
#> 2     2    NA

# Turning on strict mode forces the user to provide a .ptype
withr::with_options(
  list(vctrs.no_guessing = TRUE),
  map_vec(1:5, ~ .x)
)
#> Error: strict mode is activated; you must supply complete `.ptype`.
```

</details>


DavisVaughan:
A few questions:
- Should a `.name_repair` argument be passed to `vec_c()`?
- If a `.ptype` is specified and `vec_c()` fails because it cannot cast to the `.ptype`, we currently get a vctrs cast error. Should that be more informative? Can it be (we don't have the index where the cast failure occured)?

DavisVaughan:
Other things vctrs can help with:
- Using `vec_rbind()` in `map_dfr()`
- Using `vec_cbind()` in `map_dfc()`
- Possibly powering `map_dbl()` and friends by just doing `map_vec(.ptype = double())`?
- Alterting the current recycling rules in purrr. Current thinking suggests that this should produce an error `purrr::map2(1:2, integer(), ~ .x)`

lionel-:
Last time we talked about this with Hadley, the idea was to have `funs::s()` to simplify an output. Then our sapply would be `s(map(...))`. This approach seems better than multiplying the number of functions in purrr / other packages as it is more composable.

Similarly, `map_int(...)` could be `int(map(...))`. However the typed variants are more efficient because they can omit the intermediate representation as a list. In general, performance is a reason to provide variants.

Does this suggest we should have `map_vec()` in addition to `s(map())` if it can help performance? The prototype should be computed in advance and then we'd build the vector and vec_slice into it. This would require exporting some of the C API from vctrs.

DavisVaughan:
I see that mention of `funs::s()` [here](https://github.com/r-lib/funs/issues/23#issue-408193412). I guess it would be like `s(x, ptype)`?

I am a bit worried about performance here. If purrr was being implemented from scratch today, I feel like a C version of `map_vec()` would still exist just for its performance gains. I think even `map_int()` and some of the more common variants might even still exist because it is so much easier to type than `map_vec(x, fun, .ptype = int())` when you want to guarantee that the result is an integer + have the performance.

I do think generally `funs::s()` will be a nice solution to not multiply the number of these kinds of variants in other packages, but purrr feels like an exception.

hadley:
Superseded by #894

PR #685, opened 2019-07-02
Fixed typo in pluck documentation.

CGMossa:
Fixed a typo in the documentation for `pluck`. There was an extra `x`.


lionel-:
This is meant for `assign_in()` which inherits from this documentation topic.

PR #687, opened 2019-07-07
Fix `prepend()` with empty list

### LINKED ISSUE #637: prepend: fails for empty list

czeildi:
- changes default value for `before` argument from 1 to `NULL`
- behaviour is identical for non-empty list
- by default now appends at the beginning for empty list instead throwing an error
- if before is not NULL, still throws error for empty list
- error message now includes NULL check for all cases

Fixes #637

(I redid a previous pull request with the same content where I did not follow best practices described in https://happygitwithr.com/fork-and-clone.html . Now I am creating a PR from a non-master branch.)

### DIFF
--- a/tests/testthat/test-prepend.R
+++ b/tests/testthat/test-prepend.R
@@ -13,3 +13,27 @@ test_that("prepend is clearer version of merging with c()", {
     }
   )
 })
+
+test_that("prepend appends at the beginning for empty list by default", {
+  x <- list()
+  expect_identical(
+    x %>% prepend(1),
+    x %>% c(1, .)
+  )
+})
+
+test_that("prepend throws error if before param is neither NULL nor between 1 and length(x)", {
+  expect_error(
+    prepend(list(), 1, before = 1),
+    "is.null(before) || (before > 0 && before <= n) is not TRUE"
+  )
+  x <- as.list(1:3)
+  expect_error(
+    x %>% prepend(4, before = 0),
+    "is.null(before) || (before > 0 && before <= n) is not TRUE"
+  )
+  expect_error(
+    x %>% prepend(4, before = 4),
+    "is.null(before) || (before > 0 && before <= n) is not TRUE"
+  )
+})
--- a/NEWS.md
+++ b/NEWS.md
@@ -21,6 +21,8 @@
 * `compose()` now works with generic functions again (#629, #639). Its
   set of unit tests was expanded to cover many edge cases.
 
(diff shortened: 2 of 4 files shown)


tests/testthat/test-prepend.R
```diff
@@ -13,3 +13,27 @@ test_that("prepend is clearer version of merging with c()", {
     }
   )
 })
+
+test_that("prepend appends at the beginning for empty list by default", {
+  x <- list()
```
lionel-: Nitpick: I would write this as `expect_identical(prepend(list(), 1), 1)` for clarity.

tests/testthat/test-prepend.R
```diff
@@ -13,3 +13,27 @@ test_that("prepend is clearer version of merging with c()", {
...
+  expect_identical(
+    x %>% prepend(1),
+    x %>% c(1, .)
+  )
+})
+
+test_that("prepend throws error if before param is neither NULL nor between 1 and length(x)", {
+  expect_error(
+    prepend(list(), 1, before = 1),
+    "is.null(before) || (before > 0 && before <= n) is not TRUE"
```
lionel-: This message testing seems too stringent because the message is generated upstream (by `base::stopifnot()`) and so is not under the control of purrr. Upstream changes are unlikely but would cause these tests to fail.

lionel-:
Thanks!

Issue #696, opened 2019-08-09
partial: additional indefinite article in error message


shosaco:
There are additional indefinite articles in some error messages. This happens because `friendly_type_of` outputs the indefinite article *a* (e.g. *a character vector*) and `partial` also outputs *a friendly_type_of(.f)*, resulting in double *a* in its error message - e.g. *a a character vector*.

``` r
library(purrr)
test_path <- "C:"
partial(file.path(test_path))
#> `.f` must be a function, not a a character vector
```

shosaco:
fixed in Pull request  #697

lionel-:
Thanks for the fix.

PR #697, opened 2019-08-09
remove additional indefinite article in error message of `partial`

shosaco:
fixes issue #696 (additional indefinite article **a** in example below):

``` r
library(purrr)
test_path <- "C:"
partial(file.path(test_path))
#> `.f` must be a function, not a a character vector
```


lionel-:
Thanks!

PR #700, opened 2019-08-14
Reorder detect() documentation to replace mention of .p

Rekyt:
Hi,
I'ved change the order of arguments inheritance in `detect()` to better match the documentation of the `...` argument. So that the documentation does not mention the `.p` argument.
I've then ran `devtools::document()`.


lionel-:
Thanks!

PR #713, opened 2019-10-18
doc: Added example with named function for `imap()` functions.

jhrcook:
The documentation only had an example for using an anonymous
function declared within the call to the the `imap_***()`
functions. It was not immediately clear how to use `imap()` with
a named function. I just added a single line to the Description
and a small example to the documentation.

All tests passed only MacBook Pro running macOS Majove 10.14.4 and R version 3.6.1 (2019-07-05).


hadley:
Thanks for working on this PR, but I think this needs to documented somewhere "higher up" since you can use a named function anywhere you can use an anonymous function, not just in `imap()`.

PR #741, opened 2020-02-01
Adds guidelines and best practices for adverb-wrapped functions

### LINKED ISSUE #668: Guidelines/best practices on exporting adverb-wrapped functions from packages
`insistently()` and `safely()` have been lifesavers for my team, particularly in solving problems where low failure rates over large-scale API call queues and read/write operations would break long-running processes. Interacting with AWS S3 from R is much easier now.

Are there any recommendations/best practices for including adverb-wrapped functions in packages? Since the release of `insistently()` I’ve never called (eg) the `aws.s3::put_object()` function directly. Every script starts with creating wrappers for `insist_*`. I’d like to move the wrapped functions to an internal package, and maybe propose to include an insistent option in the main package. However, I haven’t seen any packages that export safe_* or insistent_* function variants to date, and that’s a good sign I'm planning to do something without full understanding of the side effects.

(First time filing an issue here -- thank you! `purrr` is the single largest productivity boost I've had in >10 years of writing R code...)

njtierney:
Closes #668


lionel-:
Thanks!

PR #742, opened 2020-02-01
Update pluck.Rd

sgschreiber:
Did not return NULL but NA.


lionel-:
Can you update the corresponding `.R` file please? The `.Rd` files are automatically generated, so your changes will be overwritten at the next regeneration.

sgschreiber:
Thanks @lionel- for the clarification regarding the `.Rd` files! Pull request for `pluck.R` submitted a few days ago.

lionel-:
I see, it's now merged, thanks!

PR #748, opened 2020-02-28
adds flip function

### LINKED ISSUE #728: FR: implement flip function
Occasionally it is useful to flip the arguments of a function. Particularly when pipping data, if the function you want to use expects the data argument to be the second argument the current best approaches as far as I can tell would be:

1. `data %>% partial(fn, x = "value")`
2. `data %>% fn("value", .)`

Haskell implements a function called [flip](http://zvon.org/other/haskell/Outputprelude/flip_f.html) t(this was previously mentioned by @lionel- in https://github.com/tidyverse/purrr/pull/42#issuecomment-138890361 but I can't see that there was any further discussion about this function). The implementation in R AFAIK shouldn't be too difficult. As a very minimal example of this function working:

``` r
flip <- function(fn) {
  function(.x, .y, ...) fn(.y, .x, ...)
}

(`>`)(5, 3)
#> [1] TRUE

flip(`>`)(5, 3)
#> [1] FALSE
```

A bit more of an (admittedly contrived) example of the use of this function is below. Here I am working with an sf table, which if I join some extra data to, but have the sf table as the right table, I lose all of the sf metadata. But, if I flip inner_join I am pipping my data to be the right input, and it stays as a sfc.

``` r
library(tidyverse)
library(sf)
#> Linking to GEOS 3.6.1, GDAL 2.2.3, PROJ 4.9.3
nc <- st_read(system.file("shape/nc.shp", package="sf"),
              as_tibble = TRUE)
#> Reading layer `nc' from data source `C:\Program Files\R\R-3.6.2\library\sf\shape\nc.shp' using driver `ESRI Shapefile'
#> Simple feature collection with 100 features and 14 fields
#> geometry type:  MULTIPOLYGON
#> dimension:      XY
#> bbox:           xmin: -84.32385 ymin: 33.88199 xmax: -75.45698 ymax: 36.58965
#> epsg (SRID):    4267
#> proj4string:    +proj=longlat +datum=NAD27 +no_defs

nc %>%
  as_tibble() %>%
  select(CNTY_ID) %>%
  mutate(rn = row_number()) %>%
  flip(inner_join)(nc, by = "CNTY_ID")
#> Simple feature collection with 100 features and 15 fields
#> geometry type:  MULTIPOLYGON
#> dimension:      XY
#> bbox:           xmin: -84.32385 ymin: 33.88199 xmax: -75.45698 ymax: 36.58965
#> epsg (SRID):    4267
#> proj4string:    +proj=longlat +datum=NAD27 +no_defs
#> # A tibble: 100 x 16
#>     AREA PERIMETER CNTY_ CNTY_ID NAME  FIPS  FIPSNO CRESS_ID BIR74 SID74 NWBIR74
#>    <dbl>     <dbl> <dbl>   <dbl> <fct> <fct>  <dbl>    <int> <dbl> <dbl>   <dbl>
#>  1 0.114      1.44  1825    1825 Ashe  37009  37009        5  1091     1      10
#>  2 0.061      1.23  1827    1827 Alle… 37005  37005        3   487     0      10
#>  3 0.143      1.63  1828    1828 Surry 37171  37171       86  3188     5     208
#>  4 0.07       2.97  1831    1831 Curr… 37053  37053       27   508     1     123
#>  5 0.153      2.21  1832    1832 Nort… 37131  37131       66  1421     9    1066
#>  6 0.097      1.67  1833    1833 Hert… 37091  37091       46  1452     7     954
#>  7 0.062      1.55  1834    1834 Camd… 37029  37029       15   286     0     115
#>  8 0.091      1.28  1835    1835 Gates 37073  37073       37   420     0     254
#>  9 0.118      1.42  1836    1836 Warr… 37185  37185       93   968     4     748
#> 10 0.124      1.43  1837    1837 Stok… 37169  37169       85  1612     1     160
#> # … with 90 more rows, and 5 more variables: BIR79 <dbl>, SID79 <dbl>,
#> #   NWBIR79 <dbl>, geometry <MULTIPOLYGON [°]>, rn <int>
```

If this would be a useful addition to purrr I will happily raise a PR.

tomjemmett:
Fixes #728

This implements the function, but I think the documentation is a bit week at the moment so happy to improve upon that if needed.


lionel-:
Since that doesn't seem a very useful function and it produces less clear code than a formula lambda with named arguments, I don't think we want to use up a short name like `flip()`. Generally I'd lean towards not adding that function to purrr, even with a longer name.

lionel-:
Thanks for the suggestion and PR! Closing, as discussed in #728.

PR #754, opened 2020-03-18
Correctly treat NULL return values from .f() in modify() family

### LINKED ISSUE #655: modify_depth fails when higher levels have NULLs
Hi,
`modify_depth` fails when higher levels have a mix of NULLs and non-NULLs e.g.
```r
> aa <- list(a = NULL, b = list(b1 = NULL, b2 = "hello"))
> aa
$a
NULL

$b
$b$b1
NULL

$b$b2
[1] "hello"


> modify_depth(aa, 
+              .depth = 2,
+              is.character,
+              .ragged = TRUE
+              )
Error in .x[[i]] : subscript out of bounds
```
I would expect
```r
$a
NULL

$b
$b$b1
NULL

$b$b2
[1] "TRUE"
```

Works when the first level has no NULLs
```r
> aa <- list(a = "hi", b = list(b1 = NULL, b2 = "hello"))
> aa
$a
[1] "hi"

$b
$b$b1
NULL

$b$b2
[1] "hello"


> modify_depth(aa, 
+              .depth = 2,
+              is.character,
+              .ragged = TRUE
+              )
$a
[1] "TRUE"

$b
$b$b1
[1] FALSE

$b$b2
[1] TRUE
```
And btw why is the upper level type-stable, and not the lower level ?

And why is it the same when `.ragged = FALSE` ?
```r
> modify_depth(aa, 
+              .depth = 2,
+              is.character
+              )
$a
[1] "TRUE"

$b
$b$b1
[1] FALSE

$b$b2
[1] TRUE
```

Thanks!

### LINKED ISSUE #746: modify_if() can't change a list element to NULL
I can't find a way to set certain list elements to `NULL` with `modify_if()`. Any approach I've tried either removes the elements (which seems weird, because I thought `modify_*()` was size-stable) or creates an extra layer of list-nesting. This may be related to some other issues around pluck-y assignment and modify, such as #636, #634.

``` r
library(purrr)

(x <- list("a", "b"))
#> [[1]]
#> [1] "a"
#> 
#> [[2]]
#> [1] "b"

(x_I_want <- list("a", NULL))
#> [[1]]
#> [1] "a"
#> 
#> [[2]]
#> NULL

modify_if(x, ~.x == "b", NULL)
#> Error: Can't convert NULL to function
#> Backtrace:
#>     █
#>  1. ├─purrr::modify_if(x, ~.x == "b", NULL)
#>  2. └─purrr:::modify_if.default(x, ~.x == "b", NULL)
#>  3.   ├─purrr::as_mapper(.f, ...)
#>  4.   └─purrr:::as_mapper.default(.f, ...)
#>  5.     └─rlang::as_function(.f)
#>  6.       └─rlang:::abort_coercion(x, friendly_type("function"))

modify_if(x, ~.x == "b", ~ NULL)
#> [[1]]
#> [1] "a"

modify_if(x, ~.x == "b", ~ list(NULL))
#> [[1]]
#> [1] "a"
#> 
#> [[2]]
#> [[2]][[1]]
#> NULL

modify_if(x, ~.x == "b", function(z) NULL)
#> [[1]]
#> [1] "a"

modify_if(x, ~.x == "b", function(z) list(NULL))
#> [[1]]
#> [1] "a"
#> 
#> [[2]]
#> [[2]][[1]]
#> NULL
```

<sup>Created on 2020-02-12 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0.9001)</sup>

### LINKED ISSUE #753: `modify` fails when `.f` returns NULL
`purrr::modify.default` should check explicitly for null values

``` r

X <- list(a = 1, b = NULL, c = 3)
purrr:::modify(X, identity)
#> Error in .x[[i]]: subscript out of bounds
```

<sup>Created on 2020-03-18 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

Likely duplicates #655 and #746. Creating the new issue with simple reprex to create a sense of urgency :) If a simple check for NULL in modify.default is ok with you I can create a PR and add tests.

vspinu:
Fix #753, Fix #655, Fix #746

I am using checking length to see if `[] <- list(NULL)` modifies length

### DIFF
--- a/R/modify.R
+++ b/R/modify.R
@@ -135,8 +135,12 @@ modify <- function(.x, .f, ...) {
 modify.default <- function(.x, .f, ...) {
   .f <- as_mapper(.f, ...)
 
-  for (i in seq_along(.x)) {
-    .x[[i]] <- .f(.x[[i]], ...)
+  if (is.recursive(.x)) {
+    .x <- modify_list_with_ix(.x, seq_along(.x), .f, ...)
+  } else {
+    for (i in seq_along(.x)) {
+      .x[[i]] <- .f(.x[[i]], ...)
+    }
   }
 
   .x
@@ -155,12 +159,18 @@ modify_if.default <- function(.x, .p, .f, ..., .else = NULL) {
   index <- seq_along(.x)
 
   .f <- as_mapper(.f, ...)
-  for (i in index[sel]) {
-    .x[[i]] <- .f(.x[[i]], ...)
+  if (is.recursive(.x)) {
+    .x <- modify_list_with_ix(.x, index[sel], .f, ...)
+  } else {
+    for (i in index[sel]) {
+      .x[[i]] <- .f(.x[[i]], ...)
+    }
   }
 
   if (!is_null(.else)) {
     .else <- as_mapper(.else, ...)
+    if (is.recursive(.x))
+      .x <- modify_list_with_ix(.x, index[!sel], .f, ...)
     for (i in index[!sel]) {
       .x[[i]] <- .else(.x[[i]], ...)
     }
@@ -473,3 +483,23 @@ inv_which <- function(x, sel) {
     stop("unrecognised index type", call. = FALSE)
   }
 }
+
+# protected list modifier
+modify_list_with_ix <- function(x, ix, f, ...) {
+  orig_len <- length(x)
+  for (i in ix) {
+    elt <- f(x[[i]], ...)
+    if (is.null(elt)) {
+      ## this might break if [] is unwisely overloaded
+      x[i] <- list(NULL)
+      if (length(x) != orig_len)
+        stop(sprintf("Modifier returned NULL. Cannot assin NULLs into %s",
+                     friendly_type_of(x)), call. = FALSE)
+    } else {
+      x[[i]] <- elt
+    }
+  }
+  x
+}
+
+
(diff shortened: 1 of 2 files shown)


lionel-:
This is a bit too complex for me to review right now, but you can use this branch with confidence as we will definitely fix this in the next non-patch release of purrr.

I would use `is_list(x) && !is.data.frame(x)` instead of `is.recursive()`, as the latter is too permissive. The former is too permissive as well, but an improvement.

Style-wise, can you please double check the patch is consistent with our style. This includes:

- Wrap single-line branches in `{`.
- Sentence case for comments, no full stop unless two sentences.
- `elt` instead of `el`.
- `abort()` instead of `stop()`.
-  ```
    foo(
      a,
      b,
    )
    ```

    instead of

    ```
    foo(a,
        b)
    ```

vspinu:
> I would use is_list(x) && !is.data.frame(x)


This would be an unnecessary restriction.  In case of data.frames and other containers which don't behave as we expect on  `[] <- list(NULL)` (aka keep same length) we throw with an informative error.

vspinu:
Ok, applied all you suggestions except for abort and data.frame check as per my comment above. The `modify.R` uses stop, please modify yourself all of them as you see fit. 

>  but you can use this branch with confidence as we will definitely fix this in the next non-patch release of purrr.

I could, but there are other people involved and a production system which I don't want to rub needlessly. 
Could you just pull it for now and deal with the "sophistication" after? The PR is simple, all tests passing, but the bug is critical.

PR #771, opened 2020-06-26
Make the imap docs more consistent.

rorynolan:
I think here you either need to remove the`_xxx` or add it to the `map2()` later in the sentence to make them `map2_xxx()`. Personally, I don't like the `_xxx` since the argument is also `x`.


lionel-:
Thanks!

Issue #776, opened 2020-07-10
Bug: list_modify and zap do not remove key-value pairs whose values are lists


adamroyjones:
I am using purrr 0.3.4 with R 3.6.3 (on Ubuntu 20.04).

The [documentation](https://purrr.tidyverse.org/reference/list_modify.html) to `list_modify` indicates that one should "[u]se `zap()` to remove values" from a list. This doesn't seem to work in the way that I, a naïve user, would expect: it does not work when the value associated with a key is a list:

```r
library(purrr)

list(foo = 1, bar = 2) %>% list_modify(foo = zap())
# { bar: 2 }

list(foo = c(1, 2), bar = 3) %>% list_modify(foo = zap())
# { bar: 3 }

list(foo = list(fst = 1, snd = 2), bar = 3) %>% list_modify(foo = zap())
# { foo: { fst: 1, snd: 2 }, bar: 3 }
```

(This behaviour is reflected in the documentation itself.)

Is this expected behaviour? Am I being an idiot?

adamroyjones:
I should add that, if this turns out to be unexpected behaviour, I'm happy to have a stab at fixing it.

adamroyjones:
I've raised a pull request that hopefully fixes this (#777).

PR #777, opened 2020-07-10
Changes list_modify to be able to remove key-value pairs whose value is a list

adamroyjones:
The documentation to `list_modify` indicates that one should "[u]se `zap()` to remove values" from a list. This does not work in the expected way when attempting to remove a key-value pair whose value is a list:

```r
library(purrr)

list(foo = 1, bar = 2) %>% list_modify(foo = zap())

list(foo = c(1, 2), bar = 3) %>% list_modify(foo = zap())

list(foo = list(fst = 1, snd = 2), bar = 3) %>% list_modify(foo = zap())
```

The first two examples above return an object identical to `list(bar = 3)`; the third is such that the call to `list_modify` is a no-op. This pull request fixes this behaviour. It also adds a unit test to cover this case.

Issues: #776

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,9 @@
 
 # purrr 0.3.4
 
+* Fixed issue in `list_modify()` that prevented lists from being
+  removed with `zap()` (@adamroyjones, #777).
+
 * Added documentation for exporting functions created with purrr
   adverb (@njtierney, #668). See `?faq-adverbs-export`.
 
--- a/R/list-modify.R
+++ b/R/list-modify.R
@@ -69,9 +69,10 @@ list_recurse <- function(x, y, base_case) {
     abort("`...` arguments must be either all named, or all unnamed")
   }
 
+  # N.B. is_list(zap()) is TRUE.
   if (is_null(y_names)) {
     for (i in rev(seq_along(y))) {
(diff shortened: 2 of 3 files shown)


adamroyjones:
I've removed the whitespace changes (but it's a little unclear whether I need to re-raise the request; this is the first time I've opened a (real) PR in GitHub, so please let me know if I've done something wrong).

I'm quite happy to contribute more of my time to purrr (and the tidyverse more broadly)—is there a register of issues, or something similar?

lionel-:
No need to reopen the pull request as this would just add noise. Pushing the changes directly in the pull request is the way to go.

You also need to add a NEWS bullet explaining your change and linking to your github handle and the relevant github issue. I've done that now.

> is there a register of issues, or something similar?

We sometimes have "help wanted" or "good first contribution" labels on github issues.

Thanks for the fix!

BrianDiggs:
I would like to recommend an additional test that fails on the current release but should be fixed by this patch:

```r
expect_equal(list_modify(list(list(1,2), 2, 3), zap(), zap()), list(3))
```

This test exercises the branch of code which is used for replacing a list when the arguments are not named. It parallels the added test in this pull request for named arguments.

Issue #758, opened 2020-04-16
Use vctrs instead of dplyr


hadley:

hadley:
This involves replacing `bind_rows()`/`bind_cols()` with `vec_cbind()`/`vec_rbind()` and passing along all arguments.

hadley:
Be sure to look into https://github.com/tidyverse/purrr/issues/376 and https://github.com/tidyverse/purrr/issues/838

hadley:
Suggestion from @lionel- — deprecated `map_dfr()` and `map_dfc()` in favour of `map_rbind()` and `map_cbind()`

PR #781, opened 2020-08-04
Use vctrs in `simplify()` and `as_vector()`

lionel-:
Progress towards r-lib/vctrs#183
Progress towards #758

- Use `vec_simplify()` in `simplify()` and `list_as_vector()` in `as_vector()`.

- The `.type` argument of `simplify()` is completely deprecated.

- The `.type` argument of `as_vector()` must now be a size-zero prototype. The support for type names (like "double") and type moulds (prototypes whose length is meaningful) is now deprecated. This is part of a transition towards vctrs prototypes. A compatibility layer allows a period of transition during the deprecation period.


lionel-:
Closing in favour of #909.

PR #782, opened 2020-08-05
Add `.simplify` argument to `accumulate()` and `accumulate2()`

### LINKED ISSUE #774: Bug: accumulate2 returns a list, not a vector
The documentation for `accumulate` and `accumulate2` says that both functions return vectors ("A vector the same length of `.x` with the same names as `.x`"). In fact, `accumulate` returns a vector, but `accumulate2` returns a list.

## Expected behavior

`accumulate2` returns a vector

## Observed behavior

`accumulate2` returns a list

## Reprex

``` r
library(purrr)

f <- function(x, y, z) {
  x + y + z
}

accumulate(seq(5), f, 1)
#> [1]  1  4  8 13 19

accumulate2(seq(5), seq(4), f)
#> [[1]]
#> [1] 1
#> 
#> [[2]]
#> [1] 4
#> 
#> [[3]]
#> [1] 9
#> 
#> [[4]]
#> [1] 16
#> 
#> [[5]]
#> [1] 25
```

<sup>Created on 2020-07-10 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

## Session Info

``` r
> sessionInfo()
R version 4.0.2 (2020-06-22)
Platform: x86_64-w64-mingw32/x64 (64-bit)
Running under: Windows 10 x64 (build 18363)

Matrix products: default

locale:
[1] LC_COLLATE=English_United States.1252  LC_CTYPE=English_United States.1252    LC_MONETARY=English_United States.1252 LC_NUMERIC=C                          
[5] LC_TIME=English_United States.1252    

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[1] reprex_0.3.0

loaded via a namespace (and not attached):
 [1] ps_1.3.3               digest_0.6.25          packrat_0.5.0          crayon_1.3.4           R6_2.4.1               lifecycle_0.2.0       
 [7] magrittr_1.5           evaluate_0.14          pillar_1.4.4           rlang_0.4.6            rstudioapi_0.11.0-9000 fs_1.4.2              
[13] callr_3.4.3            whisker_0.4            vctrs_0.3.1            ellipsis_0.3.1         rmarkdown_2.3.2.9002   tools_4.0.2           
[19] purrr_0.3.4            processx_3.4.3         xfun_0.15              compiler_4.0.2         pkgconfig_2.0.3        clipr_0.7.0           
[25] htmltools_0.5.0        knitr_1.29             tibble_3.0.2          
```

lionel-:
Because the original implementation was based on `Reduce()` we've accidentally implemented the original `accumulate()` function with automatic simplification. I think ideally `accumulate()` should always return a list and simplification would be achieved by piping into `simplify()`. However that would probably be too large a breaking change. Instead, `accumulate()` now has a `.simplify` argument to optionally disable simplification.

`accumulate2()` gains this argument as well. The default is set to `TRUE` even though it didn't use to automatically simplify. This is because consistency with `accumulate()` seems more important. This is a behaviour change but I don't see any use of `accumulate2()` on CRAN. Closes #774.


hadley:
Also fixes #809

R/reduce.R
```diff
@@ -491,6 +501,17 @@ accumulate_names <- function(nms, init, dir) {
...
 }
 
+acc_simplify <- function(x) {
+  # It would be unappropriate to simplify the result rowwise with
+  # `accumulate()` because it has invariants defined in terms of
+  # `length()` rather than `vec_size()`
+  if (some(x, is.data.frame)) {
+    x
+  } else {
+    vec_simplify(x)
```
hadley, line 511: I think this is the only place that `vec_simplify()` is used, so it probably should be incorporated into this function.

hadley:
Should we allow a `.ptype` here?

lionel-:
Closing in favour of #909.

PR #783, opened 2020-08-05
Recycle names in `map2()`

### LINKED ISSUE #779: Recycling `.x` should recycle names
When recycling named `.x`, names aren't recycled and don't appear on the output. I think it would make sense to recycle names here

```r
> purrr::map2(c(a=1), 1:2, ~.x)
[[1]]
[1] 1
[[2]]
[1] 1
```

lionel-:
Closes #779


DavisVaughan:
That should mean we get recycling for `pmap()` too.

Also, this is another case to maybe use the R level `length()` callback https://github.com/tidyverse/purrr/pull/783/files#diff-ed6bf403234aac5ffa77de8c10222317R15

PR #796, opened 2020-09-24
add alternative example for using adverbs in packages

### LINKED ISSUE #795: Clarification on using `safely` in a package
Would it be possible to clarify the usage of `safely` within a package.  Reading the [documentation](https://purrr.tidyverse.org/reference/faq-adverbs-export.html) it discusses using the `.onload` hook to prevent problems when `purrr` is updated but is it not preferable to instead create explicit wrapper functions, e.g.

```r
fun1 <- function(x) {
    x^2
}

# this would need to be done with .onLoad
fun2 <- purrr::safely(fun1)

# is there a problem with doing the following instead and exporting?
fun3 <- function(x) purrr::safely(fun1)(x)
```

This may be a misunderstanding on my part, so it would be great to get clarification.

tjtnew:
closes #795


lionel-:
Thanks!

PR #799, opened 2020-10-20
Improve doc

salim-b:
- use term "dynamic dots" instead of "tidy dots" and link to proper help page
- fix comma


lionel-:
Thanks!

PR #801, opened 2020-11-03
implement {map,map2,pmap}_{lgl,int,dbl,chr,raw}_matrix()

### LINKED ISSUE #798: Add `map_*_matrix` functions similar to `vapply` with FUN.VALUE of length > 1.
Sometimes I need to compute a matrix column-wise using a function.
Base-R code to achieve that would be `vapply()` with a `FUN.VALUE` of length greater than 1.

For a toy example with an easier solution consider this function:

```r
rnorm_mat <- function(n, means = 0:5, sd = 1) {
  vapply(
    means,
    function(mu) rnorm(n, mean = mu, sd = sd),
    numeric(n)
  )
}
```

If `n > 1`, this works as expected and returns a matrix of normal random variates with true column means `means` and constant standard deviation `sd`. However, the return type is not stable for `n = 1` in which case the value will be a length `length(means)` vector instead of a 1-row matrix.

I propose a family of functions `map_*_matrix()` which _always_ return a matrix and behaviour akin to

```r
map_dbl_matrix <- function(.x, .f, .n, ...) {
  .f <- purrr::as_mapper(.f, ...)
  res <- vapply(.x, .f, numeric(.n), ...)
  dim(res) <- c(.n, length(.x))
  res
}
```

Maybe additional convenience arguments such as `.byrow = FALSE` to decide whether to create the matrix rowwise or column-wise are necessary as well.

If there is interest in such a feature, I'll gladly try to get a PR started, so please let me know.

AshesITR:
closes #798

I'd like to have general feedback on the approach before creating tests and documentation.
The PR adds a new family of mapping functions, ending in `_matrix`. They have a new required argument, `.n`, the number of elements that `.f` returns in a single call and an optional argument, `.by_row = FALSE` that allows transposing the result matrix.

Default behaviour returns a matrix with `.n` rows and `length(.x)` columns of the respective type.

 * Are the argument names and locations fine?
 * The C implementation has a lot of duplication from `map_impl`, `map2_impl` and `pmap_impl`. Would it be worth deduplicating the codebase by altering / extending the original _impl functions?

 - [x] implement
 - [ ] test
 - [ ] document

### DIFF
--- a/R/map.R
+++ b/R/map.R
@@ -219,6 +219,41 @@ map_raw <- function(.x, .f, ...) {
 }
 
 #' @rdname map
+#' @export
+map_lgl_matrix <- function(.x, .f, .n, ..., .by_row = FALSE) {
+  .f <- as_mapper(.f, ...)
+  .Call(map_matrix_impl, environment(), ".x", ".f", "logical", .n, .by_row)
+}
+
+#' @rdname map
+#' @export
+map_chr_matrix <- function(.x, .f, .n, ..., .by_row = FALSE) {
+  .f <- as_mapper(.f, ...)
+  .Call(map_matrix_impl, environment(), ".x", ".f", "character", .n, .by_row)
+}
+
+#' @rdname map
+#' @export
+map_int_matrix <- function(.x, .f, .n, ..., .by_row = FALSE) {
+  .f <- as_mapper(.f, ...)
+  .Call(map_matrix_impl, environment(), ".x", ".f", "integer", .n, .by_row)
+}
+
+#' @rdname map
+#' @export
+map_dbl_matrix <- function(.x, .f, .n, ..., .by_row = FALSE) {
+  .f <- as_mapper(.f, ...)
+  .Call(map_matrix_impl, environment(), ".x", ".f", "double", .n, .by_row)
+}
+
+#' @rdname map
+#' @export
+map_raw_matrix <- function(.x, .f, .n, ..., .by_row = FALSE) {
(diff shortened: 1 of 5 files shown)


AshesITR:
Ubuntu on 3.2 fails with

>  Error in loadNamespace(j <- i[[1L]], c(lib.loc, .libPaths()), versionCheck = vI[[j]]) : 
  there is no package called ‘digest’

lionel-:
Hello,

Thank you for looking into this. However it's probably a bad time for such a contribution. We will change the internals of purrr to vctrs in the next major version. At that point you could supply a `ptype` to get a matrix output. I think that would be preferable to extending the API of purrr for seldomly used target types.

AshesITR:
Hello lionel,

thanks for your feedback. Are there issues I can help work on for this step?
I'd really love to have fast map_*_matrix functions available as I see many uses for them in numerical applications.

lionel-:
You can try wrapping around https://github.com/r-lib/vctrs/pull/1227 and see if you can implement `map_lgl_matrix()` etc. If that is sufficient, these functions could eventually go in a separate package rather than in purrr.

There probably are performance issues in vctrs with matrices and arrays. If that's the case, reports would be helpful.

FYI I'm likely not resuming work on vctrs until December or January.

hadley:
Thanks for working on this, but per my comment in the related issue, at this time we don't want to expand the API of purrr this much. Either the upcoming vctrs work will help resolve the problem, or you might consider putting these functions in their own package.

Issue #149, opened 2015-12-10
Progress bars


hadley:
Would be nice to have support for progress bars in all map functions. This is a nice feature of plyr.

Could use https://github.com/gaborcsardi/progress, although we might need to ask @gaborcsardi to also provide a C API.

gaborcsardi:
There is a header only C++ API, isn't that good? Although I have to say that is not very well tested and have less features. https://github.com/gaborcsardi/progress#c-api

gaborcsardi:
It would also make sense to put the C++ part in another package, as it is completely independent.

lionel-:
This will not work with mapping functions because we eval R functions from the C code :/ Any user interruption or R error will cause a long jump that bypasses all C++ destructors.

Thus if you have any data on the heap, you'll get leaks. For example all STL containers or even a simple `std::string` allocate memory dynamically so need to be destructed appropriately. See discussion in https://github.com/hadley/purrr/commit/e2def88a4039b15e2c8f92247808264fd03bcf4a

gaborcsardi:
Well, there is an R API and a C++ API. It seems reasonable that you would be able to use at least one of them. :)

vixr:
+1 for this feature

gaborcsardi:
FWIW I wanted to note that I am adding some new progress bar API, which has the nice feature of having (almost) zero overhead when the progress bars are not shown (e.g. non-interactive use), in addition to ease of use. This is how it will look:

```r
progress %~~% lapply(seq, fun, ...)
```

If progress bars are turned off, then it simply runs the `lapply`. If progress bars are on, then it appends the progress bar ticks to `fun`.

I am saying this, because it would be great to use it for `purrr` functions as well.

lionel-:
It'd probably be more purrr-like to have an adverb functional or function operator that takes mapping functionals and add progress bars to them. With a functional:

```r
# ..f must be a function that takes a .x and a .f
# ..f must have the usual purrr signature ..f(.x, .f, ...)
with_progress <- function(..x, ..f, .f, ...) {
  .f <- add_progress(.f, length(..x))
  ..f(..x, .f, ...)
}

mtcars %>% with_progress(map, as.character)
```

lionel-:
A quick thought: I think it's natural to have adverb functionals when we're modifying another functional, like in the example above. But otoh it's natural to have adverb function operators when we're modifying a regular function, e.g. `safely()`, `lift()`, etc.

[...]

PR #830, opened 2021-06-01
Progress bar support

gaborcsardi:
@lionel FYI. Comments are welcome.


src/map.c
```diff
@@ -26,14 +28,17 @@ void check_vector(SEXP x, const char *name) {
...
+               SEXP progress) {
   // Create variable "i" and map to scalar integer
   SEXP i_val = PROTECT(Rf_ScalarInteger(1));
   SEXP i = Rf_install("i");
   Rf_defineVar(i, i_val, env);
 
+  SEXP bar = PROTECT(cli_progress_bar(n, progress));
   SEXP out = PROTECT(Rf_allocVector(type, n));
   for (int i = 0; i < n; ++i) {
+    if (SHOULD_TICK) cli_progress_set(bar, i);
```
lionel-, line 41: Should that macro/constant be `CLI_SHOULD_TICK`?

lionel- (commented):
Nice to see progress on that front? Do you have some benchmarks for the enabled/disabled cases?

src/map.c
gaborcsardi: yeah.

gaborcsardi:
> Nice to see progress on that front? 

Hah!

> Do you have some benchmarks for the enabled/disabled cases?

The cost of `if (SHOULD_TICK) cli_progress_set(bar, i);` is about half a nanosecond on my laptop, per iteration. 

The actual progress bar display is not fast yet (cli theming, pluralization, interpolation, etc.) but by default there are only 5 updates per second, and we can certainly make cli much-much faster in the near future.

lionel-:
> Nice to see progress on that front?

oh I meant to write "!" not "?" ;)

gaborcsardi:
Picking up this old thread. :D

>  Do you have some benchmarks for the enabled/disabled cases?

There are a bunch of benchmarks at https://cli.r-lib.org/dev/articles/progress-benchmark.html#c-benchmarks

What is important for purrr is that the per iteration cost of the C API is extremely small, about 1ns on my machine. 

The other cost is the display cost, but we only update the display at most 5 times per second. (Not counting updates explicitly forced by the user.) This is 6-7ms in the article, but that's actually wrong I believe, because knitr/`evaluate()`/`capture.output()` has a big overhead, and if I measure it interactively it is about 2ms on both macOS and Linux.

gaborcsardi:
Questions:

- [ ] Is this the API we want?
- [ ] Should `.progress = FALSE` be the default or something like `.progress = getOption("purrr.progress")`?
- [ ] Should we make an effort to set parameters automatically? See parameters of `config` here: https://cli.r-lib.org/dev/articles/progress-advanced.html#c-api-reference. E.g. we could set `name`, maybe, based on the current package and the function name?
    ```r
    out <- map(
      1:1000, 
      function(x) Sys.sleep(.01), 
      .progress = list(name = "mypkg::myfun", clear = FALSE)
    )
    ```

hadley:
Let's make it `.progress = NULL` for now, continuing to default to `FALSE`. That gives us room to experiment with the API later.

It feels like there's some similar principle to what we show in errors, but maybe it's better to start with show (e.g) `purrr::map()` and then figure out some way for the user to customise later? That doesn't even have to be a 1.0.0 feature, IMO.

hadley:
Fixes #149.

R/map.R
```diff
@@ -23,6 +23,7 @@
 #' @inheritParams as_mapper
 #' @param .x A list or atomic vector.
 #' @param ... Additional arguments passed on to the mapped function.
+#' @param .progress Whether to show a progress bar.
```
hadley, line 26: If `TRUE`, will only show a progress bar if deemed necessary, using some heuristics to ensure that very short tasks don't generate a progress bar.

src/map.c
```diff
@@ -26,14 +28,17 @@ void check_vector(SEXP x, const char *name) {
...
 // call must involve i
-SEXP call_loop(SEXP env, SEXP call, int n, SEXPTYPE type, int force_args) {
+SEXP call_loop(SEXP env, SEXP call, int n, SEXPTYPE type, int force_args,
+               SEXP progress) {
   // Create variable "i" and map to scalar integer
   SEXP i_val = PROTECT(Rf_ScalarInteger(1));
   SEXP i = Rf_install("i");
   Rf_defineVar(i, i_val, env);
 
+  SEXP bar = PROTECT(cli_progress_bar(n, progress));
```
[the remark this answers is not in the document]
gaborcsardi: It can also be a list of configuration options (see example above), but yeah, it does validate the input: https://github.com/r-lib/cli/blob/1fd7cb4f16eb672528627d9fb79badfa853bf672/src/progress.c#L175

gaborcsardi:
> then figure out some way for the user to customise later?

Oh, it is already fully customizable, `.progress` can be a list of options, see `config` here: https://cli.r-lib.org/dev/articles/progress-advanced.html#cli_progress_bar

We should probably document this in purrr as well.

R/map.R
gaborcsardi: I added a bullet for better docs.

hadley:
Fixes #149

gaborcsardi:
How should we document this? The documentation is probably a bit too long to inherit it into multiple manual pages, so maybe we create a `purrr-progress-bar` manual page, and link to there?

hadley:
Yeah, I think that's a good idea. Do you want to sketch that out? Then I'm happy to carry this PR over the finish line.

gaborcsardi:
Should we add a global option to turn on progress bars? That would make it easier to test this.

R/map-raw.R
```diff
@@ -9,42 +9,46 @@
 #'
 #' @keywords internal
 #' @export
-map_raw <- function(.x, .f, ...) {
+map_raw <- function(.x, .f, ..., .progress = NULL) {
```
hadley, line 12: Since these are now deprecated, I think you can drop `.progress` support.

gaborcsardi:
I wrote some docs. Should we add some examples? Nice examples (that include the output in HTML) would need asciicast, which might be stable enough now for this.

hadley:
I think we should merge it (to avoid more merge conflicts as I update the docs broadly) and then maybe consider some examples as part of #915? We'll definitely also want them for the blog post.

hadley:
Do you want to add a news bullet?

gaborcsardi:
Added NEWS bullet. 

Btw. another thing is to think about how to add progress bars for the more complex cases, that are not a simple iteration. Plus I guess there are also simple iterations that I missed.

PR #879, opened 2022-08-25
Correctly modify with NULL values

### LINKED ISSUE #655: modify_depth fails when higher levels have NULLs

### LINKED ISSUE #746: modify_if() can't change a list element to NULL

### LINKED ISSUE #753: `modify` fails when `.f` returns NULL

hadley:
Fixes #655. Fixes #746. Fixes #753. Closes #754.

`list_assign()` from @DavisVaughan. Tests from @vspinu. I chose not to worry about the data frame case, as the existing error is ok.

### DIFF
--- a/R/modify.R
+++ b/R/modify.R
@@ -136,7 +136,7 @@ modify.default <- function(.x, .f, ...) {
   .f <- as_mapper(.f, ...)
 
   for (i in seq_along(.x)) {
-    list_assign2(.x, i) <- .f(.x[[i]], ...)
+    list_slice2(.x, i) <- .f(.x[[i]], ...)
   }
 
   .x
@@ -156,13 +156,13 @@ modify_if.default <- function(.x, .p, .f, ..., .else = NULL) {
 
   .f <- as_mapper(.f, ...)
   for (i in index[sel]) {
-    list_assign2(.x, i) <- .f(.x[[i]], ...)
+    list_slice2(.x, i) <- .f(.x[[i]], ...)
   }
 
   if (!is_null(.else)) {
     .else <- as_mapper(.else, ...)
     for (i in index[!sel]) {
-      list_assign2(.x, i) <- .else(.x[[i]], ...)
+      list_slice2(.x, i) <- .else(.x[[i]], ...)
     }
   }
 
@@ -336,7 +336,7 @@ modify2.default <- function(.x, .y, .f, ...) {
   .y <- args[[2]]
 
   for (i in seq_along(.x)) {
-    list_assign2(.x, i) <- .f(.x[[i]], .y[[i]], ...)
+    list_slice2(.x, i) <- .f(.x[[i]], .y[[i]], ...)
   }
 
   .x
@@ -474,7 +474,7 @@ inv_which <- function(x, sel) {
   }
 }
 
-`list_assign2<-` <- function(x, i, value) {
+`list_slice2<-` <- function(x, i, value) {
   if (is.null(value)) {
     x[i] <- list(NULL)
   } else {
--- a/R/modify.R
+++ b/R/modify.R
@@ -136,7 +136,7 @@ modify.default <- function(.x, .f, ...) {
   .f <- as_mapper(.f, ...)
 
   for (i in seq_along(.x)) {
-    .x[[i]] <- .f(.x[[i]], ...)
+    list_assign2(.x, i) <- .f(.x[[i]], ...)
   }
 
   .x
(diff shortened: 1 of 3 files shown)


R/modify.R
```diff
@@ -473,3 +473,12 @@ inv_which <- function(x, sel) {
     stop("unrecognised index type", call. = FALSE)
   }
 }
+
+`list_assign2<-` <- function(x, i, value) {
```
lionel-, line 477: Interesting, I have never seen an assigner function like this before. Should it be called `list_assigner2()`, i.e. a noun? With a verb I think I'm expecting a UI like `vec_assign()`.

lionel-: oh it should probably be `list_slice2<-` actually?

lionel- (approved):

R/modify.R
hadley: Oh yeah, duh.

PR #884, opened 2022-08-25
First tranche of NSE deprecations

### LINKED ISSUE #858: Is `update_list()` deprecated?  Are there usage examples?
I am looking at the [documentation page](https://purrr.tidyverse.org/reference/list_modify.html), and it mentions that it may be deprecated, but I do not see any indication of that.  Also, I see no usage examples for this function.  I have no idea of what arguments are taken, how to use, etc.  If it is indeed still not deprecated, I am requesting some usage examples be added to the documentation.

### LINKED ISSUE #874: Deprecate `*_at()`
(i.e. all functions that use `at_selection()`). 

These were designed to mimic dplyr functions that no longer exist, and I think it's now more clear that tidyselect usage doesn't belong in purrr.

### LINKED ISSUE #877: Deprecate `rerun()`
Since we now believe NSE functions don't belong in purrrr

hadley:
Fixes #858. Fixes #877. Fixes #874.


R/rerun.R
```diff
@@ -15,22 +16,24 @@
...
+#' # old
 #' 10 %>%
 #'   rerun(x = rnorm(5), y = rnorm(5)) %>%
 #'   map_dbl(~ cor(.x$x, .x$y))
+#' # new
+#' 1:10 %>%
+#'   map(~ list(x = rnorm(5), y = rnorm(5))) %>%
+#'   map_dbl(~ cor(.x$x, .x$y))
 rerun <- function(.n, ...) {
+  lifecycle::deprecate_warn("0.4.0", "rerun()", "map()")
```
lionel-, line 35: I'm a bit worried that this function might be used in scripts. Should we just supersede?

tests/testthat/test-rerun.R
```diff
@@ -1,17 +1,28 @@
+test_that("is deprecated", {
+  expect_snapshot({
+    . <- rerun(1, 1)
+  })
+
+})
+
 test_that("single unnamed arg doesn't get extra list", {
+  options(lifecycle_verbosity = "quiet")
```
lionel-, line 9: Here and elsewhere: Should be `local_options()`. Could also be pulled at top level since the whole file tests a deprecated function.

tests/testthat/test-list-modify-update.R
```diff
@@ -91,11 +91,19 @@ test_that("list_merge handles duplicate names", {
...
 # update_list ------------------------------------------------------------
 
+test_that("update_list() is deprecated", {
+  expect_snapshot({
+    . <- update_list(list())
+  })
+})
+
 test_that("can modify element called x", {
+  options(lifecycle_verbosity = "quiet")
```
lionel-, line 101: `local_options()`.

lionel- (commented):
I'm worried about deprecating functions that are likely used in analysis scripts. I know that purrr isn't a 1.0 yet, but shouldn't we treat it as if it were? Is there much downside to just superseding these functions?

R/rerun.R
hadley: You only see deprecation warnings every 8 hours so I don't feel like it's super annoying. What about if I just made more prominent advice?

tests/testthat/test-rerun.R
hadley: Doh

R/rerun.R
lionel-: ah yes it might be worth a few more details about the switch.

PR #886, opened 2022-08-25
Deprecate splice()

### LINKED ISSUE #869: Example code in splice does not run
The example code included in the roxygen2 notes from splice throws an error when this example code is ran. Please refer to 
 ?purrr::splice() 

```r
inputs <- list(arg1 = "a", arg2 = "b")

# splice() concatenates the elements of inputs with arg3
splice(inputs, arg3 = c("c1", "c2")) %>% str()
list(inputs, arg3 = c("c1", "c2")) %>% str()
c(inputs, arg3 = c("c1", "c2")) %>% str()
```

hadley:
Fixes #869


R/splice.R
```diff
@@ -25,6 +20,8 @@
 #' c(inputs, arg3 = c("c1", "c2")) %>% str()
 #' @export
 splice <- function(...) {
+  lifecycle::deprecate_warn("0.4.0", "splice()")
```
lionel-, line 23: Do we now plan for a 1.0.0 right away?

lionel- (approved):

R/splice.R
hadley: Maaaaaybe. If so I'll do another pass to update 0.4.0 to 1.0.0

PR #890, opened 2022-08-26
Improve adverb docs

### LINKED ISSUE #863: Please clarify `possibly()` documentation
Greetings! Thank you for allowing an easier way to catch errors from functions using `possibly()`. 

However, the function is a bit unclear to me in what it returns (under "Value" )

```
Value
safely: wrapped function instead returns a list with components result and error. If an error occurred, error is an error object and result has a default value (otherwise). Else error is NULL.

quietly: wrapped function instead returns a list with components result, output, messages and warnings.

possibly: wrapped function uses a default value (otherwise) whenever an error occurs.
```
What precisely is a "wrapped function"?   I had trouble using it, and I found two stack-overflow posts that describe how to use `possibly()` correctly. 

https://stackoverflow.com/questions/50486527/how-to-use-map-with-possibly
https://stackoverflow.com/questions/72365327/using-purrrpossibly-to-catch-non-positive-definite-matrix/72365359#72365359

Something like: 
`possibly()` takes a function `.f` as input and exports it as a function, say `possibly.f()`. Any arguments that are passed to `possibly.f()` are passed to the function, `.f`.

Thanks!

hadley:
Now each adverb lives in its own file (`adverb-`), has it's own documentation, and consistently links to exporting advice and the other adverbs.

Fixes #863.


R/faq.R
```diff
@@ -1,56 +1,42 @@
...
-#'   purrr::insistently(my_function)(...)
-#' }
-#' ```
+#' requires some precautions because the functions will contain internal
+#' purrr code. This means that creating them once and for all when
+#' the package is built may cause problems when purrr is updated, because
+#' a function that the adverb uses might no longer exist.
+#'
+#' Instead, either create the function on package load or by wrapping the call
+#' within another function, as shown below.
```
lionel-, line 11:
```suggestion
#' Instead, either create the modified function once per session on package load, or 
#' at each usage by wrapping the call within another function, as shown below.
```

lionel- (approved):

PR #894, opened 2022-08-27
Apply vctrs principles to `map()` and `modify()`

### LINKED ISSUE #435: map_vec
Would take a col_type argument a la readr (although the semantics would be a little different)

### LINKED ISSUE #928: Use some vctrs principle in `modify()` and friends
Or maybe this can be `map_vec()`?

```R
modify.vector <- function(.x, .f, ...) {
  map_vec(.x, .f, ..., .ptype = .x)
} 
```

hadley:
Fixes #435. Fixes #928.

For the purposes of discussion. If we agree that these semantics are what we want, need to:

* [x] Add news bullet
* [x] Implement `map2_vec()`
* [x] Implement `pmap_vec()`
* [x] Update modify


R/map.R
```diff
@@ -218,6 +221,27 @@ map_raw <- function(.x, .f, ...) {
...
+  out <- map(.x, .f, ...)
+
+  .ptype <- vec_ptype_common(!!!out, .ptype = .ptype)
+  for (i in seq_along(out)) {
+    if (vec_size(out[[i]]) != 1L) {
+      stop_bad_element_vector(out[[i]], i, .ptype, 1L, what = "Result")
+    }
+  }
+
+  vec_c(!!!out, .ptype = .ptype)
```
lionel-, line 241: Minor: I'd probably use `vec_unchop()` here.

tests/testthat/test-map.R
```diff
@@ -170,3 +170,32 @@ test_that("map() with empty input copies names", {
   expect_identical(map_chr(named_list, identity), named(chr()))
   expect_identical(map_raw(named_list, identity), named(raw()))
 })
+
+
+# map_vec -----------------------------------------------------------------
+
```
lionel-, line 176: Add a test for data frame concatenation/row-binding?

R/map.R
```diff
@@ -218,6 +221,27 @@ map_raw <- function(.x, .f, ...) {
...
+#' @param .ptype If `NULL`, the default, the output type is the common type
+#'   of the elements of the result. Otherwise, supply a "prototype" giving
+#'   the desired type of output.
+#' @importFrom vctrs vec_c vec_size vec_ptype_common
+#' @export
+map_vec <- function(.x, .f, ..., .ptype = NULL) {
+  out <- map(.x, .f, ...)
+
+  .ptype <- vec_ptype_common(!!!out, .ptype = .ptype)
+  for (i in seq_along(out)) {
```
lionel-, line 235: Might be worth adding a `size` argument to `vctrs::list_check_all_vectors()`.

Then this'd be:

```r
out <- map(.x, .f, ...)
list_check_all_vectors(out, size = 1, arg = "")
vec_unchop(out)
```

R/map.R
lionel-: Should we pass a `name_spec` argument?

tests/testthat/test-map.R
lionel-: Needs tests for lists to make the expected behaviour explicit.

lionel- (commented):
Another problem to solve is to generically map over classed lists like `list_of()` with full restoration behaviour. I think that's out of scope for `map_vec()` but could live in another function, perhaps `list_map()`?

R/map.R
hadley: I didn't do use `vec_unchop()` because the docs say "When no indices are supplied, it is identical to vec_c()." Is that not correct, or do you think think `vec_unchop()` conveys the semantics better?

DavisVaughan: Avoiding the `!!!` typically is a little more memory efficient, not sure if there is an rlang reason why this is the case or not

``` r
library(vctrs)

x <- as.list(1:10000)

bench::mark(vec_unchop(x), vec_c(!!!x), iterations = 1000)[c("expression", "median", "mem_alloc")]
#> # A tibble: 2 × 3
#>   expression      median mem_alloc
#>   <bch:expr>    <bch:tm> <bch:byt>
#> 1 vec_unchop(x)   2.37ms    81.8KB
#> 2 vec_c(!!!x)     2.45ms   159.7KB
```

<sup>Created on 2022-08-29 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

R/map.R
DavisVaughan: In the meantime if you want to use the `stop_bad_element_vector()` error message then I'd suggest

`sizes <- list_sizes(out); any(sizes != 1L)`

R/map.R
```diff
@@ -218,6 +221,27 @@ map_raw <- function(.x, .f, ...) {
...
+    if (vec_size(out[[i]]) != 1L) {
+      stop_bad_element_vector(out[[i]], i, .ptype, 1L, what = "Result")
+    }
+  }
+
+  vec_c(!!!out, .ptype = .ptype)
+}
```
DavisVaughan, line 242: Regarding output names, I think we need to follow the `map_int()` behavior of prioritizing names from the input.

i.e. the invariant is that the output is the same size as the input, with the same names.

``` r
library(purrr)

map_dbl(list(x = 1, y = 2), ~ c(a = .x))
#> x y 
#> 1 2
```

I do this in `slider::slide_vec()` with something like this:

```r
# it assumes the size 1 checks have already occurred
> slider:::vec_simplify
function (x, ptype) 
{
    names <- vec_names(x)
    x <- vec_set_names(x, NULL)
    out <- vec_unchop(x, ptype = ptype)
    vec_set_names(out, names)
}
```

So I don't think we need `name_spec`, we just uphold this predictable invariant

lionel-: Good point.

hadley:
Updated the implementation (and tests) based on your comments. Do you think this looks good enough to finish off `map2_vec()` etc?

lionel- (approved):
LGTM!

Do we want to document this and other `_vec` variants in a different topic marked as experimental?

R/map.R
```diff
@@ -219,6 +222,29 @@ map_raw <- function(.x, .f, ...) {
   .Call(map_impl, environment(), ".x", ".f", "raw")
 }
 
+
+#' @rdname map
+#' @param .ptype If `NULL`, the default, the output type is the common type
+#'   of the elements of the result. Otherwise, supply a "prototype" giving
+#'   the desired type of output.
+#' @importFrom vctrs vec_c vec_size vec_ptype_common
```
DavisVaughan, line 230: It looks like `@import vctrs` is already used?

R/map.R
```diff
@@ -219,6 +222,29 @@ map_raw <- function(.x, .f, ...) {
...
+#' @export
+map_vec <- function(.x, .f, ..., .ptype = NULL) {
+  out <- map(.x, .f, ...)
+
+  .ptype <- vec_ptype_common(!!!out, .ptype = .ptype)
+  bad_sizes <- which(list_sizes(out) != 1L)
+  if (length(bad_sizes) >= 1) {
+    i <- bad_sizes[[1L]]
+    stop_bad_element_vector(out[[i]], i, .ptype, 1L, what = "Result")
+  }
```
DavisVaughan, line 240:
```suggestion
  bad_sizes <- list_sizes(out) != 1L
  if (any(bad_sizes)) {
    i <- which(bad_sizes)[[1L]]
    stop_bad_element_vector(out[[i]], i, .ptype, 1L, what = "Result")
  }
```

Probably a little faster

R/map.R
```diff
@@ -219,6 +222,29 @@ map_raw <- function(.x, .f, ...) {
...
+
+  out <- vec_set_names(out, NULL)
+  out <- vec_unchop(out, ptype = .ptype)
+  vec_set_names(out, names(.x))
```
DavisVaughan, line 244:
```suggestion
  names <- names(out)
  out <- vec_set_names(out, NULL)
  out <- vec_unchop(out, ptype = .ptype)
  vec_set_names(out, names)
```

If you take the names from `out`, that would be a bit forward-thinking for the possibility of `map()` iterating rowwise over data frames, in which case `names(.x)` wouldn't be right

DavisVaughan (approved):

hadley:
This will fail until it uses #942, but the implementation is now very simple/clear.

You might wonder why `map_vec()` gets its own function while `map_c()` does not. I think there are three reasons:

* `map_vec()` is a real map in that `map_vec(x, f)[[i]] == f(x[[i]])`
* We consider simplification to be an internal detail, and its done automatically in `accumulate()` and `list_transpose()`
* We could one day implement this more efficiently, especially when `.ptype` is supplied.

hadley:
Sorry that this has become such a monstrous PR, but you should be able to mostly ignore the changes to the `map2` and `pmap` tests since I felt obliged to update them to at least cover a few of the key invariants.

The most important stuff to hit:

* Are we still confident in `map_vec()` now that the implementation is so simple?
* Have I reasoned through `modify()` and friends correctly?

R/map2.R
```diff
@@ -64,11 +64,16 @@ map2_chr <- function(.x, .y, .f, ..., .progress = NULL) {
   .Call(map2_impl, environment(), ".x", ".y", ".f", "character", .progress)
 }
 
+#' @rdname map2
+#' @export
+map2_vec <- function(.x, .y, .f, ..., .ptype = NULL, .progress = NULL) {
```
lionel-, line 69: Why is `.progress`'s default `NULL` instead of `FALSE` as in the other functions?

R/modify.R
```diff
@@ -93,41 +93,18 @@ modify <- function(.x, .f, ...) {
 modify.default <- function(.x, .f, ...) {
   .f <- as_mapper(.f, ...)
 
-  for (i in seq_along(.x)) {
-    list_slice2(.x, i) <- .f(.x[[i]], ...)
+  if (vec_is_list(.x) || is.data.frame(.x)) {
+    out <- map(vec_proxy(.x), .f, ...)
+    vec_restore(out, .x)
```
lionel-, line 98: For the df case you need to recreate a data frame with the correct number of rows before calling `vec_restore()`.

R/modify.R
```diff
@@ -93,41 +93,18 @@ modify <- function(.x, .f, ...) {
...
+  } else if (vec_is(.x)) {
+    map_vec(.x, .f, ..., .ptype = .x)
+  } else if (is.null(.x) || is.list(.x)) {
```
lionel-, line 101: It's a bit strange to go through that path for `.x = NULL`. Maybe branch into `return(NULL)`?

R/modify.R
```diff
@@ -195,76 +136,59 @@ modify_at <- function(.x, .at, .f, ...) {
...
-modify2_base <- function(mapper, .x, .y, .f, ...) {
   .f <- as_mapper(.f, ...)
-  out <- mapper(.x, .y, .f, ...)
 
-  # if .x got recycled by map2
-  if (length(out) > length(.x)) {
-    .x <- .x[rep(1L, length(out))]
+  if (vec_is_list(.x) || is.data.frame(.x)) {
+    out <- map2(vec_proxy(.x), .y, .f, ...)
+    vec_restore(out, .x)
```
lionel-, line 153: Same here regarding df restoration.

R/modify.R
```diff
@@ -195,76 +136,59 @@ modify_at <- function(.x, .at, .f, ...) {
...
 }
+
+
+# helpers -----------------------------------------------------------------
+
+modify_where <- function(.x, .where, .f, ..., .error_call = caller_env()) {
+  if (vec_is_list(.x) || is.data.frame(.x)) {
+    out <- vec_proxy(.x)
+    out[.where] <- map(out[.where], .f, ...)
+    vec_restore(out, .x)
```
lionel-, line 181: df-restoration

R/pmap.R
```diff
@@ -129,6 +129,16 @@ pmap_chr <- function(.l, .f, ..., .progress = NULL) {
   .Call(pmap_impl, environment(), ".l", ".f", "character", .progress)
 }
 
+#' @export
+#' @rdname pmap
+pmap_vec <- function(.l, .f, ..., .progress = NULL) {
```
lionel-, line 134: Needs `.ptype` arg?

R/modify.R
```diff
@@ -93,41 +93,18 @@ modify <- function(.x, .f, ...) {
...
+  if (vec_is_list(.x) || is.data.frame(.x)) {
+    out <- map(vec_proxy(.x), .f, ...)
+    vec_restore(out, .x)
+  } else if (vec_is(.x)) {
+    map_vec(.x, .f, ..., .ptype = .x)
+  } else if (is.null(.x) || is.list(.x)) {
+    .x[] <- map(.x, .f, ...)
+    .x
+  } else {
+    cli::cli_abort("Don't know how to modify {.obj_type_friendly {.x}}.")
```
lionel-, line 105: Can we use a "can't" form here?

```
! Can't modify a function.
```

Actually a "must" form works well I think:

```
! `.x` must be a vector, not a function.
```

lionel- (approved):

R/modify.R
hadley: Can you suggest a test that illustrates why what I'm currently doing is incorrect, because this currently appears to modify data frames just fine.

R/modify.R
hadley: I had this originally, but it also felt weird — it is nice that `NULL` just falls out as a special case of a zero length list.

R/modify.R
hadley: Do you think I should just ignore the fact it's a generic? It is unlikely that people have provided methods. Do you think I verify that there aren't methods on CRAN then remove the genericity?

R/modify.R
lionel-: It's to support the case of a data frame with 0 columns but positive number of rows.

R/map2.R
DavisVaughan: It seems to happen for all map2 variants. I added that to https://github.com/tidyverse/purrr/issues/947

R/modify.R
DavisVaughan:
``` r
x <- data_frame(.size = 5)
modify(x, identity)
#> data frame with 0 columns and 0 rows
```

R/modify.R
DavisVaughan: FWIW I also prefer the explicit `is.null(x)` branch

R/modify.R
DavisVaughan: I'd probably have a separate data frame branch like this


```r
if (is.data.frame(.x)) {
  size <- vec_size(.x)
  out <- vec_proxy(.x)
  out <- map(out, .f, ...)
  out <- new_data_frame(out, n = size)
  vec_restore(out, .x)
}
```

DavisVaughan: Hmm actually that wouldn't work because then people can't use `modify()` to summarise a data frame like this

```r
df <- data.frame(x = 1:5, y = 6:10)
modify(df, sum)
```

@lionel- I'm not sure we can do this reliably for 0 column data frames. I think the output `size` is going to have to be purely determined by the results of the `.f` calls, and not from the `size` of `.x`. In other words, I think we can only maintain a length invariant, not a size invariant.

R/modify.R
```diff
@@ -195,76 +136,59 @@ modify_at <- function(.x, .at, .f, ...) {
...
   .f <- as_mapper(.f, ...)
-  out <- mapper(.x, .y, .f, ...)
 
-  # if .x got recycled by map2
-  if (length(out) > length(.x)) {
-    .x <- .x[rep(1L, length(out))]
+  if (vec_is_list(.x) || is.data.frame(.x)) {
+    out <- map2(vec_proxy(.x), .y, .f, ...)
+    vec_restore(out, .x)
+  } else if (vec_is(.x)) {
```
DavisVaughan, line 154: This is simultaneously technically correct and also makes me sad that we generate the duplicated column names

``` r
modify2(data.frame(x = 1), 1:3, \(x, y) x)
#>   x x x
#> 1 1 1 1
```

tests/testthat/_snaps/map.md
```diff
@@ -22,3 +22,25 @@
...
       ! Result 2 must have length 1, not 2.
 
+# requires output be length 1 and have common type
+
+    Code
+      map_vec(1:2, ~ rep(1, .x))
+    Condition
+      Error in `map_vec()`:
+      ! All elements must be size 1.
+      i `out[[2]]` is size 2.
```
DavisVaughan, line 32: To confirm, you want the `error_arg` to be `out`?

tests/testthat/test-map2.R
```diff
@@ -1,46 +1,61 @@
...
 
-test_that("map2() recycle names (#779)", {
-  expect_identical(
-    map2(c(a = 1), 1:2, ~ .x),
-    list(a = 1, a = 1)
-  )
+test_that("don't evaluate symbolic objects (#428)", {
+  map2(exprs(1 + 2), NA, ~ expect_identical(.x, quote(1 + 2)))
+  walk2(exprs(1 + 2), NA, ~ expect_identical(.x, quote(1 + 2)))
 })
```
DavisVaughan, line 61: Maybe a `map2_vec()` test about enforcing `.ptype`

tests/testthat/test-pmap.R
```diff
@@ -66,21 +89,7 @@ test_that("preserves S3 class of input vectors (#358)", {
...
-
-test_that("pmap() with empty input copies names", {
-  named_list <- list(named(list()))
-  expect_identical(    pmap(named_list, identity), named(list()))
-  expect_identical(pmap_lgl(named_list, identity), named(lgl()))
-  expect_identical(pmap_int(named_list, identity), named(int()))
-  expect_identical(pmap_dbl(named_list, identity), named(dbl()))
-  expect_identical(pmap_chr(named_list, identity), named(chr()))
+  pwalk(list(exprs(1 + 2)), ~ expect_identical(.x, quote(1 + 2)))
 })
```
DavisVaughan, line 95: Maybe a `pmap_vec()` test about enforcing `.ptype` (which would catch what Lionel pointed out before about missing the ptype arg)

tests/testthat/_snaps/map.md
hadley: Yup

R/modify.R
hadley: Maybe I should just remove the `is.data.frame()` test and it will fail naturally into the `is.list()` branch, which will preserve the existing behaviour.

lionel-: But what should happen with the zero-column case where the size can't be determined from outputs? Should we create a brand new zero-row data frame or should we restore the size? I lean towards the second but I don't know what makes most sense here.

hadley: Ok, not naturally, but I could rearrange the branches to:

```R
  if (is.null(.x)) {
    NULL
  } else if (vec_is_list(.x)) {
    out <- map(vec_proxy(.x), .f, ...)
    vec_restore(out, .x)
  } else if (is.list(.x)) {
    .x[] <- map(.x, .f, ...)
    .x
  } else if (vec_is(.x)) {
    map_vec(.x, .f, ..., .ptype = .x)
  } else {
    cli::cli_abort("Don't know how to modify {.obj_type_friendly {.x}}.")
  }
```

hadley: Behaviour with CRAN purrr:

``` r
df <- data.frame(x = 1:5)
df |> purrr::modify(sum)
#>    x
#> 1 15
#> 2 15
#> 3 15
#> 4 15
#> 5 15
```

<sup>Created on 2022-09-16 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

DavisVaughan: Oh CRAN purrr retains the size of the input when it is a data frame? Nice, then you could do this:

```r
if (is.data.frame(.x)) {
  size <- vec_size(.x)
  out <- vec_proxy(.x)
  out <- map(out, .f, ...)
  out <- vec_recycle_common(!!!out, .size = size)
  out <- new_data_frame(out, n = size)
  vec_restore(out, .x)
}
```

hadley: I just pushed a concrete proposal — I think it's worth explicitly recycling here in order to get a better error message.

DavisVaughan: That fixes all of the cases I brought up:

``` r
x <- data_frame(.size = 5)
modify(x, identity)
#> data frame with 0 columns and 5 rows

df <- data.frame(x = 1:5, y = 6:10)
modify(df, sum)
#>    x  y
#> 1 15 40
#> 2 15 40
#> 3 15 40
#> 4 15 40
#> 5 15 40

modify(
  data.frame(x = 1, y = 2), 
  \(x) if (x == 1) 1:2 else 1
)
#> Error in `modify()`:
#> ! Can't recycle `x` (size 2) to size 1.
```

R/modify.R
hadley: There are five packages with `S3method(modify,` in their namespace (https://cs.github.com/?scopeName=All+repos&scope=&q=org%3Acran+path%3A%2F%5ENAMESPACE%2F+%2FS3method%5C%28modify%2C%2F)

* heemod — own modify generic
* timbr — purrr generic
* purrr — duh
* tfarima — own generic
* yamlet — own generic

timbr's `modify.forest` looks considerably more complex than our implementations, so it's probably ok for it to be its own function: https://github.com/UchidaMizuki/timbr/blob/main/R/purrr.R

R/modify.R
hadley: I'm going to leave it with list because (a) it's not that import and (b) it needs the same recycling behaviour as a zero-length list in `modify2()`.

PR #897, opened 2022-08-28
Check dots in pluck() and chuck()

### LINKED ISSUE #788: `chuck()` and `pluck()` should check for named dots
I accidentally used `chuck()` instead of `pluck()` and was quite confused by the error as I provided a default. It was something along the lines of

``` r
purrr::chuck(list(a = NULL), "a", .default = "a")
#> Error: Plucked object can't be NULL
```

<sup>Created on 2020-08-13 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

Then I realised that I actually wanted to use `pluck()` and `.default` was actually used as an index.

I guess `chuck()` and `pluck()` should use `ellipsis::check_dots_unnamed()`. At least for `chuck()` and `.default` this would make sense.

hadley:
Fixes #788


NEWS.md
```diff
@@ -11,6 +11,8 @@
 
 ## Features and fixes
 
+* `pluck()` and `chuck()` now warn if you provide named inputs to ... (#788).
```
lionel-, line 14:
```suggestion
* `pluck()` and `chuck()` now fail if you provide named inputs to ... (#788).
```

lionel- (approved):

Issue #787, opened 2020-08-13
Better error messages in `chuck()`


mgirlich:
I think the error messages in `chuck` are sometimes a bit confusing and could be more understandable and detailed:

``` r
purrr::chuck(list(a = list(b = 1)), "a", "c")
#> Error: Can't find name `c` in vector

# Proposal:
# Can't find name `c` (index 2) in `.x[["a"]]`
# (i) `.x[["a"]]` has names `b`

purrr::chuck(list(a = 1), "a", 2)
#> Error: Index 2 exceeds the length of plucked object (2 > 1)

# Proposal:
# Can't extract element 2 (index 2) of `.x[["a"]]`
# (i) `.x[["a"]]` has length 1

purrr::chuck(list(a = NULL), "a", 1)
#> Error: Plucked object can't be NULL

# Proposal:
# Can't extract element 1 (index 2) from `.x[["a"]]`
# (i) `.x[["a"]]` is NULL

purrr::chuck(list(a = list()), "a", "b")
#> Error: Plucked object must have at least one element

# Proposal:
# Can't extract name `b` (index 2) from `.x[["a"]]`
# (i) `.x[["a"]]` is empty

purrr::chuck(list(a = list(1, 2)), "a", "b")
#> Error: Index 2 is attempting to pluck from an unnamed vector using a string name

# Proposal:
# Can't extract name `b` (index 2) from `.x[["a"]]`
# (i) `.x[["a"]]` has no names
```

<sup>Created on 2020-08-13 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

hadley:
I agree that the error messages could be improved, and I certainly like your proposals. But these errors messages are all generated in `C`, and I think the work needed to improve them is unlikely to be worth the effort. If you disagree and want to do a PR, let me know and I'll re-open this issue.

PR #898, opened 2022-08-28
Support negative values in pluck()

### LINKED ISSUE #603: Negative indices in pluck()
For indexing from the end.

hadley:
Fixes #603


src/pluck.c
```diff
@@ -296,21 +298,29 @@ static int check_double_index_finiteness(double val, SEXP index, int i, bool str
 }
 
 static int check_double_index_length(double val, int n, int i, bool strict) {
-  if (val < 0) {
+  if (val == 0) {
+    if (strict) {
+      Rf_errorcall(R_NilValue, "Index %d is zero", i + 1);
```
lionel-, line 303: These error messages are missing full stops.

lionel-: It feels like we could do better but I'm not sure how. Maybe:

```
Error in `chuck()`:
! Problem with 3rd subscript.
Cause by error:
! Can't extract at location zero.
```

And

```
Error in `chuck()`:
! Problem with 3rd subscript.
Caused by error:
! Can't extract elements past the end.
i Location 10 doesn't exist, there are only 4 elements.

Error in `chuck()`:
! Problem with 3rd subscript.
Caused by error:
! Can't extract elements past the beginning.
i Negative location -10 doesn't exist, there are only 4 elements.
```

src/pluck.c
```diff
@@ -296,21 +298,29 @@ static int check_double_index_finiteness(double val, SEXP index, int i, bool str
...
+  } else if (val < 0) {
     if (strict) {
+      // Negative values have already been subtracted from end
       Rf_errorcall(R_NilValue,
-                   "Index %d must be greater than 0, not %.0f",
+                   "Negative index %d must be greater than or equal to %d, not %.0f",
                    i + 1,
-                   val + 1);
+                   -n,
+                   -n + val);
```
lionel-, line 314:
```suggestion
                   val - n - 1);
```

lionel- (approved):
Nice

src/pluck.c
hadley: See also #787

src/pluck.c
hadley: Good catch, thanks!

PR #899, opened 2022-08-28
Improve the performance of as_mapper()

### LINKED ISSUE #820: map* functions are slow when `.f` is a string
For example, `map_chr()` is 10x slower than `vapply()`:

```R
library(purrr)
x <- lapply(1:10000, function(name) list(a = "A"))

microbenchmark::microbenchmark(
  map_chr(x, 'a'),
  vapply(x, `[[`, 'a', FUN.VALUE = "")
)
#> Unit: milliseconds
#>                                  expr       min        lq      mean    median        uq      max neval
#>                       map_chr(x, "a") 28.079812 30.615371 31.434516 31.281444 31.856803 42.96536   100
#>  vapply(x, `[[`, "a", FUN.VALUE = "")  3.080842  3.346405  3.826315  3.640201  3.894849 13.50321   100
```

A bit slower would be OK, but this is way slower than the base alternative. For my particular use case, I was surprised to find that a single `map_chr` call was the most expensive part of my code (taking about 78% of the time); after switching to `vapply`, it was much faster.

hadley:
Fixes #820


R/as_mapper.R
```diff
@@ -109,10 +109,11 @@ find_extract_default <- function(.null, .default) {
 
 plucker <- function(i, default) {
   x <- NULL # supress global variables check NOTE
+  i <- as.list(i)
 
   new_function(
```
lionel-, line 114: hmm... Why are we using `new_function()` here?

hadley: I presume it's so we get readable source code?

lionel-: Ah yes probably.

lionel- (approved):
Maybe add a not in `plucker()` about why we're creating a function with metaprogramming?

Also I guess we use `new_function()` instead of `inject(function() ...)` because the latter creates srcrefs.

oh and I'd remove `env = caller_env()` in the `new_function()` call. It raised my eyebrows and I don't think there is a reason for it. The scoping is clearer if the function inherits from the current env, just like it would if we had created it with `inject()`.

PR #901, opened 2022-08-28
Rename vec_depth() to pluck_depth()

### LINKED ISSUE #818: vec_depth fails if list has values of class "name"
Hi Team `purrr`,
Grateful for the work ya'll have done on this package. I've used `purrr` - 974 times since I first became aware of it. Needless to say, it's an essential part of my R toolkit - so thank you!

I'm actually trying to come up with a solution for Advanced R 18.5.3 Q4 and have just used `rapply` to convert an expression to a series of lists with:
`.calls <- rapply(as.list(x), as.list, classes = "call", how = "list")`
I would like to map over the resulting list using `purrr` with something like the following:
`purrr::map_depth(.calls, purrr::vec_depth(.calls), call_match_lgl, pattern = "fn") `

But I'm finding that an element of class `name` apparently causes `vec_depth` to error. 

Here's a simple reprex to illustrate:

```r
purrr::vec_depth(as.list(rlang::call2(mean, 1:10, na.rm = TRUE)))
```

It seems like modifying it like this might make it return results match with what one would expect when visualizing with `lobstr::ast`:
```
vec_depth <- function(x) {
  if (is_null(x)) {
    0L
  } else if (is_atomic(x)) {
    1L
  } else if (rlang::is_symbol(x) || rlang::is_function(x)) {
    1L
  } else if (is_list(x) || rlang::is_call(x)) {
    depths <- map_int(as.list(x), vec_depth)
    1L + max(depths, 0L)
  } else {
    abort("`x` must be a vector")
  }
}
```
There might be some exceptions where this would have unexpected results though. 
Thoughts?

hadley:
And generalise

Fixes #818


R/pluck-depth.R
```diff
@@ -0,0 +1,34 @@
...
+  } else {
+    0L
+  }
+}
+
+#' @export
+#' @rdname pluck_depth
+#' @usage NULL
+vec_depth <- function(x) {
+  lifecycle::deprecate_warn("0.4.0", "vec_depth()", "pluck_depth()")
```
lionel-, line 32: Do we now always skip soft-deprecation?

lionel- (approved):
Mention in NEWS?

PR #902, opened 2022-08-28
Use vctrs instead of dplyr

### LINKED ISSUE #758: Use vctrs instead of dplyr

hadley:
Includes revdep check results. >100 failures 😬 

Fixes #758


hadley:
Too many breakages, so will take a different approach.

Issue #812, opened 2020-12-12
pass dots to anonymous function?


mkoohafkan:
I'm getting some weird behavior when trying to pass `...` arguments to anonymous functions, i.e., functions specified using `~`. I initially ran into this issue using `slider::slide2()` but I am reproducing the issue with `map2()` so I'm thinking the issue has to do with `as_mapper()`. The issue crops up when trying wrapping a function around `map()`, e.g. this reprex:

```r
library(purrr)

# a silly function
test_fun = function(x, mask, a = 5) {
  x[mask] = a
  x
}
# using map2_dbl
bar1 = function(x, mask, ...) {
  map2_dbl(x, mask, test_fun, ...)
}
# using map2_dbl and ~ function
bar2 = function(x, mask, ...) {
  map2_dbl(x, mask, ~ test_fun(.x, .y, ...))
}

set.seed(42)
xs = rnorm(10)
mk = as.logical(rbinom(10, 1, 0.5)) 

bar1(xs, mk, a = 3)
## [1]  3.0000000 -0.5646982  3.0000000  3.0000000  0.4042683  3.0000000
## [7]  1.5115220  3.0000000  2.0184237  3.0000000bar2(xs, mk)

bar2(xs, mk, a = 3)
## Error in test_fun(.x, .y, ...) : unused argument (TRUE)
```

In more complex cases I have gotten weird results where it appears that the components of `...` are getting mixed up or erroneously recycled from the initial two arguments. Is passing `...` to anonymous functions as in `bar2()` simply a no-go?

hadley:
Minimal reprex:

``` r
x <- 1:5

fun1 <- function(x, mask, a = 5) {
  x[mask] <- a
  x
}
fun2 <- purrr::as_mapper(~ fun1(.x, .y, ...))

fun1(x, 1:3, 0)
#> [1] 0 0 0 4 5
fun2(x, 1:3, 0)
#> Error in fun1(.x, .y, ...): unused arguments (1:3, 0)
```

<sup>Created on 2022-08-24 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

The problem is somewhat obvious if you look at `fun2`:

```
> fun2
<lambda>
function (..., .x = ..1, .y = ..2, . = ..1) 
fun1(.x, .y, ...)
attr(,"class")
[1] "rlang_lambda_function" "function" 
```

The problem is that `.x` and `.y` end up passed in twice because they're not removed from `...`

[...]

PR #907, opened 2022-08-30
Improve function documentation

### LINKED ISSUE #887: Adverbs should use `as_function()` not `as_mapper()`
i.e. if `is.function()` or `is.formula()` call `rlang::as_function()`. Otherwise `lifecycle::deprecate_warn()` and call `as_mapper()`

### LINKED ISSUE #905: Consider implications of base lambdas
Can't easily use in examples, but should update text where we promote formulas.

hadley:
Mentioning and using lambda's where appropriate and possible (i.e. not the examples)

Fixes #905. Fixes #887. (This fix improves the docs which was my motivation for filing the issue, and means we don't need to worry about deprecation/backward compatibility).


hadley:
Should finish of #812 in this too

R/adverb-safely.R
```diff
@@ -10,7 +10,11 @@
 #' function (a verb). If you'd like to include a function created an adverb
 #' in a package, be sure to read [faq-adverbs-export].
 #'
-#' @inheritParams map
+#' @param .f A function to modify, specified in one of the following ways:
+#'   * A named function, e.g. `mean`.
+#'   * An anonymous function, e.g. `function(x) x + 1`.
+#'   * A lambda function, e.g. `\(a) a + 1`.
+#'   * A formula, e.g. `~ .x + 1`. Only appropriate for simple tasks.
```
lionel-, line 17: Are we superseding formulas in purrr 1.0? If yes, probably best to omit mentioning them prominently in doc.

lionel-: Though I guess that requires first solving the issue of using base lambdas in example. As long as we use formulas in examples, they need to be documented.

R/keep.R
```diff
@@ -1,21 +1,23 @@
...
+#'   or `FALSE`) specified in one of the following ways:
 #'
-#'   For `compact()`, a function that is applied to each element of
-#'   `.x`. Only those elements where `.p` evaluates to an empty vector
-#'   will be discarded.
+#'   * A named function, e.g. `is.character`.
+#'   * An anonymous function, e.g. `function(x) all(x < 0)`.
+#'   * A lambda function, e.g. `\(x) all(x < 0)`.
+#'   * A formula, e.g. `~ all(.x < 0)`. You must use `.x` to refer to the first
+#'     argument). Only appropriate for very tasks.
```
lionel-, line 20:
```suggestion
#'     argument). Only appropriate for very simple tasks.
```

R/adverb-safely.R
```diff
@@ -10,7 +10,11 @@
 #' function (a verb). If you'd like to include a function created an adverb
 #' in a package, be sure to read [faq-adverbs-export].
 #'
-#' @inheritParams map
+#' @param .f A function to modify, specified in one of the following ways:
+#'   * A named function, e.g. `mean`.
+#'   * An anonymous function, e.g. `function(x) x + 1`.
+#'   * A lambda function, e.g. `\(a) a + 1`.
```
lionel-, line 16: Could we say that `function` is an anonymous function, and `\` is the shorthand for anonymous functions? Or call them both lambdas, but I'm not sure the term "lambda" brings any value.

In any case, making a distinction between anonymous functions and lambda functions might wrongly suggest they are something different rather than two ways of writing the same thing.

lionel-, line 16: Just so that we don't imply that the shorthand variant needs different naming conventions:

```suggestion
#'   * A lambda function, e.g. `\(x) x + 1`.
```

R/adverb-safely.R
lionel-, line 17: What do you mean by simple tasks here?

lionel- (approved):

R/adverb-safely.R
hadley: Yeah, that was exactly my thought process 😄

hadley: I'm trying to imply that you shouldn't expect `...` to work as in #812.

PR #909, opened 2022-09-02
Rework simplification

### LINKED ISSUE #774: Bug: accumulate2 returns a list, not a vector

### LINKED ISSUE #809: accumulate() issues with list output of one row
This is an odd "bug" I've encountered. It is not unexpected given the functions used. So I'm both asking if there is a better way to do this or if the functionality should be updated. If this is beyond the scope of `purrr`, feel free to close.

I wish to accumulate a list of unique values into a cell, but when we accumulate on a single row, it returns a single value. This single value doesn't fit with the other outputs because they are lists. This

``` r
library(purrr)
library(dplyr)

running_unique <- function(x, y) { unique(c(x, y)) }

# Example simple function use
running_unique(1, c(1, 12, 3))
#> [1]  1 12  3

# Example using accumulate
accumulate(1:3, running_unique)
#> [[1]]
#> [1] 1
#> 
#> [[2]]
#> [1] 1 2
#> 
#> [[3]]
#> [1] 1 2 3
accumulate(1, running_unique)
#> [1] 1


df <- tribble(
  ~a, ~b,
  1, 1,
  2, 1,
  2, 2
)
df
#> # A tibble: 3 x 2
#>       a     b
#>   <dbl> <dbl>
#> 1     1     1
#> 2     2     1
#> 3     2     2

df %>%
  group_by(a) %>%
  mutate(accum = accumulate(b, running_unique))
#> Error: Problem with `mutate()` input `accum`.
#> x Input `accum` must return compatible vectors across groups
#> i Input `accum` is `accumulate(b, running_unique)`.
#> i Result type for group 1 (a = 1): <double>.
#> i Result type for group 2 (a = 2): <list>.
```

<sup>Created on 2020-12-04 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

If we just focus on groups with more than one row, then all is well. Notice that even with one value, it works.

```
df %>%
  filter(a == 2) %>%
  group_by(a) %>%
  mutate(accum = accumulate(b, running_unique))
#> # A tibble: 2 x 3
#> # Groups:   a [1]
#>       a     b accum    
#>   <dbl> <dbl> <list>   
#> 1     2     1 <dbl [1]>
#> 2     2     2 <dbl [2]>
```

### LINKED ISSUE #875: Rename `transpose()` to `list_transpose()`
And broadly reconsider its scope and implementation given vctrs and tidyr.

Related issues:

* #397 — doesn't work if names don't align 
* #471 — only works for atomic vectors

### LINKED ISSUE #900: Plan for `flatten()`, `simplify()` and friends
I suggest we at least we mark `as_vector()`, `simplify()`, `simplify_all()`, `flatten()`, and `flatten_*()` as questioning and remove from the function index. The semantics of these functions are not super clear, and I'm not sure it's worth completely nailing them down because many of the problems they were originally designed to solve have moved into tidyr. Their use in their purrr docs reveals few compelling uses cases:

* `as_vector()` is only used in its own examples.

* `simplify()` (a version of `as_vector()` that returns `x` if it can simplify) isn't used at all. 

* `simplify_all()` is used with the output of `transpose()`. The most compelling use case is with `safely()` to get lists of error and results. But I suspect we could tackle this a different way, perhaps with a `list_transpose()` inspired by the work we've done in `unnest_longer()`/`unnest_wider()` or possibly powered by `vec_simplify()`.

* `flatten()` is only used in its own examples. I can see `flatten()` being useful in conjunction with `map()` but it might be easier to create the `flat_map()` functions directly.

Long term, if we do discover that `flatten()` and `simplify()` are useful enough to bother fully fleshing out their semantics, I'd suggest deprecating them and introducing new `list_flatten()` and `list_simplify()`.

Additionally, it might be worth deprecating `flatten_dfc()` and `flatten_dfr()` as they have poor semantics (#648, #757). They are similar to both `dplyr::bind_rows(x)` and `vctrs::vec_rbind(!!!x)`. If we think it's unappealing to recommend a vctrs function we could add `list_rbind <- function(x) vctrs::vec_rbind(!!!x)` (and similar for `list_cbind()`). I think we'd probably require individual elements to be data frames.

hadley:
With this, `map_vec()` will become:

```R
map_vec <- function(.x, .f, ..., .ptype = NULL) {
   out <- unname(map(.x, .f, ...))
   out <- list_simplify(out ,f. simplify = TRUE, ptype = .ptype)
   set_names(out, names(x))
}
```

* Fixes #875 by introducing `list_transpose()`
* Fixes #809 by allowing `accumulate()` to opt out of simplification.
* Fixes #774 by aligning `accumulate()` and `accumulate2()` semantics.
* Fixes #900 by concluding the flatten/simplification re-alignment.


R/list-simplify.R
```diff
@@ -0,0 +1,40 @@
+list_simplify <- function(x, simplify = NA, ptype = NULL) {
```
hadley, line 1: I'm neutral on whether or not we should export this. I can see advantages to keeping it internal, but exporting it would make it easier to deprecate `as_vector()` and `simplify()`

R/list-simplify.R
```diff
@@ -0,0 +1,40 @@
...
+  vec_assert(x, list())
```
lionel-, line 2: I'd create our own assertion here because `vec_assert()` is still experimental and needs more work. For instance this will fail with S3 objects that explictly inherit from `"list"`.

I think we'd disallow data frames and check for `vec_is_list()`?

R/list-simplify.R
```diff
@@ -0,0 +1,40 @@
...
+  if (!is.null(ptype) && isFALSE(simplify)) {
+    abort("Must not specify `ptype` when `simplify = FALSE`")
+  }
+
+  if (isFALSE(simplify)) {
+    return(x)
+  }
+  strict <- !is.na(simplify)
+
+  # We choose not to simply data frames to preserve length invariants
```
lionel-, line 15: I think the size invariants will be preserved for data frames of size 1? I would expect `list_simplify()` to return a data frame of the same size as the input in that case.

lionel- (commented):
Errors are missing full stops.

R/list-simplify.R
hadley: This comes from https://github.com/tidyverse/purrr/pull/782/files#diff-3c7e24ffc0f2dc10f8acf5c629ffa5da0a1479675f3c992d4afa586b4470bb67R515, which you wrote 😄  Maybe I misinterpreted your comment?

lionel-: hmm I think I was confused!

hadley: Now, or then? 😆

R/list-simplify.R
hadley: If we do export, we'll probably need to consider the interface as calling `list_simplify(simplify = FALSE)` would be weird.

R/list-simplify.R
```diff
@@ -0,0 +1,89 @@
...
+#' list_simplify(list(1, 2, 3))
+#'
+#' try(list_simplify(list(1, 2, "x")))
+#' try(list_simplify(list(1, 2, 1:3)))
+list_simplify <- function(x, strict = TRUE, ptype = NULL) {
+  simplify_impl(x, strict = strict, ptype = ptype)
+}
+
+# Wrapper used by purrr functions that do automatic simplification
+list_simplify_internal <- function(
```
lionel-, line 26: Wrong indentation.

R/list-simplify.R
```diff
@@ -0,0 +1,89 @@
...
+    x,
+    simplify = NA,
+    ptype = NULL,
+    error_arg = "x",
+    error_call = caller_env()
+  ) {
+  if (length(simplify) > 1 || !is.logical(simplify)) {
+    cli::cli_abort("{.arg simplify} must be `TRUE`, `FALSE`, or `NA`.")
```
lionel-, line 34:
```suggestion
    cli::cli_abort("{.arg simplify} must be `TRUE`, `FALSE`, or `NA`.", arg = "simplify")
```

R/list-simplify.R
```diff
@@ -0,0 +1,89 @@
...
+  }
+  if (!is.null(ptype) && isFALSE(simplify)) {
+    cli::cli_abort("Must not specify {.arg ptype} when `simplify = FALSE`.")
```
lionel-, line 37: I've always found "must not" strange sounding. In this case I would have reached for "Can't specify" which sounds more natural to me. But that may be because I'm not a native speaker?

R/list-simplify.R
```diff
@@ -0,0 +1,89 @@
...
+  vec_check_list(x, arg = error_arg, call = error_call)
+
+  can_simplify <- every(x, vec_is, size = 1)
+
+  if (can_simplify) {
+    tryCatch(
+      vec_unchop(x, ptype = ptype),
+      vctrs_error_incompatible_type = function(err) {
+        if (strict || !is.null(ptype)) {
+          cli::cli_abort(
```
lionel-, line 69: I'd consider `vec_unchop()` an internal detail so it's probably better to zap its error call:

```r
err$call <- NULL
```

If it had an `error_call` argument, we could zap it there instead, but it hasn't yet.

lionel-: Actually I think that's a case where we'd just take ownership of `vec_unchop()` errors instead of creating a chained error? So we'd do something like:

```r
vec_unchop(x, ptype = ptype, error_call = current_env())
```

R/list-simplify.R
```diff
@@ -0,0 +1,89 @@
...
+          )
+        } else {
+          x
+        }
+      }
+    )
+  } else {
+    if (strict) {
+      cli::cli_abort(
+        "Failed to simplify {error_arg}: not all elements vectors of length 1.",
```
lionel-, line 82:
```suggestion
        "Can't simplify {.arg {error_arg}} because all elements vectors must be length 1.",
```

lionel-: Also need to pass `arg = `

R/list-transpose.R
```diff
@@ -0,0 +1,98 @@
+#' Transpose a list
+#'
+#' @description
+#' `list_transpose()` turns a list-of-lists "inside-out"; it turns a pair of
```
lionel-, line 4: To avoid implying this only works with pairs:

```suggestion
#' `list_transpose()` turns a list-of-lists "inside-out". For instance it turns a pair of
```

R/list-transpose.R
```diff
@@ -0,0 +1,98 @@
...
+#' `b` that contained lists of length n.
+#'
+#' It's called transpose because `x[["a"]][["b"]]` is equivalent to
+#' `transpose(x)[["b"]][["a"]]`, i.e. transposing a list flips the order of
+#' indices in a similar way to transposing a matrix.
+#'
+#' @param x A list of vectors to transpose.
+#' @param template A "template" that specifies the names of output list.
+#'   Usually taken from the name of the first element of `x`.
+#' @param simplify Should the result be simplified?
```
lionel-, line 17:
```suggestion
#' @param simplify Should the result be [simplified][list_simplify]?
```

R/list-transpose.R
```diff
@@ -0,0 +1,98 @@
...
+#' ll %>% list_transpose(c("x", "y", "z"), default = NA)
+list_transpose <- function(x, template = NULL, simplify = NA, ptype = NULL, default = NULL) {
+  vec_check_list(x)
+  if (length(x) == 0) {
+    return(list())
+  }
+
+  template <- template %||%
+    names(x[[1]]) %||%
+    cli::cli_abort("First element of {.arg x} is unnamed, please supply {.arg template}.")
```
lionel-, line 62:
```suggestion
    cli::cli_abort("Must supply either {.arg template} or a named {.arg x}.")
```

R/list-transpose.R
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  out
+}
+
+match_template <- function(x, template, error_arg = caller_arg(x), error_call = caller_env()) {
+  if (is_bare_list(x) && is_named(x)) {
+    extra_names <- setdiff(names(x), template)
+    if (length(extra_names)) {
+      cli::cli_abort(
```
lionel-, line 89: `arg = error_arg`

R/list-transpose.R
lionel-: I don't understand why these fail:

```r
list_transpose(
  list(
    list(1, 2, 3),
    list(4, 5, 6),
    list(7, 8, 9)
  )
)
#> Error in `list_transpose()`:
#> ! First element of `x` is unnamed, please supply `template`.

list_transpose(
  list(
    a = list(1, 2, 3),
    b = list(4, 5, 6),
    c = list(7, 8, 9)
  )
)
#> Error in `list_transpose()`:
#> ! First element of `x` is unnamed, please supply `template`.
```

lionel-: `list_transpose()` no longer round-trips:

```r
x <- list(list(a = 1, b = 2), list(a = 3, b = 4), list(a = 5, b = 6))
x %>% list_transpose()
#> $a
#> [1] 1 3 5
#>
#> $b
#> [1] 2 4 6

x %>% list_transpose() |> list_transpose()
#> Error in `list_transpose()`:
#> ! First element of `x` is unnamed, please supply `template`.
```

R/list-transpose.R
```diff
@@ -0,0 +1,98 @@
...
+#' # Put all the errors and results together
+#' y %>% list_transpose() %>% str()
+#' # Supply a default result to further simplify
+#' y %>% list_transpose(default = list(result = NA)) %>% str()
+#'
+#' # list_transpose() will try to simplify by default:
+#' x <- list(list(a = 1, b = 2), list(a = 3, b = 4), list(a = 5, b = 6))
+#' x %>% list_transpose()
+#' # use simplify = FALSE to always return lists:
+#' x %>% list_transpose(simplify = FALSE) %>% str()
```
lionel-, line 42: Perhaps show what the round-trip with simplification is equivalent to:

```r
x <- list(
  foo = list(a = 1, b = 2),
  bar = list(a = 3, b = 4),
  baz = list(a = 5, b = 6)
)

x %>% list_transpose()
#> $a
#> foo bar baz
#>   1   3   5
#>
#> $b
#> foo bar baz
#>   2   4   6

x %>% list_transpose() |> list_transpose()
#> $foo
#> a b
#> 1 2
#>
#> $bar
#> a b
#> 3 4
#>
#> $baz
#> a b
#> 5 6

# Equivalent to:
map(x, list_simplify)
#> $foo
#> a b
#> 1 2
#>
#> $bar
#> a b
#> 3 4
#>
#> $baz
#> a b
#> 5 6
```

tests/testthat/_snaps/reduce.md
```diff
@@ -1,3 +1,13 @@
+# accumulate() does fail when simpification is required
+
+    Code
+      accumulate(list(1, "a"), ~.y, .simplify = TRUE)
+    Condition
+      Error in `accumulate()`:
+      ! Failed to simplify accumulated results.
+      Caused by error:
+      ! Can't combine `..1` <double> and `..2` <character>.
```
lionel-, line 9: Not sure we need chained errors here.

tests/testthat/test-list-transpose.R
```diff
@@ -0,0 +1,72 @@
...
+  expect_snapshot(list_transpose(x, simplify = list(c = FALSE)), error = TRUE)
+})
+
+test_that("can supply `ptype` globally or individually", {
+  x <- list(list(a = 1, b = 2), list(a = 3, b = 4))
+  expect_identical(
+    list_transpose(x, ptype = integer()),
+    list(a = c(1L, 3L), b = c(2L, 4L))
+  )
+  expect_equal(
```
lionel-, line 46:
```suggestion
  expect_identical(
```

R/list-simplify.R
hadley: Yeah, I couldn't quite figure out how to get all the errors to be consistent here. I'll take another pass.

R/list-simplify.R
hadley: I thought this was the double-indent style we reached agreement on in https://style.tidyverse.org/functions.html. I have mentally adjusted myself to like this style 😆/😭

R/list-transpose.R
hadley: I made it stricter because naming seems like a good principle, but it does seem undesirable to break the symmetry, so I'll bring that back.

R/list-simplify.R
lionel-: I think I may be getting used to it too. But I thought we'd only use it when the lines get too long?

hadley: I think maybe we didn't mean the same thing by "lines that are too long", because it seems to me like it applies here because I can't otherwise fit the function definition on a single line.

lionel-: 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱

R/list-simplify.R
```diff
@@ -0,0 +1,77 @@
...
+}
+
+# Wrapper used by purrr functions that do automatic simplification
+list_simplify_internal <- function(x,
+                                   simplify = NA,
+                                   ptype = NULL,
+                                   error_call = caller_env()
+  ) {
+  if (length(simplify) > 1 || !is.logical(simplify)) {
+    cli::cli_abort("{.arg simplify} must be `TRUE`, `FALSE`, or `NA`.", arg = "simplify")
```
DavisVaughan, line 32: Pass `error_call` through

R/list-simplify.R
```diff
@@ -0,0 +1,77 @@
...
+  }
+  if (!is.null(ptype) && isFALSE(simplify)) {
+    cli::cli_abort("Can't specify {.arg ptype} when `simplify = FALSE`.")
```
DavisVaughan, line 35: Pass `error_call` through

DavisVaughan: Should `arg = "ptype"` be set?

tests/testthat/test-reduce.R
```diff
@@ -162,8 +152,8 @@ test_that("basic accumulate2() works", {
   paste2 <- function(x, y, sep) paste(x, y, sep = sep)
 
   x <- c("a", "b", "c")
-  expect_equal(accumulate2(x, c("-", "."), paste2), list("a", "a-b", "a-b.c"))
-  expect_equal(accumulate2(x, c(".", "-", "."), paste2, .init = "x"), list("x", "x.a", "x.a-b", "x.a-b.c"))
+  expect_equal(accumulate2(x, c("-", "."), paste2), c("a", "a-b", "a-b.c"))
```
DavisVaughan, line 155: I wonder if the switch to auto simplification will break much code

NEWS.md
```diff
@@ -37,6 +37,13 @@
 * `*_dfc()` and `*_dfr()` have been deprecated in favour of using the 
   appropriate map function along with `list_rbind()` or `list_cbind()` (#912).
 
+* `simplify()`, `simplify_all()`, and `as_vector()` have been deprecated in
+  favour of `list_simplify()`. It provides a more consistent definition of 
+  simplification (#900).
+
+* `transpose()` has been deprecated in favour of `list_transpose()` (#875).
+  It has built-in simplification.
```
DavisVaughan, line 45: I think my biggest worry is that `list_transpose()` is probably a lot slower than `transpose()`, but I guess it is way more featureful and generic

R/list-simplify.R
```diff
@@ -0,0 +1,77 @@
...
+list_simplify <- function(x, strict = TRUE, ptype = NULL) {
+  simplify_impl(x, strict = strict, ptype = ptype)
+}
+
+# Wrapper used by purrr functions that do automatic simplification
+list_simplify_internal <- function(x,
+                                   simplify = NA,
+                                   ptype = NULL,
+                                   error_call = caller_env()
+  ) {
```
DavisVaughan, line 30:
```suggestion
                                   error_call = caller_env()) {
```

R/list-simplify.R
```diff
@@ -0,0 +1,77 @@
...
+  )
+}
+
+simplify_impl <- function(x,
+                          strict = TRUE,
+                          ptype = NULL,
+                          error_call = caller_env()) {
+  vec_check_list(x, call = error_call)
+
+  can_simplify <- every(x, vec_is, size = 1)
```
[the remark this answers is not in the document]
DavisVaughan: Even if you want to support scalar types, it might be worth using the approach above and wrapping it in `tryCatch()` that returns `FALSE` on error, as it will still probably be an order of magnitude or more faster than `vec_is()`

DavisVaughan (approved):

R/list-transpose.R
```diff
@@ -0,0 +1,127 @@
...
+#' example, if you had a list of length `n` where each component had values `a`
+#' and `b`, `list_transpose()` would make a list with elements `a` and
+#' `b` that contained lists of length n.
+#'
+#' It's called transpose because `x[["a"]][["b"]]` is equivalent to
+#' `transpose(x)[["b"]][["a"]]`, i.e. transposing a list flips the order of
+#' indices in a similar way to transposing a matrix.
+#'
+#' @param x A list of vectors to transpose.
+#' @param template A "template" that specifies the names of output list.
```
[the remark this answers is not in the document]
hadley: Oops forgot to update these docs. Hopefully it's more obvious why it's called `template` and not names when you realise it can also take positions.

R/list-transpose.R
```diff
@@ -0,0 +1,127 @@
...
+#'   * `TRUE`: simplify or die trying.
+#'   * `NA`: simplify if possible.
+#'   * `FALSE`: never try to simplify, always leaving as a list.
+#'
+#'   Alternatively, a named list specifying the simplification by output column.
+#' @param ptype An optional vector prototype used to control the simplification.
+#'   Alternatively, a named list specifying the prototype by output column.
+#' @param default A default value to use if a value is absent of `NULL`.
+#'   Alternatively, a named list specifying the prototype by output column.
+#' @export
```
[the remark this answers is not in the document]
hadley: Now: Length of `default` (1) and `template` (2) must be the same when transposing by position.

R/list-transpose.R
```diff
@@ -0,0 +1,127 @@
...
+#'
+#' # And specify default if you want to simplify
+#' ll %>% list_transpose(c("x", "y", "z"), default = NA)
+list_transpose <- function(x, template = NULL, simplify = NA, ptype = NULL, default = NULL) {
+  vec_check_list(x)
+  if (length(x) == 0) {
+    return(list())
+  }
+
+  template <- template %||% vec_index(x[[1]])
```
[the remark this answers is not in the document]
hadley: I think this is a bit simpler:

```
  if (length(x) == 0) {
    template <- integer()
  } else {
    template <- template %||% vec_index(x[[1]])
  }
```

PR #912, opened 2022-09-04
Rework flattening

### LINKED ISSUE #376: map_dfc() fails when the result contains list columns

### LINKED ISSUE #405: Bring back flatmap
as `map_flat_type`.

### LINKED ISSUE #472: map_dfr() column binds vectors

### LINKED ISSUE #575: Is flattening c + splicing?
i.e. should `flatten(x, .type = foo)` be equivalent to `vec_c(!!!x, .type = foo)` or should it be equivalent to `simplify()`?

### LINKED ISSUE #757: Reconsider flatten_dfc() and flatten_dfr()
It's concerning that they both return the same results in this case:

``` r
library(purrr)

dfs <- list(c(a = 1), c(b = 2))
flatten_dfc(dfs)
#> # A tibble: 1 x 2
#>       a     b
#>   <dbl> <dbl>
#> 1     1     2
flatten_dfr(dfs)
#> # A tibble: 1 x 2
#>       a     b
#>   <dbl> <dbl>
#> 1     1     2
```

<sup>Created on 2020-04-16 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

### LINKED ISSUE #758: Use vctrs instead of dplyr

hadley:
This is a large PR that reworks flattening, removing inconsistencies and switching to vctrs as a backend. 

The key idea is to introduce a new family of "combining" functions: `list_c()`, `list_rbind()`, and `list_cbind()`, which replace `flatten_lgl()`, `flatten_int()`, `flatten_dbl()`, `flatten_chr()` (now `list_c()`), `flatten_dfc()` (`list_cbind()`), and `flatten_dfr()` (`list_rbind()`). The new functions are straightforward wrappers around vctrs functions, but somehow feel natural in purrr to me. 

This leaves `flatten()`, which had a rather idiosyncratic interface. It's now been replaced by `list_flatten()` which now always removes a single layer of list hierarchy (and nothing else). While working on this I realised that this was actually what `splice()` did, so overall this feels like a major improvement in naming consistency.

With those functions in place we can deprecate `map_dfr()` and `map_dfc()` which are actually "flat" map functions because they combine, rather than simplify, the results. They have never actually belonged with `map_int()` and friends because they don't have the restriction that `.f` needs to return a length-1 results. This also strongly implies that `flat_map()` would just be `map_c()` and is thus not necessary.

* Fixes #376 by deprecating `map_dfc()`
* Fixes #405 by clearly ruling against `map_c()`
* Fixes #472 by deprecating `map_dfr()`
* Fixes #575 by introducing `list_c()`, `list_rbind()`, and `list_cbind()`
* Fixes #757 by deprecating `flatten_dfr()` and `flatten_dfc()`
* Fixes #758 by introducing `list_rbind()` and `list_cbind()`
* Part of #900

----

Updated to reflect that we no longer believe that `_cbind()`, `_rbind()` and `_c()` are necessary as they're very lightweight wrappers.


R/list-combine.R
```diff
@@ -0,0 +1,68 @@
...
+  vctrs::vec_unchop(x, ptype = ptype)
+}
+
+#' @export
+#' @rdname list_c
+list_cbind <- function(
+    x,
+    name_repair = c("unique", "universal", "check_unique"),
+    size = NULL
+  ) {
```
DavisVaughan, line 49: I really can't get on board with this new style 😬

R/list-combine.R
```diff
@@ -0,0 +1,68 @@
...
+  check_is_list(x)
+  vctrs::vec_cbind(!!!x, .name_repair = name_repair, .size = size)
+}
+
+#' @export
+#' @rdname list_c
+list_rbind <- function(x, id = rlang::zap(), ptype = NULL) {
+  check_is_list(x)
+  vctrs::vec_rbind(!!!x, .names_to = id, .ptype = ptype)
```
DavisVaughan, line 58: There are a few other `vec_unchop/rbind/cbind()` args that may be useful to pass through?

DavisVaughan, line 58: dev `vec_cbind/rbind()` have `.call` args now too, but they default to `current_env()` so you'd need to pass `.call = current_env()` manually

R/list-combine.R
```diff
@@ -0,0 +1,68 @@
...
+#' x2 <- list(
+#'   a = data.frame(x = 1:2),
+#'   b = data.frame(y = "a")
+#' )
+#' list_rbind(x2)
+#' list_rbind(x2, id = "id")
+#' list_rbind(unname(x2), id = "id")
+#'
+#' list_cbind(x2)
+list_c <- function(x, ptype = NULL) {
```
DavisVaughan, line 38: I think it would be useful to use `...` in all 3 of these to force named arguments. Especially since these really do seem like optional and rarely used args, and we might add more args in the future if the underlying vctrs functions evolve further

R/list-combine.R
```diff
@@ -0,0 +1,68 @@
...
+}
+
+#' @export
+#' @rdname list_c
+list_rbind <- function(x, id = rlang::zap(), ptype = NULL) {
+  check_is_list(x)
+  vctrs::vec_rbind(!!!x, .names_to = id, .ptype = ptype)
+}
+
+check_is_list <- function(x, error_call = caller_env()) {
```
DavisVaughan, line 61: `vctrs::vec_check_list()`?

R/list-flatten.R
```diff
@@ -0,0 +1,42 @@
+#' Flatten a list
+#'
+#' Flattening a list removes a single layer of internal hierarchy.
```
DavisVaughan, line 3: I would mention something about non-list elements being skipped over.

Part of me wants `list_flatten()` to require that each element of `x` actually be a list, rather than "skipping" non-lists. But I guess that is too strict?

R/list-flatten.R
```diff
@@ -0,0 +1,42 @@
...
+#' # Use name_spec to control how inner and outer names are combined
+#' x <- list(x = list(a = 1, b = 2), y = list(c = 1, d = 2))
+#' x %>% list_flatten() %>% names()
+#' x %>% list_flatten(name_spec = "{outer}") %>% names()
+#' x %>% list_flatten(name_spec = "{inner}") %>% names()
+list_flatten <- function(x, name_spec = "{outer}_{inner}") {
+  check_is_list(x)
+
+  x <- modify_if(x, vctrs::vec_is_list, identity, .else = list)
+  vec_unchop(x, ptype = list(), name_spec = name_spec)
```
DavisVaughan, line 41: Maybe pass `name_repair` arg through

R/list-flatten.R
```diff
@@ -0,0 +1,42 @@
...
+#'
+#' # Use name_spec to control how inner and outer names are combined
+#' x <- list(x = list(a = 1, b = 2), y = list(c = 1, d = 2))
+#' x %>% list_flatten() %>% names()
+#' x %>% list_flatten(name_spec = "{outer}") %>% names()
+#' x %>% list_flatten(name_spec = "{inner}") %>% names()
+list_flatten <- function(x, name_spec = "{outer}_{inner}") {
+  check_is_list(x)
+
+  x <- modify_if(x, vctrs::vec_is_list, identity, .else = list)
```
DavisVaughan, line 40:
```suggestion
  x <- modify_if(x, vec_is_list, identity, .else = list)
```

R/map-df.R
```diff
@@ -0,0 +1,152 @@
+#' Functions that return data frames
+#'
+#' @description
+#' `r lifecycle::badge("deprecated")`
+#'
+#' These variants of [map()], [map2()], [imap()], and [pmap()] return data
+#' frames. They have been deprecated because they use `dplyr::bind_rows()`
+#' and `dplyr::bind_cols()` which have often confusing semantics, and their
```
DavisVaughan, line 8:
```suggestion
#' and `dplyr::bind_cols()` which often have confusing semantics, and their
```

R/map-df.R
```diff
@@ -0,0 +1,152 @@
...
+#' names are suboptimal because they suggest they work like `_lgl()`, `_int()`,
+#' and friends which require length 1 outputs, but actually they accept results
+#' of anything because the results are combined to together.
```
DavisVaughan, line 11:
```suggestion
#' and friends which require length 1 outputs, but actually they return results
#' of any size because the results are combined together without any size checks.
```

R/map-df.R
```diff
@@ -0,0 +1,152 @@
...
+#'
+#' You can now instead use functions with `_rbind()` and `_cbind()` suffixes
+#' which use `vctrs::vec_rbind()` and `vctrs::vec_cbind()` under the hood,
+#' and have names that more clearly reflect their semantics.
```
DavisVaughan, line 15: Maybe link to them? i.e. `[list_rbind()]`

tests/testthat/test-list-combine.R
```diff
@@ -0,0 +1,59 @@
...
+  expect_snapshot(error = TRUE, {
+    list_rbind(list(df1), ptype = data.frame(x = character()))
+  })
+})
+
+test_that("assert input is a list", {
+  expect_snapshot(error = TRUE, {
+    list_c(1)
+    list_rbind(1)
+    list_cbind(1)
```
DavisVaughan, line 57: Maybe worth a test that data frames aren't allowed here

NAMESPACE
```diff
@@ -129,8 +129,12 @@ export(lift_lv)
 export(lift_vd)
```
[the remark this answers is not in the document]
DavisVaughan: I think the other reason I think these are separate is because `map_c()` physically can't be faster than sequential calls to `map() + list_c()`, because you need all of the results from the `map()` to be able to compute the output size that we'd get from `list_c()`, and to me that suggests they shouldn't be merged together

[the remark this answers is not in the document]
hadley: Yeah, agreed. I think the `length(x) == length(map*(x, f))` invariant is key.

`x |> map_vec(f)` feels a bit like this because it'll begin by being implemented as `x |> map(f) |> list_simplify()` but I think that's more of a current implementation detail. Especially if you do supply a `ptype` we could potentially implement `map_vec()` much more efficiently.

R/list-combine.R
```diff
@@ -0,0 +1,68 @@
...
+#'   b = data.frame(y = "a")
+#' )
+#' list_rbind(x2)
+#' list_rbind(x2, id = "id")
+#' list_rbind(unname(x2), id = "id")
+#'
+#' list_cbind(x2)
+list_c <- function(x, ptype = NULL) {
+  check_is_list(x)
+  vctrs::vec_unchop(x, ptype = ptype)
```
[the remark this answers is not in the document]
hadley: I've left it for now for consistency with the others.

R/list-combine.R
hadley: Did you have any specifics in mind? I thought I got all the useful ones.

R/list-flatten.R
lionel-: > Part of me wants list_flatten() to require that each element of x actually be a list, rather than "skipping" non-lists.

That would be non-standard and make it hard to use with actual trees.

R/list-combine.R
```diff
@@ -0,0 +1,59 @@
...
+#'   together with [vctrs::vec_rbind()].
+#'
+#' * `list_cbind()` combines elements into a data frame by column-binding them
+#'   together with [vctrs::vec_cbind()].
+#'
+#' @param x A list.
+#' @param ptype An optional prototype to ensure that the output type is always
+#'   the same.
+#' @param id By default, `names(x)` are lost. Alternatively, supply a string
+#'   and the names will be saved into a column with name `id`. If `id`
```
lionel-, line 17: This makes it sound like the column name is always named `id`. Perhaps:

> Alternatively, supply `id` with a column name into which the names will be saved.

R/list-combine.R
```diff
@@ -0,0 +1,59 @@
...
+  vec_check_list(x)
+  vctrs::vec_cbind(!!!x, .name_repair = name_repair, .size = size, .call = current_env())
+}
+
+#' @export
+#' @rdname list_c
+list_rbind <- function(x, id = rlang::zap(), ptype = NULL) {
+  vec_check_list(x)
+  vctrs::vec_rbind(!!!x, .names_to = id, .ptype = ptype, .call = current_env())
+}
```
lionel-, line 59: Do we want to require each element to be a data frame, or do we embrace the way `vec_rbind()` and `vec_cbind()` convert vectors to data frames?

R/list-flatten.R
```diff
@@ -0,0 +1,55 @@
...
+#' x %>% list_flatten(name_spec = "{outer}") %>% names()
+#' x %>% list_flatten(name_spec = "{inner}") %>% names()
+list_flatten <- function(
+    x,
+    name_spec = "{outer}_{inner}",
+    name_repair = c("minimal", "unique", "check_unique", "universal")
+  ) {
+  vec_check_list(x)
+
+  x <- modify_if(x, vec_is_list, identity, .else = list)
```
lionel-, line 48: Why not `map_if()`?

R/list-flatten.R
```diff
@@ -0,0 +1,55 @@
...
+  vec_unchop(
+    x,
+    ptype = list(),
```
lionel-, line 51: I'm thinking this should be `ptype = x` (the original `x`) to preserve the input type.

R/map-df.R
```diff
@@ -0,0 +1,153 @@
...
+
+  .f <- as_mapper(.f, ...)
+  res <- map(.x, .f, ...)
+  dplyr::bind_rows(res, .id = .id)
+}
+
+#' @rdname map_dfr
+#' @export
+map_dfc <- function(.x, .f, ...) {
+  lifecycle::deprecate_warn("0.4.0", "map_dfc()", "map_cbind()")
```
lionel-, line 82: Replaced by `map()` + `list_cbind()`

R/map-df.R
```diff
@@ -0,0 +1,153 @@
...
+  .f <- as_mapper(.f, ...)
+  res <- map(.x, .f, ...)
+  dplyr::bind_rows(res, .id = .id)
+}
+
+#' @rdname map_dfr
+#' @usage NULL
+#' @export
+map_df <- function(.x, .f, ..., .id = NULL) {
+  lifecycle::deprecate_warn("0.4.0", "map_df()", "map_rbind()")
```
lionel-, line 71: Replaced by `map()` + `list_rbind()`

R/map-if-at.R
```diff
@@ -0,0 +1,143 @@
...
+
+  if (.depth == 0) {
+    return(.f(.x, ...))
+  }
+
+  if (.depth == 1) {
+    return(map(.x, .f, ...))
+  }
+
+  # Should this be replaced with a generic way of figuring out atomic
```
lionel-, line 136: Can we now use `vec_is_list()`?

R/list-combine.R
hadley: I did what the [style guide](https://style.tidyverse.org/functions.html#long-lines-1) says 😄

R/list-combine.R
hadley: I'll use `{id}` since that's a convention we use elsewhere and should be familiar to many readers.

R/list-combine.R
hadley: Let's try requiring data frames.

R/list-flatten.R
hadley: Can this ever be something than a list though? `x` can't be a data frame, and it seems unlikely to be a `list_of` unless it's a `list_of<list>`?

R/list-combine.R
lionel-: I think my suggestion conforms to the style guide? It's not great but so much better than completely changing the overall shape of function definitions...

R/map-if-at.R
hadley: Yeah, tracking in https://github.com/tidyverse/purrr/issues/920

R/list-flatten.R
lionel-: oh it has to be a list, which we check with `vec_check_list()`. But it could be some kind of vctrs list. I agree that `list_of` is unlikely until we start supporting more complex type definitions.

I think my concern is that it might make sense to preserve types in the future if we have more vctrs lists, but it will be difficult to change then, so better do it now.

hadley: Can we assume a `list()` can always be coerced to `x`, even if `x` is a subtype of list? I don't think we can, so this seems risky? 

But I tried it anyway, and then I got:

```R
list_flatten(list_of(list(1, 2, 3), list(4), list(4)))
#> Error:
#> ! Can't convert <list> to <list_of<list>>.
```

So I'm going to leave as is for this PR and we can revisit in the future.

lionel-: ah yes. I think the proper way would be to initialise the ptype of the input to the full output size, then assign elements into it.

lionel- (approved):
🎉

R/list-combine.R
DavisVaughan: I like @lionel-'s suggestion more than what is currently there

R/list-combine.R
DavisVaughan: We have this well meaning but not great error in the case where both an outer and inner name exist, supplying a `name_spec` is required to fix it

```r
list_c(list(x = c(y = 1)))
#> Error:
#> ! Can't merge the outer name `x` with a vector of length > 1.
#> Please supply a `.name_spec` specification.
```

DavisVaughan: rbind equivalent

```r
list_rbind(
  list(x = data.frame(y = 1, row.names = "row")), 
  id = NULL
)
#> Error:
#> ! Can't merge the outer name `x` with a vector of length > 1.
#> Please supply a `.name_spec` specification.
```

I think cbind is ok

R/list-flatten.R
DavisVaughan: Sounds good to me, just a thought

R/list-combine.R
DavisVaughan: @hadley I guess you disagree? I'm just going off previous painful experience adding new args to functions without dots

R/list-combine.R
```diff
@@ -0,0 +1,79 @@
...
+#' * `list_c()` combines elements into a vector by concatenating them together
+#'   with [vctrs::vec_c()].
+#'
+#' * `list_rbind()` combines elements into a data frame by row-binding them
+#'   together with [vctrs::vec_rbind()].
+#'
+#' * `list_cbind()` combines elements into a data frame by column-binding them
+#'   together with [vctrs::vec_cbind()].
+#'
+#' @param x A list.  For `list_rbind()` and `list_cbind()` the list must
```
DavisVaughan, line 13:
```suggestion
#' @param x A list. For `list_rbind()` and `list_cbind()` the list must
```

DavisVaughan (approved):

R/reduce.R
```diff
@@ -447,7 +447,7 @@ seq_len2 <- function(start, end) {
 #' rerun(5, rnorm(100)) %>%
 #'   set_names(paste0("sim", 1:5)) %>%
 #'   map(~ accumulate(., ~ .05 + .x + .y)) %>%
-#'   map_dfr(~ tibble(value = .x, step = 1:100), .id = "simulation") %>%
+#'   map_rbind(~ tibble(value = .x, step = 1:100), .id = "simulation") %>%
```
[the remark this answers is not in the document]
hadley: Good catch. IMO it reads better now that `list_rbind()` is a separate step.

R/list-combine.R
hadley: Fixed by hand.

PR #914, opened 2022-09-06
Use vctrs coercions

### LINKED ISSUE #904: Match vctrs coercion rules for simple types
``` r
library(purrr)

# Should succeed
map_lgl(0, identity)
#> Error: Can't coerce element 1 from a double to a logical
map_lgl(1, identity)
#> Error: Can't coerce element 1 from a double to a logical

map_int(1, identity)
#> Error: Can't coerce element 1 from a double to a integer

# Should error
map_chr(TRUE, identity)
#> [1] "TRUE"
map_chr(1L, identity)
#> [1] "1"
map_chr(1, identity)
#> [1] "1.000000"
```

<sup>Created on 2022-08-29 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

Requires a set of (relatively straightforward) changes to `set_vector_value()`

May need to deprecate if this breaks too many revdeps

hadley:
So far have just removed coercions that vctrs doesn't support, which breaks ~15 packages.

Fixes #904


hadley:
I've now implemented the remaining pieces. See news for details.

NEWS.md
```diff
@@ -38,6 +38,21 @@
...
+* `_lgl()`, `_int()`, `_int()`, and `_dbl()` now use the same (strict) coercion
+  methods as vctrs (#904). This means that:
+  
+    * `map_chr(TRUE, identity)`, `map_chr(0L, identity)`, and
+      `map_chr(1L, identity)` now fail because converting a 
+      logical/integer/double to a character vector should require an explicit
+      coercion.
+      
+    * `map_dbl(1.5, identity)` now fails because you shouldn't silently truncate.
+      But note that `map_dbl(1, identity)` since no numeric precision is lost.
```
lionel-, line 50:
```suggestion
    * `map_int(1.5, identity)` now fails because you shouldn't silently truncate.
      But note that `map_int(1, identity)` since no numeric precision is lost.
```

lionel-: "shouldn't" seems to make assumptions here. Maybe:

> now fails to protect the caller from unexpected truncations

lionel- (commented):
We would switch to `vec_cast()` in a second step right?

I think we may need to provide an edition or a legacy global option to keep supporting existing scripts that these changes will break. Maybe we should keep the existing code as is and provide a vctrs default path?

NEWS.md
hadley: Oh yeah I was trying to make a general statement, i.e. "it's a bad idea to truncate". I'll reword.

hadley:
For now, should deprecate character coercions rather than eliminating them all together.

src/coerce.c
```diff
@@ -18,21 +54,48 @@ double logical_to_real(int x) {
...
+  UNPROTECT(4);
 }
-SEXP integer_to_char(int x) {
+
+SEXP logical_to_char(int x, SEXP from, SEXP to, int i) {
+  if (x == NA_LOGICAL) {
+    return NA_STRING;
+  } else {
+    if (i == 0)
+      deprecate_to_char("logical");
```
hadley, line 79: @lionel- This is a bit of a hack because I want to deprecate aggressively so `deprecate_to_char()` uses `always = TRUE` and then I only call it for the first element. There's some small risk that the deprecation message won't get generated if the transformation is not type stable (e.g. `map_chr(list("x", 1), identity)`) but I think it's worth taking that small risk is return for much simpler code.

NEWS.md
```diff
@@ -38,6 +38,22 @@
...
+  methods as vctrs (#904). This means that:
+  
+    * `map_chr(TRUE, identity)`, `map_chr(0L, identity)`, and 
+      `map_chr(1L, identity)` are deprecated because we now believe that 
+      converting a logical/integer/double to a character vector should require 
+      an explicit coercion.
+      
+    * `map_dbl(1.5, identity)` now fails because we believe that silently 
+      truncating doubles to integers is dangerous. But note that 
+      `map_dbl(1, identity)` still works since no numeric precision is lost.
```
DavisVaughan, line 51: Isn't this `map_int()` in both examples here?

R/coerce.R
```diff
@@ -8,4 +8,13 @@ coerce_lgl <- function(x) coerce(x, "logical")
 coerce_int <- function(x) coerce(x, "integer")
 coerce_dbl <- function(x) coerce(x, "double")
 coerce_chr <- function(x) coerce(x, "character")
-coerce_raw <- function(x) coerce(x, "raw")
+
+
+deprecate_to_char <- function(type) {
+  lifecycle::deprecate_warn(
+    "0.4.0",
+    I(paste0("Automatical coercion from ", type, " to character")),
```
DavisVaughan, line 16: `Automatic` not `Automatical`?

NEWS.md
hadley: Ooops yes.

PR #917, opened 2022-09-06
Use tidyverse recycling rules

### LINKED ISSUE #878: Ensure that recycling uses tidyverse rules
i.e. only recycle vectors of length 1. 

Problems include at least:

``` r
purrr:::recycle_args(list(1, integer(0)))
#> Error in purrr:::recycle_args(list(1, integer(0))): all(lengths == 1L | lengths == n) is not TRUE
purrr::map2(integer(0), 1:2, list)
#> list()
purrr::pmap(list(integer(0), 1:2), list)
#> list()
```

<sup>Created on 2022-08-24 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

hadley:
Fixes #878


src/map.c
```diff
@@ -111,21 +111,15 @@ SEXP map2_impl(SEXP env, SEXP x_name_, SEXP y_name_, SEXP f_name_, SEXP type_) {
...
 
   int nx = Rf_length(x_val), ny = Rf_length(y_val);
-  if (nx == 0 || ny == 0) {
-    SEXP out = PROTECT(Rf_allocVector(type, 0));
-    copy_names(x_val, out);
-    UNPROTECT(3);
-    return out;
-  }
-  if (nx != ny && !(nx == 1 || ny == 1)) {
+  if (nx != ny && nx != 1 && ny != 1) {
```
lionel-, line 114: Can we unclass data frames and use `vec_recycle_common()` instead?

lionel- (approved):

src/map.c
hadley: I thought about that too, but I worried (possibly needlessly) about the performance implications of now `rep`-ing scalars into large vectors instead. And the set of changes seemed small enough.

lionel-: I think it's fine merging as is. Just noting that in the future the rep'd scalars would become altrep repetitions so the problem would go away eventually.

PR #919, opened 2022-09-07
Re-implement assign_in()

### LINKED ISSUE #634: Should `pluck<-` actually be `chuck<-`?
I am not sure that `pluck()` and `pluck<-()` are consistent. `pluck()` returns `NULL` if the element does not exist, but `pluck<-()` throws an error. In that way, `pluck<-()` feels more like the counterpart to `chuck()`. 

I am not sure that `pluck<-()` should even exist, since `assign_in()` is very strict and requires that the plucked-location must exist. I am just arguing that is should be named `chuck<-()` instead.

``` r
library(purrr)

pluck(mtcars, "y")
#> NULL

chuck(mtcars, "y")
#> Error: Can't find name `y` in vector

# should this be `chuck<-`()?
pluck(mtcars, "y") <- 1 
#> Error: Can't find name `y` in vector

# this should be `pluck<-`() I think
mtcars$y <- 1

# what I think should be `chuck<-`() uses chuck() already
`pluck<-`
#> function (.x, ..., value) 
#> {
#>     assign_in(.x, list2(...), value)
#> }
#> <bytecode: 0x7f9558ae34e8>
#> <environment: namespace:purrr>
purrr:::assign_in
#> function (x, where, value) 
#> {
#>     chuck(x, !!!where)
#>     call <- reduce_subset_call(quote(x), as.list(where))
#>     call <- call("<-", call, value)
#>     eval_bare(call)
#>     x
#> }
#> <bytecode: 0x7f9558ae2790>
#> <environment: namespace:purrr>
```

<sup>Created on 2019-02-04 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1.9000)</sup>

### LINKED ISSUE #636: `assign_in()` `NULL` should assign rather than remove
The modfied object returned by `pluck<-` should always have the same length structure.

### LINKED ISSUE #704: `assign_in` and `modify_in` should create nested elements
Current implementation of `assign_in`, `update_in` and `pluck<-` doesn't create new elements. 
This is in contrast to `[[<-` semantics and most importantly to the parallel functional tooling in other languages ([clojure](https://clojuredocs.org/clojure.core/assoc-in), [python](https://toolz.readthedocs.io/en/latest/api.html#toolz.dicttoolz.update_in)).

``` r
assign_in(list(), "aa", 34)
#> Error: Plucked object must have at least one element
pluck(list(), "aa") <- 34
#> Error in pluck(list(), "aa") <- 34: invalid (NULL) left side of assignment
```

Moreover, this is inconsistent with `pluck` itself which returns `NULL` on missing elements.  So one would expect no error for missing elements in `pluck<-` as well.

Seems related: #636

hadley:
* Can assign even when parent doesn't exist. Fixes #704.
* Can assign `NULL`. Fixes #636

Also fixes #634 since `pluck()` is now permissive.


lionel- (approved):
Nice!

Could we add support for `zap()`, perhaps via `list_slice2<-`?

hadley:
We can't do it via `list_splice2<-` because that would break the `modify()`/`map()` size invariants, but we can special case it here.

PR #921, opened 2022-09-08
Implement list_update()

### LINKED ISSUE #822: Add support for non-recursive version of `list_modify`
Sometimes recursion is not needed when modifying lists. Would be nice to have `recursive = TRUE` argument to `list_modify` and `list_merge`. Or maybe a bit more control over which objects should be considered atomic. For instance, one often it's useful to treat `data.frames` as atomic.

```
> a <- list(x = data.frame(one = 1, two = 2))
> b <- list(x = data.frame(three = 3))
> list_modify(a, b)
# current
$x
  one two three
1   1   2     3
# desired
$x
  three
1     3
```

hadley:
An non-recursive list_modify()

Fixes #822


lionel- (approved):
The naming is a bit unfortunate because it seems easy to forget which variant does what, and then users will find themselves needing to double check the doc to understand code. But it's probably too late to do better.

hadley:
Yeah, I think ideally we'd have started with `list_modify(recurse = FALSE)`, but I don't see any way to back out of this behaviour now. But it's also not awful that `list_modify()` stays close to `modifyList()`.

PR #922, opened 2022-09-08
Implement pluck_exists()

### LINKED ISSUE #568: Equivalent of has_element() for element names in a list?
I'm wondering if there is a purrr function that detects whether an element with a given name appears in a list. For example, in the following code:
```
> library(purrr)
> ingredients <- list(tomatoes = 3, radishes = 5, cucumbers = 2)
```

I'm looking for a simple way to find out if the `ingredients` list contains "tomatoes".

I tried the `has_element()` function, but it only works with the values in a list, not the names:
```
> has_element(ingredients, "tomatoes")
[1] FALSE

> has_element(ingredients, 3)
[1] TRUE
```
This makes sense to me as you probably want to avoid the situation where you accidentally match a name in the list when you are looking to match a value instead.

I ended up writing a function like this:
```
has_element_named <- function(list, name){!is.null(list[[name]])}

> has_element_named(ingredients, "tomatoes")
[1] TRUE

> has_element_named(ingredients, "celery")
[1] FALSE
```
But this runs the same risk: that it will also match a value as well as the name:
```
> has_element_named(ingredients, 3)
[1] TRUE
```
This might be desirable in some circumstances, but most people will probably want to match against either names or values, not both. So I wrote:
```
has_element_called <- function(list, name){name %in% names(list)}

> has_element_called(ingredients, "tomatoes")
[1] TRUE
> has_element_called(ingredients, "celery")
[1] FALSE
> has_element_called(ingredients, 3)
[1] FALSE
```
Which is simple enough to do I guess, but there may be a better designed or robust function that  could be written (or maybe it already exists I just can't find it). This function doesn't, for example, account for any depth in the list you are matching against.

hadley:
Fixes #568


R/pluck.R
```diff
@@ -92,12 +95,26 @@
...
+`pluck<-` <- function(.x, ..., value) {
+  assign_in(.x, list2(...), value)
+}
+
+#' @rdname pluck
+#' @export
+pluck_exists <- function(.x, ...) {
+  check_dots_unnamed()
+
+  !identical(pluck_raw(.x, list2(...), .default = zap()), zap())
```
lionel-, line 115:
```suggestion
  !is_zap(pluck_raw(.x, list2(...), .default = zap()))
```

lionel- (approved):
Reading this doc I realised that pluck and check make a confusion between `NULL` elements and absent elements. This seems like it only makes sense in the context of `map_int()` etc, and is likely to cause programming errors elsewhere.

Do we want to add a `.null = NULL` argument that `as_mapper()` would set to `.null = .default`?

hadley:
In a slack conversation we decided that this is seems mostly like a potential problem, and we'll revisit the design if it ever becomes an actual problem.

PR #926, opened 2022-09-09
Use vec_is_list() in map_depth()

### LINKED ISSUE #920: Use `vec_is_list()` in `map_depth_rec()`

hadley:
Fixes #920


R/map-if-at.R
```diff
@@ -116,28 +125,20 @@ map_depth_rec <- function(.x,
...
+  } else {
+    if (vec_is_list(.x)) {
+      map(.x, map_depth_rec, .depth - 1, .f, ..., .ragged = .ragged)
+    } else {
+      if (.ragged) {
+        map(.x, .f, ...)
+      } else {
+        abort("List not deep enough")
+      }
     }
```
lionel-, line 141: Would the control flow be easier to read as a flat if-else chain?

R/map-if-at.R
```diff
@@ -116,28 +125,20 @@ map_depth_rec <- function(.x,
...
-
-  if (.atomic) {
-    if (!.ragged) {
-      abort("List not deep enough")
+  } else if (.depth == 0) {
+    .f(.x, ...)
+  } else if (.depth == 1) {
+    map(.x, .f, ...)
+  } else {
+    if (vec_is_list(.x)) {
```
lionel-, line 133: Probably we still want to recurse into data frames, since `map()` treats them as lists?

lionel- (approved):

hadley:
Need to make a similar change to `modify_depth()`

R/map-if-at.R
lionel-: I do want `vec_is_list()`, but purrr _may_ require special-casing data frames.

hadley: I think maybe we should stick with `is.list()` for now; it feels like we need to ensure that `pluck()`, `pluck_depth()`, and `map_depth()` should all be consistent. 

Or maybe we should have `purrr_is_list()` which would be `vec_is_list(x) || is.data.frame(x)`? But that doesn't work for `pluck()` since it makes sense to (e.g.) pluck elements out of a linear model.

lionel-: yeah, that's probably safest. Ideally you'd either use accessors or unclass before plucking, with the understanding that you're reaching into unexported internals, but I guess it's too late to be this strict.

hmm... But did `is_atomic()` use to allow reaching into linear models with `map_depth()`?

hadley: I think so, just because of the way we flipped testing for a leaf vs. testing for recursion.

hadley:
I've merged this since it's a big improvement and effectively orthogonal to what the recursion critera should be.

NEWS.md
```diff
@@ -63,10 +63,6 @@
 * Some mapping functions have now a `.progress` argument to create a
   progress bar. See `?progress_bars` (#149).
 
-* purrr is now licensed as MIT (#805).
```
hadley, line 66: Ooops. Fixed on main.

PR #927, opened 2022-09-09
Extend scope of `.at`; implement `keep_at()` and `discard_at()`

### LINKED ISSUE #817: Feature request: ability to keep/discard list elements by name
From the discussion in [this twitter thread](https://twitter.com/eamcvey/status/1352699291765248010?s=20) it seems there is a  need to remove elements from lists by name. The current "best" solution is to assign an element NULL using base R commands, which does not have an elegant tidy piping implementation. Since this is a fairly common task that needs to be done, it would be helpful to create a purrr a function that can easily be put within a sequence of piped purrr calls:

image

One approach I have been thinking of after writing that last tweet would be to create a `purrr::keep_names()` and `purrr::discard_names()`. The point of these functions would be to closely mimic the existing `purrr::keep()` and `purrr::discard()`, but to have the functions be applied to the names of the list rather than the values. It could also work on a vectors of names as the input rather than a function, for the common case when you just want to keep/remove specific elements. So something like this:

```
library(purrr)

example <- as.list(1:4)
names(example) <- list("a", "b", "c", "rstudioconf_2022","cat")

> keep_names(example, c("a","b"))
$a
[1] 1

$b
[1] 2

> discard_names(example, ~ .x %in% letters)
$rstudioconf_2022
[1] 3

$cat
[1] 4
```

And then in the case of the Twitter thread, Elaine could have just added `discard_names("b")` into her code.

---

Things I like about adding these functions:

1. They solve a problem I personally have as well.
2. They seem to open the door a little bit to more functions that work on names that could be useful in purrr. For example, I also do `x <- setnames(x, x)` a lot at the start of my purrr piping sequences and that could have a convenience function.

This I dislike about adding these functions:

1. They seem like they could be inefficient applying to each element independently. In the case of the `discard_names` example above, the anonymous function would have been called twice, but since %in% is vectorized here a single function call across the whole vector of names would have been fine to get the boolean results of the function. On a list thousands of elements long this could be a problem.
2. The set of functions you could theoretically open to having some sort of "name" equivalent (like `map_names()`) is so big I do fall into some "slippery slope" fears of this going too far.

These functions seems simple enough that I would think I could personally make a PR request to add them. I would love some feedback on if other people would want them included or if they should be changed somehow. Thank you!!

hadley:
I'm not sure the type logic is quite right. These are things I think should work:

* `discard_at(x, letters)`, regardless of what the names of `x` are.
* `keep_at(x, \(nm) nm %in% letters)`

But the second implies that `keep_at(x, names(x) %in% letters)` should also work which starts to overlap with `keep()`, since `keep(x, names(x) %in% letters))` would do the same thing.

Fixes #817


R/utils.R
```diff
@@ -8,14 +8,39 @@
...
+  if (is.numeric(at) || is.logical(at) ||is.character(at)) {
+    if (is.character(at) && !is.null(names(x))) {
+      at <- intersect(at, names(x))
+    }
+
+    at <- vec_as_subscript(at, arg = "at", call = error_call)
+    vec_as_location(
+      at,
+      length(x),
+      names(x),
```
DavisVaughan, line 28: I wonder if you want to use `names2()` everywhere to avoid this error?

```r
keep_at(1, letters)
#> Error in `vec_as_location()` at purrr/R/utils.R:25:4:
#> ! Can't use character names to index an unnamed vector.
```

R/keep.R
```diff
@@ -58,3 +59,29 @@ compact <- function(.x, .p = identity) {
...
+keep_at <- function(x, at) {
+  where <- at_selection(x, at)
+  x[where]
+}
+
+#' @export
+#' @rdname keep_at
+discard_at <- function(x, at) {
+  where <- at_selection(x, at)
+  x[-where]
```
DavisVaughan, line 86: I don't think you can use `-where` because it has issues when you select nothing

``` r
discard_at(c(x = 1), "y")
#> named numeric(0)
```

I think you could maybe use `where <- inv_which(x, where); x[where]`? At first glance that helper looks right

R/utils.R
DavisVaughan: A second example you could use as a test:

```r
keep_at(1, \(nms) nms %in% letters)
#> Error in `keep_at()`:
#> ! Must subset elements with a valid subscript vector.
#> ℹ Logical subscripts must match the size of the indexed input.
#> ✖ Input has size 1 but subscript `at` has size 0.
```

This one tests the function case

R/utils.R
```diff
@@ -8,14 +8,39 @@
...
-    lifecycle::deprecate_warn("0.4.0", I("using `vars()` in .at"))
+at_selection <- function(x, at, error_arg = caller_arg(at), error_call = caller_env()) {
+  if (is_formula(at)) {
+    at <- rlang::as_function(at, arg = error_arg, call = error_call)
+  }
+  if (is.function(at)) {
+    at <- at(names(x))
+  }
 
+  if (is.numeric(at) || is.logical(at) ||is.character(at)) {
```
DavisVaughan, line 19:
```suggestion
  if (is.numeric(at) || is.logical(at) || is.character(at)) {
```

R/utils.R
```diff
@@ -8,14 +8,39 @@
...
+    if (is.character(at) && !is.null(names(x))) {
+      at <- intersect(at, names(x))
+    }
```
DavisVaughan, line 22: Probably don't need the null check if you use `names2()`

R/utils.R
```diff
@@ -8,14 +8,39 @@
...
+
+    at <- vec_as_subscript(at, arg = "at", call = error_call)
```
DavisVaughan, line 24: I don't _think_ you need to call `vec_as_subscript()` if you follow up with `vec_as_location()`, right @lionel- ?

R/utils.R
```diff
@@ -8,14 +8,39 @@
...
+  } else if (is_quosures(at)) {
+    lifecycle::deprecate_warn("0.4.0", I("using `vars()` in .at"))
     check_installed("tidyselect", "for using tidyselect in `map_at()`.")
-    .at <- tidyselect::vars_select(.vars = nm, !!!.at)
+
+    tidyselect::vars_select(.vars = names(x), !!!at)
+  } else {
+    cli::cli_abort(
+      "{.arg {error_arg}} must be a numeric vector, character vector, or function, not {.obj_type_friendly {at}}.",
+      caller = error_call
```
DavisVaughan, line 41:
```suggestion
      call = error_call
```

And regenerate snapshots

Noticed this because a snapshot for this references `at_selection()`

tests/testthat/test-utils.R
```diff
@@ -43,15 +43,47 @@ test_that("vec_simplify() ignores complex inputs", {
...
+  expect_equal(at_selection(x, ~ .x %in% LETTERS), 2)
+  expect_equal(at_selection(x, ~ intersect(.x, LETTERS)), 2)
+})
+
+test_that("validates its inputs", {
+  x <- list(a = 1, b = 1, c = 1)
+  expect_snapshot(at_selection(x, list()), error = TRUE)
+})
+
+test_that("tidyselect in is deprecated", {
```
DavisVaughan, line 76:
```suggestion
test_that("tidyselect `at` is deprecated", {
```

Assuming you meant `at` not `in`

DavisVaughan (approved):
Mainly just some edge case comments (empty selection in discard_at and unnamed inputs)

I think the overall idea is solid though

DavisVaughan:
I don't mind that `keep_at(x, \(nm) nm %in% letters)` overlaps with `keep(x, names(x) %in% letters))` because I think the 99% case of `keep()` is to use the _values_ to determine whether or not to keep, not the names. So this is just a small special case

tests/testthat/test-utils.R
hadley: I think I meant `in at` 😄

R/utils.R
hadley: Oh yeah, it doesn't seem to do anything. I think it was a vestige from an experiment where I made `logical` vectors an error.

hadley:
You can also supply a logical vector directly to keep:

``` r
library(purrr)
x <- c(a = 1, B = 2, c = 3)

keep_at(x, \(nm) nm %in% letters)
#> a c 
#> 1 3
keep(x, names(x) %in% letters)
#> a c 
#> 1 3
```

<sup>Created on 2022-09-13 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

DavisVaughan:
Hmm. Which implies this works too

``` r
library(purrr)
x <- c(a = 1, B = 2, c = 3)

keep_at(x, c(2, 3))
#> B c 
#> 2 3

keep(x, seq_along(x) %in% c(2, 3))
#> B c 
#> 2 3
```

hadley:
It's not super useful, but it does seem nice to able to write a function that operates on names. Especially in a pipeline with the base pipe since you can't write `x |> keep(seq_along(.) %in% c(2, 3))`.

R/utils.R
lionel-: You only need to call `vec_as_subscript()` if you use different parameters than the defaults for `vec_as_location()`.

R/utils.R
```diff
@@ -8,14 +8,39 @@
...
+      call = error_call
+    )
+  } else if (is_quosures(at)) {
+    lifecycle::deprecate_warn("1.0.0", I("Using `vars()` in .at"))
     check_installed("tidyselect", "for using tidyselect in `map_at()`.")
-    .at <- tidyselect::vars_select(.vars = nm, !!!.at)
+
+    tidyselect::vars_select(.vars = names(x), !!!at)
+  } else {
+    cli::cli_abort(
```
lionel-, line 39: Pass `arg = `

R/utils.R
```diff
@@ -8,14 +8,39 @@
 #' @usage lhs \%>\% rhs
 NULL
 
-at_selection <- function(nm, .at){
-  if (is_quosures(.at)) {
-    lifecycle::deprecate_warn("1.0.0", I("using `vars()` in .at"))
+at_selection <- function(x, at, error_arg = caller_arg(at), error_call = caller_env()) {
+  if (is_formula(at)) {
```
lionel-, line 12: `is_formula()` returns `TRUE` for quosures. Do we have tests for tidyselect features?

hadley: We do, so this took me a minute to figure out — `vars` is a `<list_of<quosure>>` not a `quosure`
`

lionel-: Ah that makes sense.

PR #932, opened 2022-09-12
Make `list_flatten()` generic

lionel-:
Closes r-lib/vctrs#395.

Won't be very useful until `list_of()` supports type recursion, but I think it's better to make `list_flatten()` generic now to avoid changing behaviour in the future.

### DIFF
--- a/R/list-flatten.R
+++ b/R/list-flatten.R
@@ -10,9 +10,10 @@
 #' @param name_repair One of `"minimal"`, `"unique"`, `"universal"`, or
 #'   `"check_unique"`. See [vctrs::vec_as_names()] for the meaning of these
 #'   options.
-#' @return A list. The list might be shorter if `x` contains empty lists,
-#'   the same length if it contains lists of length 1 or no sub-lists,
-#'   or longer if it contains lists of length > 1.
+#' @return A list of the same type as `x`. The list might be shorter
+#'   if `x` contains empty lists, the same length if it contains lists
+#'   of length 1 or no sub-lists, or longer if it contains lists of
+#'   length > 1.
 #' @export
 #' @examples
 #' x <- list(1, list(2, 3), list(4, list(5)))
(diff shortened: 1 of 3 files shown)


R/list-flatten.R
```diff
@@ -45,11 +46,20 @@ list_flatten <- function(
   ) {
   vec_check_list(x)
 
-  x <- map_if(x, vec_is_list, identity, .else = list)
-  vec_unchop(
-    x,
+  # Take the proxy as we restore on exit
+  proxy <- vec_proxy(x)
+
+  # Unclass S3 lists to avoid their coercion methods. Wrap atoms in a
```
lionel-, line 52: Unclassing provides support for strict types that are not coercible to list (as the current `list_of` implementation).

R/list-flatten.R
```diff
@@ -45,11 +46,20 @@ list_flatten <- function(
   ) {
   vec_check_list(x)
 
-  x <- map_if(x, vec_is_list, identity, .else = list)
-  vec_unchop(
-    x,
+  # Take the proxy as we restore on exit
+  proxy <- vec_proxy(x)
```
lionel-, line 50: Currently this isn't necessary, as lists are always their own proxies, but it's good practice to only call restore on proxies, and also in the future we might want to have proxied lists e.g. to support special bioc types.

tests/testthat/test-list-flatten.R
```diff
@@ -22,3 +22,31 @@ test_that("can control names if both present", {
 test_that("requires a list", {
   expect_snapshot(list_flatten(1:2), error = TRUE)
 })
+
+test_that("list_flatten() restores", {
+  my_num_list <- function(...) {
```
lionel-, line 27: This simulates a recursive list-of type.

tests/testthat/test-list-flatten.R
```diff
@@ -22,3 +22,31 @@ test_that("can control names if both present", {
...
+    new_my_num_list(list2(...))
+  }
+  new_my_num_list <- function(xs) {
+    stopifnot(
+      purrr::every(xs, \(x) {
```
hadley, line 32:
```suggestion
      purrr::every(xs, function(x) {
```

tests/testthat/test-list-flatten.R
```diff
@@ -22,3 +22,31 @@ test_that("can control names if both present", {
...
+test_that("list_flatten() restores", {
+  my_num_list <- function(...) {
+    new_my_num_list(list2(...))
+  }
+  new_my_num_list <- function(xs) {
+    stopifnot(
+      every(xs, function(x) {
+        is_null(x) ||
+          is.numeric(x) ||
+          inherits(x, "my_num_list")
```
hadley, line 35:
```suggestion
        is_null(x) || is.numeric(x) || inherits(x, "my_num_list")
```

R/list-flatten.R
hadley: Add test of `list_flatten()` applied to `list_of()`?

hadley (approved):

R/list-flatten.R
lionel-: I added a test for a flat `list_of` and another for strict classes (as list-of will be relaxed soon).

PR #939, opened 2022-09-13
PoC: do not run examples on R < 4.1

gaborcsardi:
To avoid errors because of base R pipes.


hadley:
Now integrated in to #938

PR #940, opened 2022-09-13
Set `error_call` for C errors

### LINKED ISSUE #937: Update C error generation
* Teach `stop_bad_*()` how to find correct environment — https://github.com/tidyverse/purrr/pull/930#discussion_r968081200
* Update `pluck()`/`chuck()` errors
* Move recycling check in `map2()` to R code

hadley:
Fixes #937


src/pluck.c
```diff
@@ -244,19 +243,19 @@ SEXP pluck_impl(SEXP x, SEXP index, SEXP missing, SEXP strict_arg, SEXP env) {
...
       REPROTECT(x, idx);
       break;
     case S4SXP:
-      x = extract_s4(x, index_i, i, strict, env);
+      x = extract_s4(x, index_i, i, strict);
       REPROTECT(x, idx);
       break;
     default:
-      Rf_errorcall(R_NilValue, "Can't pluck from a %s.", Rf_type2char(TYPEOF(x)));
+      purrr_abort("Can't pluck from a %s.", Rf_type2char(TYPEOF(x)));
```
lionel-, line 258: rlang provides a C callable for `obj_type_friendly()` that could be useful for type errors:

https://github.com/r-lib/rlang/blob/6a32decd9e3ba87bbc2be480dd7ae9769345e4df/src/internal/init.c#L276.

It returns vmax-protected C strings, implementation is here: https://github.com/r-lib/rlang/blob/6a32decd9e3ba87bbc2be480dd7ae9769345e4df/src/internal/cnd.c#L205

lionel- (approved):
👍

src/pluck.c
hadley: How do I use a C callable? Do we have an example somewhere I can copy?

lionel-: You declare a function pointer of the right type, then assign it in some kind of initialisation routine. After that you call it like a regular function.

Here is a way that uses static variable initialisation: https://github.com/r-lib/cleancall/blob/9606df5779537da8c64a80ec96c2b7fcb722bc6a/inst/include/cleancall.h#L27-L32

PR #943, opened 2022-09-14
Standardise `_if` & `_at` helpers

hadley:
* `probe()` -> `where_if()`
* `at_selection()` -> `where_at()`

`where_at()` now generates a logical vector directly, rather than going through `inv_which()`.

---

@DavisVaughan the most useful feedback would be on the names.

### DIFF
--- a/R/deprec-splice.R
+++ b/R/deprec-splice.R
@@ -28,7 +28,7 @@ splice <- function(...) {
 }
 
 splice_if <- function(.x, .p) {
-  unspliced <- !if_idx(.x, .p)
+  unspliced <- !where_if(.x, .p)
   out <- modify_if(.x, unspliced, list)
   list_flatten(out, name_spec = "{inner}")
 }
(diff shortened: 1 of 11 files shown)


DavisVaughan (approved):
For vctrs functions that return logical vectors the same size as `x`, we use the verb `detect`, like `vec_detect_complete()`

So I might do something like:

```r
where_if -> detect_predicate (somewhat nice because internally it calls `as_predicate`)
where_at -> detect_at
```

Using `where` as the name of the returned logical vector seems fine, I don't have a better name for that

hadley:
We could rename `as_predicate` to `as_if()` 😆

DavisVaughan:
`detect_names()` feels a little weird because it allows numeric indices too. `detect_location()`? To go with the fact that it calls `vec_as_location()`?

hadley:
I think I'll just keep them as is — they're internal helpers, the names here are already substantially better, and I'm worried that `detect_*` is too confusing given the other exported `detect()` functions.

PR #945, opened 2022-09-15
Add iteration index to errors

### LINKED ISSUE #929: Report map index in error
e.g. in the reprex below, want to report that the error comes from index 3.

``` r
a <- function(x) b(x)
b <- function(x) c(x)
c <- function(x) {
  if (x > 2) {
    rlang::abort("Error!")
  } else {
    x
  }
}

purrr::map_int(1:3, a)
#> Error in `c()`:
#> ! Error!

#> Backtrace:
#>     ▆
#>  1. └─purrr::map_int(1:3, a)
#>  2.   └─global .f(.x[[i]], ...)
#>  3.     └─global b(x)
#>  4.       └─global c(x)
#>  5.         └─rlang::abort("Error!")
```

<sup>Created on 2022-09-11 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

hadley:
Fixes #929

So far, just a quick proof of concept

To do:

* [x] Wrap `map2` and `pmap` functions
* [x] Update C code as below
* [x] Pass explicit env to `check_vector()`/`r_abort()` or move that check into R
* [x] News bullet


tests/testthat/_snaps/map-depth.md
```diff
@@ -3,7 +3,13 @@
     Code
       map_depth(x1, 6, length)
     Condition
-      Error in `map_depth()`:
+      Error in `.fmap()`:
```
hadley, line 6: This is verbose, but it's technically correct and possibly helpful.

tests/testthat/_snaps/keep.md
```diff
@@ -3,11 +3,15 @@
     Code
       keep(1:3, ~NA)
     Condition
-      Error in `keep()`:
+      Error in `map_lgl()`:
```
hadley, line 6: The nesting is a bit weird here, but it is good to know where the predicate failed.

DavisVaughan:
Noting that when all of the map friends are converted over to use this, on the C side we shouldn't create and overwrite `i` like this:
https://github.com/tidyverse/purrr/blob/426acdd50424b8cd6029d237c4d4e81d94ec42a6/src/map.c#L39-L42

Since it already exists in the function environment through `i <- 0L`, we should use something like `Rf_findVarInFrame()` to look it up and modify it directly

DavisVaughan:
Can we improve on these two cases?

```r
> map(-1:1, stop)
Error in `map()`:
! Computation failed in index 1
Caused by error in `withCallingHandlers()`:
! -1
Run `rlang::last_trace()` to see where the error occurred.
```

```r
fn <- function(x) {
  if (x == 2) {
    rlang::abort("oh no")
  }
  x
}

map(-1:3, fn)
Error in `map()`:
! Computation failed in index 4
Caused by error in `.f()`:
! oh no
Run `rlang::last_trace()` to see where the error occurred.
```

R/map.R
```diff
@@ -159,3 +174,22 @@ walk <- function(.x, .f, ...) {
   map(.x, .f, ...)
   invisible(.x)
 }
+
+with_indexed_errors <- function(expr, i, error_call = caller_env()) {
+  try_fetch(
```
hadley, line 179: Use `withCallingHandlers()` ?

hadley:
@DavisVaughan I think our conclusion was that we can't easily improve on that, right?

R/map.R
```diff
@@ -160,3 +175,21 @@ walk <- function(.x, .f, ..., .progress = FALSE) {
...
+with_indexed_errors <- function(expr, i, error_call = caller_env()) {
+  withCallingHandlers(
+    expr,
+    error = function(cnd) {
+      if (i == 0L) {
+        # error happened before or after loop
+        cnd_signal(cnd)
+      } else {
+        cli::cli_abort(
+          "Can't compute index {i}.",
```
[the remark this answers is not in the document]
lionel-: For instance we'd have:

```r
f <- function(x) x + 1

tibble(z = 1:2) %>%
  group_by(z) %>%
  summarise(y = f(""))
#> Error in `summarise()`:
#> ℹ In argument `y = f("")`.
#> ℹ In group 1: `z = 1`.
#> Caused by error in `x + 1`:
#> ! non-numeric argument to binary operator
```

Instead of:

```r
#> Error in `summarise()`:
#> ! Can't compute `y = f("")`.
#> ℹ In group 1: `z = 1`.
#> Caused by error in `x + 1`:
#> ! non-numeric argument to binary operator
```

By contrast, in this case, we'd keep the "Can't" form because an error actually happened at the contextual level:

```r
tibble(z = 1:2) %>%
  group_by(z) %>%
  summarise(x = 1:3, y = 1:2)
#> Error in `summarise()`:
#> ! Can't recycle `y = 1:2`.
#> ℹ In group 1: `z = 1`.
#> Caused by error:
#> ! `y` must be size 3 or 1, not 2.
#> ℹ An earlier column had size 3.
```

tests/testthat/_snaps/map2.md
```diff
@@ -3,13 +3,17 @@
...
+      Caused by error:
+      ! Can't coerce from a character to a integer.
     Code
       map2_int(1, 1, ~ 1:2)
     Condition
       Error in `map2_int()`:
-      ! Result 1 must have length 1, not 2.
+      ! Can't compute index 1.
+      Caused by error:
+      ! Result must be length 1, not 2
```
lionel-, line 16: Missing full stop.

tests/testthat/_snaps/map2.md
```diff
@@ -3,13 +3,17 @@
     Code
       map2_int(1, 1, ~"x")
     Condition
-      Error:
-      ! Can't coerce element 1 from a character to a integer
+      Error in `map2_int()`:
+      ! Can't compute index 1.
+      Caused by error:
+      ! Can't coerce from a character to a integer.
```
lionel-, line 9: should be "an integer". You can use the `obj_type_friendly()` callable from rlang if that's helpful.

lionel- (commented):
Looks good! But I think it'd make sense to think about `call` arguments to take ownership of map errors, as argued inline.

src/conditions.c
```diff
@@ -45,6 +37,30 @@ void r_abort(const char* fmt, ...) {
...
+  va_start(dots, fmt);
+  vsnprintf(buf, BUFSIZE, fmt, dots);
+  va_end(dots);
+  buf[BUFSIZE - 1] = '\0';
+
+  SEXP env = PROTECT(caller_env());
+  r_abort0(env, buf);
+}
+
+void r_abort_call(SEXP env, const char* fmt, ...) {
```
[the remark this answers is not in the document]
hadley: Agreed 😬

tests/testthat/_snaps/keep.md
hadley: Yeah, and it was easy to implement once I refactored the code.

R/map.R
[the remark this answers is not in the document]
DavisVaughan: I like that `In` is for "informational" bullets (i.e. the ones that use the `i` bullet) and `Can't` is for error bullets that are prefixed with `!` or `x`

R/map.R
```diff
@@ -112,38 +112,44 @@
...
+  map_("double", .x, .f, ..., .progress = .progress)
 }
 
 #' @rdname map
 #' @export
 map_chr <- function(.x, .f, ..., .progress = FALSE) {
+  map_("character", .x, .f, ..., .progress = .progress)
+}
+
+map_ <- function(.type, .x, .f, ..., .progress = FALSE, .error_call = caller_env()) {
```
DavisVaughan, line 142: A side effect of having `.error_call` here is that now functions that pass `.error_call` through the dots won't be able to pass that error call all the way through to `.f`, it will get partially matched here.

A bunch of vctrs functions are getting an `.error_call` arg

Same with `.type` and `.progress` I guess

``` r
library(purrr)

error_call <- current_env()

# `error_call` is passed through to `vec_rep2()`
vec_rep2 <- function(x, times, error_call = current_env()) {
  vec_rep(x, times, call = error_call)
}
map(1:5, vec_rep2, times = -1, error_call = error_call)
#> Error in `map()`:
#> ! In index 1.
#> Caused by error:
#> ! `times` must be a positive number.

# `.error_call` is matched by `map_()` and doesn't make it
# to `vec_rep2()`
vec_rep2 <- function(x, times, .error_call = current_env()) {
  vec_rep(x, times, call = .error_call)
}
map(1:5, vec_rep2, times = -1, .error_call = error_call)
#> Error:
#> ! In index 1.
#> Caused by error in `.f()`:
#> ! `times` must be a positive number.
```

R/map.R
```diff
@@ -112,38 +112,44 @@
...
   .f <- as_mapper(.f, ...)
-  .Call(map_impl, environment(), ".x", ".f", "character", .progress)
+  i <- 0L
+  with_indexed_errors(
+    i = i,
+    error_call = .error_call,
+    .Call(map_impl, environment(), .type, .progress, .error_call)
+  )
 }
```
DavisVaughan, line 150: It is still pretty unfortunate that we get parent call information that mentions `map_()`

``` r
library(vctrs)
library(purrr)

map(1:5, vec_check_list)
#> Error in `map()`:
#> ! In index 1.
#> Caused by error in `map_()` at purrr/R/map.R:115:2:
#> ! `.x[[i]]` must be a list, not an integer.
```

It seems rare that the parent call is actually useful. I had mentioned that we might consider setting `cnd$call <- NULL` or hardcoding it to `.f`. Is that still an option? I feel like it would generally be less confusing because we'd have less chance of displaying confusing calls.

So it would look like:

```r
> map(1:5, vec_check_list)
Error in `map()`:
! In index 1.
Caused by error:
! `.x[[i]]` must be a list, not an integer.
Run `rlang::last_trace()` to see where the error occurred.
```

R/map.R
```diff
@@ -160,3 +166,20 @@ walk <- function(.x, .f, ..., .progress = FALSE) {
...
+    error = function(cnd) {
+      if (i == 0L) {
+        # Error happened before or after loop
+      } else {
+        cli::cli_abort(
+          "In index {i}.",
+          parent = cnd,
+          call = error_call
+        )
+      }
```
DavisVaughan, line 182:
```suggestion
      if (i != 0L) {
        # If `i == 0L`, error happened before or after loop
        cli::cli_abort(
          "In index {i}.",
          parent = cnd,
          call = error_call
        )
      }
```

Possibly cleaner, depending on how you feel about it

src/coerce.c
```diff
@@ -3,9 +3,15 @@
 #include <Rinternals.h>
 #include <stdio.h>
 
+#include "conditions.h"
+
 void cant_coerce(SEXP from, SEXP to, int i) {
-  Rf_errorcall(R_NilValue, "Can't coerce element %i from a %s to a %s",
-    i + 1, Rf_type2char(TYPEOF(from)), Rf_type2char(TYPEOF(to)));
+  Rf_errorcall(
+    R_NilValue,
```
DavisVaughan, line 10: Should the rlang `abort()` be used here?

R/map.R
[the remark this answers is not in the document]
lionel-: What do you think about transforming the "! Can't compute" bullets in "i In argument" bullets?

I think I'll open a dplyr PR with this change to discuss the proposal.

src/map.c
```diff
@@ -53,29 +54,28 @@ SEXP call_loop(SEXP env, SEXP call, int n, SEXPTYPE type, int force_args,
     SEXP res = PROTECT(R_forceAndCall(call, force_args, env));
 
     if (type != VECSXP && Rf_length(res) != 1) {
-      stop_bad_element_length(res, i + 1, 1, "Result", NULL, false);
+      Rf_errorcall(R_NilValue, "Result must be length 1, not %i.", Rf_length(res));
```
DavisVaughan, line 57: `r_abort()`?

src/map.c
```diff
@@ -99,30 +99,24 @@ SEXP map_impl(SEXP env, SEXP x_name_, SEXP f_name_, SEXP type_, SEXP progress) {
...
   if (nx != ny && nx != 1 && ny != 1) {
-    r_abort(
-      "Mapped vectors must have consistent lengths:\n"
-      "* `.x` has length %d\n"
-      "* `.y` has length %d",
-      nx,
-      ny
+    r_abort_call(
+      error_call,
+      "`.y must have length 1 or %i, not %i.",
```
DavisVaughan, line 118:
```suggestion
      "`.y` must be length 1 or %i, not %i.",
```

More consistent with `Result must be length 1` error from above

src/map.c
```diff
@@ -140,14 +134,17 @@ SEXP map2_impl(SEXP env, SEXP x_name_, SEXP y_name_, SEXP f_name_, SEXP type_, S
...
 
-SEXP pmap_impl(SEXP env, SEXP l_name_, SEXP f_name_, SEXP type_, SEXP progress) {
-  const char* l_name = CHAR(Rf_asChar(l_name_));
-  SEXP l = Rf_install(l_name);
+SEXP pmap_impl(SEXP env, SEXP type_, SEXP progress, SEXP error_call) {
+  SEXP l = Rf_install(".l");
   SEXP l_val = PROTECT(Rf_eval(l, env));
   SEXPTYPE type = Rf_str2type(CHAR(Rf_asChar(type_)));
 
   if (!Rf_isVectorList(l_val)) {
```
DavisVaughan, line 142:
```suggestion
  if (TYPEOF(l_val) != VECSXP) {
```

`Rf_isVectorList()` allows `EXPRSXP` for some reason and I feel like we don't want that. Does that feel right @lionel- ? https://github.com/wch/r-source/blob/79298c499218846d14500255efd622b5021c10ec/src/include/Rinlinedfuns.h#L851

src/map.c
```diff
@@ -170,7 +172,11 @@ SEXP pmap_impl(SEXP env, SEXP l_name_, SEXP f_name_, SEXP type_, SEXP progress)
     if (n == -1) {
       n = nj;
     } else if (nj != n) {
-      stop_bad_element_length(j_val, j + 1, n, NULL, ".l", true);
+      r_abort_call(
+        error_call,
+        "`.l[[%i]]` must have length 1 or %i, not %i.",
```
DavisVaughan, line 177:
```suggestion
        "`.l[[%i]]` must be length 1 or %i, not %i.",
```

src/map.c
DavisVaughan: (at the very least, catch the missing backtick on `.y`)

DavisVaughan (approved):

src/map.c
hadley: These deliberately use `Rf_errorcall` because all the necessary context is already included in the wrapper.

src/map.c
hadley: I'm not going to do this here in favour of using vctrs errors in #957

src/map.c
hadley: I will fix the backtick though.

R/map.R
hadley: Hmmm, probably worth making this `..error_call`, but generally this is why using `map(1:5, \(x) vec_rep2(x, times = -1, .error_call = error_call))` is a good idea.

R/map.R
hadley: I seem to be generally happy with an `if` branch that only contains a comment.

R/map.R
hadley: If I strip the call from the error, I definitely lose useful information in the tests. So I think in most cases it is helpful, although it doesn't work so well with error checking functions. And you can already fix that if needed:

``` r
library(purrr)

my_f <- function() {
  map(1:5, \(x) vctrs::vec_check_list(x, call = NULL))
}
my_f()
#> Error in `map()`:
#> ! Computation failed in index 1
#> Caused by error:
#> ! `x` must be a list, not an integer.
```

<sup>Created on 2022-09-23 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

PR #946, opened 2022-09-15
Implement `rmodify()`

### LINKED ISSUE #720: Recursive map over nested lists
A common scenario for me is to map over nested lists. AFAIK there is no built in or purr functionality for this. 

Basically an rapply version but which would operate not just on leaves but also on the inner nodes. 

Example:

```R
rmap(sessionInfo(), unclass)
```

hadley:
Fixes #720

@vspinu here's a concrete proposal for `rmodify()` that I'd love to get your feedback on.

To do:
* [x] News bullet
* [x] Tests
* [x] Update examples once #938 is merged


R/modify-recursive.R
```diff
@@ -0,0 +1,54 @@
+#' Recursively modify a list
+#'
+#' `rmodify()` allows you to recursively modify a vector, supplying functions
+#' to either modify each leaf or each node (or both).
+#'
+#' @param x A list.
+#' @param f_leaf A function applied to each leaf.
+#' @param f_pre A function applied to each node or leaf, before leaves are
+#'   transformed.
```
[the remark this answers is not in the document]
vspinu: This actually made me thinking, is it necessary to apply `f_pre` and `f_post` to leafs? I think most commonly the `f_pre` and `f_post` will be used on nodes and in general the same operation is unlikely to be applied on both leafs and nodes. As a result a user would need to add an extra check for leafs inside `f_pre` and `f_post`. 

Also applying all 3 transformations on leafs is redundant and slightly confusing.

NAMESPACE
```diff
@@ -200,6 +200,7 @@ export(reduce2_right)
 export(reduce_right)
 export(rep_along)
 export(rerun)
+export(rmodify)
 export(safely)
```
vspinu, line 204: I would add a link to `modify` doc page as well.

R/modify-recursive.R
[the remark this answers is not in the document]
hadley: I’m glad you bought that up because that’s what I was thinking too 😀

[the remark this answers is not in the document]
hadley: Hmmmm, what happens if `f_pre` changes `x` from a node to a leaf?

R/modify-recursive.R
```diff
@@ -0,0 +1,54 @@
...
+#' @param p_leaf A function that determines whether an element is a leaf or a
+#'   node. The default value, `NULL`, will treats lists as nodes and everything
+#'   else as leaves.
+#' @export
+#' @examples
+#' x <- list(list(1, list(2), list(3)))
+#' x %>% str()
+#' x %>% rmodify(function(x) x + 1) %>% str()
+rmodify <- function(x,
+                    f_leaf = identity,
```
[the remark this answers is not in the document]
hadley: Yeah, good idea.

R/map-depth.R
```diff
@@ -14,6 +14,8 @@
 #' @param .ragged If `TRUE`, will apply to leaves, even if they're not
 #'   at depth `.depth`. If `FALSE`, will throw an error if there are
 #'   no elements at depth `.depth`.
+#' @seealso [rmodify()] for a recursive version of `modify_depth()` that
+#'   allows you allow apply a function to every leaf or every node.
```
hyiltiz, line 18: allows you *to* apply?

R/modify-recursive.R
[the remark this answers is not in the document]
hyiltiz: As a `map`-like function, I think it should preserve the types/shape of the data structure; `f_pre`/`f_post` should have `Node -> Node` type and should not delete a node or transmute a node into a leaf. Otherwise, the behavior is a bit surprising, depending on how the tree is traversed.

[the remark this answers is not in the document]
vspinu: I think being able to change a type of a node is a very important feature. For example one might want to change the end list nodes (prior to leafs) into a proper atomic vector. 

I find it hard to think of good use-cases for f_pre, but the one you mentioned seem an important one. It's quite possible that the purpose of an operation is to transform a sub-tree into a leaf. The traversal stops right there. This can be important for deeply nested trees where performance starts playing a role.  On the other hand `p_leaf` can capture this use case as well.  

So there are 4 options:

  1) leave as it is and add a note that `f_pre` should return a node, otherwise the behavior is undefined.
  2) add an extra check for leaf after f_pre
  3) make f_pre/f_post asymetric and apply f_pre to the leaf node but not f_post
  4) remove f_leaf and add extra parameter to `f_post/f_pre`, `leaf=TRUE`

I am somewhat indifferent between 1 and 2. Though 2 seems a bit cleaner conceptually and the default penalty of `is.list` is trivial.

R/modify-recursive.R
```diff
@@ -0,0 +1,68 @@
...
+                    p_leaf = NULL,
+                    f_pre = identity,
+                    f_post = identity) {
+  if (!is_vector(x)) {
+    cli::cli_abort("{.arg x} must be a vector, not {.obj_type_friendly {x}}.")
+  }
+
+  f_post <- rlang::as_function(f_post)
+  f_pre <- rlang::as_function(f_pre)
+  f_leaf <- rlang::as_function(f_leaf)
```
lionel-, line 46: Do we still use `as_function()` in new mappers or do we force new users to write either `function` or `\`?

R/modify-recursive.R
```diff
@@ -0,0 +1,68 @@
...
+  if (is.null(p_leaf)) {
+    is_leaf <- function(x) {
+      !is.list(x)
```
lionel-, line 49: I would have expected something like `vec_is_list()` here.

R/modify-recursive.R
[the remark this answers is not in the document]
lionel-: Also take `...` with empty dots check to force named argument passing?

R/modify-recursive.R
```diff
@@ -0,0 +1,68 @@
+#' Recursively modify a list
+#'
+#' `rmodify()` allows you to recursively modify a list, supplying functions
+#' that either modify each leaf or each node (or both).
+#'
+#' @param x A list.
+#' @param f_leaf A function applied to each leaf.
```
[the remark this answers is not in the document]
lionel-: e.g. `f_leaf` and `p_leaf` could be `leaf` and `is_leaf`.

R/modify-recursive.R
```diff
@@ -0,0 +1,68 @@
...
+#' }
+#' x |> rmodify(f_post = sort_named) |> str()
+rmodify <- function(x,
+                    f_leaf = identity,
+                    p_leaf = NULL,
+                    f_pre = identity,
+                    f_post = identity) {
+  if (!is_vector(x)) {
+    cli::cli_abort("{.arg x} must be a vector, not {.obj_type_friendly {x}}.")
+  }
```
lionel-, line 42: I think it'd be more consistent if we checked for a leaf and return `leaf(x)` in that case.

R/modify-recursive.R
hadley: I think we'll use `as_function()` for a while, since people will be accustomed to using `~`. I don't think we're at the point where we want to start deprecating formulas.

lionel-: I think it'd be more like they are superseded rather than deprecated. I don't think we'll be able to deprecate formulas in purrr since they have been so central to the package.

R/modify-recursive.R
```diff
@@ -0,0 +1,68 @@
...
+#' sort_named <- function(x) {
+#'   nms <- names(x)
+#'   if (!is.null(nms)) {
+#'     x[order(nms)]
+#'   } else {
+#'     x
+#'    }
+#' }
+#' x |> rmodify(f_post = sort_named) |> str()
+rmodify <- function(x,
```
DavisVaughan, line 35: Empty `...`

R/modify-recursive.R
DavisVaughan: Why don't we require `x` to be a list with `vec_check_list()`?

I feel like allowing non-lists isn't useful because pre/post won't ever be applied? I guess that depends on what `p_leaf()` returns though (i.e. in theory it could say a numeric value is a node, but i feel like that isn't useful)

R/modify-recursive.R
DavisVaughan: Implying we might need some tests for determining whether a data frame is a node or a leaf

R/modify-recursive.R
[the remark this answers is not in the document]
DavisVaughan: And `pre` and `post`

---

If you decide to keep the first form, I feel like `f_is_leaf` is better than `p_leaf`?

tests/testthat/test-modify-recursive.R
```diff
@@ -0,0 +1,25 @@
...
+  expect_equal(
+    rmodify(list(1, list(1, list(1))), ~ .x + 9),
+    list(10, list(10, list(10)))
+  )
+})
+
+test_that("can modify nodes", {
+  expect_equal(
+    rmodify(list(1, list(2, list(3))), f_post = list_flatten),
+    list(1, 2, 3)
```
DavisVaughan, line 16: Neat, it's like `list_squash()`

R/modify-recursive.R
```diff
@@ -0,0 +1,68 @@
+#' Recursively modify a list
+#'
+#' `rmodify()` allows you to recursively modify a list, supplying functions
```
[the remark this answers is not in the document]
DavisVaughan: Strong agree!

DavisVaughan (approved):

R/modify-recursive.R
hadley: I think there's probably some invariant around  `modify_tree(x, f_pre = \(y) modify_tree(y, f_leaf)) == modify_tree(x, f_leaf)`

R/modify-recursive.R
hadley: My gut feeling is that `is.list` is right here, but I can't articulate why and if you both think `vec_is_list()` is correct then that's probably right.

lionel-: Perhaps because of Vitalie's example about stripping environments of formulas? That's a case where you want `modify_tree()` to reach everywhere. Should we export a predicate for this case, i.e. to reach everywhere `[[<-` can? But I think the default should be to only descend explicit lists.

R/modify-recursive.R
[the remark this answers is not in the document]
vspinu: My preference is for `fleaf`, `fpre` and `pleaf`. Looks, reads and writes nicer.  Unprefixed is also reasonable. `f_is_leaf` a bit of a torment IMO.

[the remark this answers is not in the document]
lionel-: `fleaf` is not the house style though, so it's either `f_leaf` or `leaf`.

R/modify-recursive.R
lionel-: We're generally switching to `vec_is()` and `vec_is_list()` to determine scalarness and listness.

R/modify-recursive.R
vspinu: My preference would be to allow for as wide class of objects as possible. @lionel-'s point about allowing `rmodify` on all objects that `[[<-` is allowed on makes sense to me. 

Use cases: 
  - Side effectful recursive modification of nested environments
  - Recursive processing of language objects. For instance, if one wants to modify specific parts of a function.

R/modify-recursive.R
hadley: @lionel- generally, yes, but not that much inside of purrr to maintain backward compatibility. I think the main question here is it better for purrr to be internally consistent or for newer functions to be consistent with newer vctrs standards. I suspect package consistency is more important.

(That said, I think @vspinu argued that `list_modify()` shouldn't recurse into data frames)

R/modify-recursive.R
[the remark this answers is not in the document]
hadley: I think I like unprefixed best.

R/modify-recursive.R
hadley: In the interests of simplicity, lets stick with `is.list` here. I do expect that many people will need to customise `is_leaf` so it's ok for the default to have limited use.

PR #949, opened 2022-09-17
Use vctrs argument name instead of purrr

hadley:
Since it's more informative

### DIFF
--- a/R/list-combine.R
+++ b/R/list-combine.R
@@ -14,10 +14,10 @@
 #'   only contain data frames.
 #' @param ptype An optional prototype to ensure that the output type is always
 #'   the same.
-#' @param id By default, `names(x)` are lost. Alternatively, supply a string
-#'   and the names will be saved into a column with name `{id}`. If `id`
-#'   is supplied and `x` is not named, the position of the elements will
-#'   be used instead of the names.
+#' @param names_to By default, `names(x)` are lost. To keep them, supply a
+#'   string to `names_to` and the names will be saved into a column with that
+#'   name. If `names_to` is supplied and `x` is not named, the position of
+#'   the elements will be used instead of the names.
 #' @param size An optional integer size to ensure that every input has the
 #'   same size (i.e. number of rows).
 #' @param name_repair One of `"unique"`, `"universal"`, or `"check_unique"`.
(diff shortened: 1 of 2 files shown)


R/list-combine.R
```diff
@@ -14,10 +14,10 @@
...
 #' @param ptype An optional prototype to ensure that the output type is always
 #'   the same.
-#' @param id By default, `names(x)` are lost. Alternatively, supply a string
-#'   and the names will be saved into a column with name `{id}`. If `id`
-#'   is supplied and `x` is not named, the position of the elements will
-#'   be used instead of the names.
+#' @param names_to By default, `names(x)` are lost. To keep them, supply a
+#'   string to `names_to` and the names will be saved into a column with that
+#'   name. If `names_to` is supplied and `x` is not named, the position of
+#'   the elements will be used instead of the names.
```
DavisVaughan, line 20: I was showing this function to Max, and he was rather excited about the fact that when positions are used, they will actually show up as an _integer_ vector rather than what `bind_rows()` does, which is to create a _character_ vector of positions 😱 

``` r
dplyr::bind_rows(dplyr::tibble(x="a"), .id = "id")
#> # A tibble: 1 × 2
#>   id    x    
#>   <chr> <chr>
#> 1 1     a
```

<sup>Created on 2022-09-17 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

DavisVaughan (approved):
Agreed that this is a more informative name

R/list-combine.R
hadley: Did you check that `list_rbind()` is different? Because I have a bad feeling…

hadley: Oh weird:

``` r
df <- tibble::tibble(x = 1)
purrr::list_rbind(list(df, df), names_to = "id")
#> # A tibble: 2 × 2
#>      id     x
#>   <int> <dbl>
#> 1     1     1
#> 2     2     1
dplyr::bind_rows(df, df, .id = "id")
#> # A tibble: 2 × 2
#>   id        x
#>   <chr> <dbl>
#> 1 1         1
#> 2 2         1
```

<sup>Created on 2022-09-18 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

I don't understand what's happening here because `dplyr::bind_rows()` finishes by calling `vctrs::vec_rbind()`. 

....

Hmmm, that's because `dplyr::bind_rows()` does `names(dots) <- seq_along(dots)` for reasons I don't understand.

PR #952, opened 2022-09-20
Combine functions should ignore NULL

hadley:


R/list-combine.R
```diff
@@ -65,15 +65,15 @@ list_rbind <- function(x, names_to = rlang::zap(), ptype = NULL) {
...
-  if (all(is_df)) {
+  if (all(is_df_or_null)) {
     return()
   }
 
-  bad <- which(!is_df)
+  bad <- which(!is_df_or_null)
   cli::cli_abort(
-    "All elements of {.arg x} must be data frames. Elements {bad} are not.",
+    "Each element of {.arg x} must be either a data frame or NULL. Elements {bad} are not.",
```
DavisVaughan, line 76:
```suggestion
    "Each element of {.arg x} must be either a data frame or {.code NULL}. Elements {bad} are not.",
```

Maybe? Seems weird to not at least backtick null

tests/testthat/test-list-combine.R
```diff
@@ -52,6 +52,15 @@ test_that("list_rbind() can enforce ptype", {
...
 })
 
+test_that("NULLs are ignored", {
+  df1 <- data.frame(x = 1)
+  df2 <- data.frame(y = 1)
+
+  expect_equal(list_c(list(1, NULL, 2)), c(1, 2))
+  expect_equal(list_rbind(list(df1, NULL, df1)), vec_rbind(df1, df1))
+  expect_equal(list_cbind(list(df1, NULL, df2)), vec_cbind(df1, df2))
+})
```
DavisVaughan, line 62: Do you have tests for empty lists? Like:

```r
list_c(list())
list_rbind(list())
list_cbind(list())

list_c(list(NULL))
list_rbind(list(NULL))
list_cbind(list(NULL))
```

Even though this is basically testing a vctrs behavior, it still checks that `check_list_of_data_frames()` and anything else you build on top of the vctrs functions works with empty input, so it may be worth it.

DavisVaughan (approved):

PR #953, opened 2022-09-20
Default otherwise to NULL

hadley:
I feel like this is probably the most common/useful value and increases the consistency with `pluck()`, and makes it work nicely with #952.


R/adverb-possibly.R
```diff
@@ -12,7 +12,7 @@
 #' # To replace errors with a default value, use possibly().
 #' list("a", 10, 100) |>
 #'   map_dbl(possibly(log, NA_real_))
```
DavisVaughan, line 14: Maybe worth adding a `map(x, possibly(f)) %>% list_rbind()` example?

DavisVaughan (approved):
Seems reasonable to me

PR #956, opened 2022-09-21
Add dots to new functions

### LINKED ISSUE #951: Consider giving new functions empty `...`
- Forces names of optional args for better user code robustness and readability 
- Allows us the freedom to add new optional args in whatever order we choose in the future

Full list of functions that could get empty dots:

```r
list_flatten
list_c
list_cbind
list_rbind
list_simplify
list_transpose
```

hadley:
Fixes #951

### DIFF
--- a/R/list-combine.R
+++ b/R/list-combine.R
@@ -12,6 +12,7 @@
 #'
 #' @param x A list. For `list_rbind()` and `list_cbind()` the list must
 #'   only contain only data frames or `NULL`.
+#' @param ... Reserved for future usage. Must be empty.
 #' @param ptype An optional prototype to ensure that the output type is always
 #'   the same.
 #' @param names_to By default, `names(x)` are lost. To keep them, supply a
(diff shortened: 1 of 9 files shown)


R/list-combine.R
```diff
@@ -12,6 +12,7 @@
 #'
 #' @param x A list. For `list_rbind()` and `list_cbind()` the list must
 #'   only contain only data frames or `NULL`.
+#' @param ... Reserved for future usage. Must be empty.
```
DavisVaughan, line 15: You can use

```r
#' @inheritParams rlang::args_dots_empty
```

for consistent documenting of empty dots

R/list-combine.R
```diff
@@ -36,27 +37,31 @@
 #' list_rbind(unname(x2), names_to = "id")
 #'
 #' list_cbind(x2)
-list_c <- function(x, ptype = NULL) {
+list_c <- function(x, ..., ptype = NULL) {
   vec_check_list(x)
+  check_dots_empty()
```
DavisVaughan, line 42: This is like the most mild of comments, but I always make `check_dots_empty()` the _very first_ thing my functions do. Somehow that shows that it is a technical check rather than more meaningful check like `check_bool()` or `check_list_of_data_frames()`

R/list-combine.R
DavisVaughan: Here and in the other places you added dots

DavisVaughan (approved):
Thanks!

R/list-combine.R
hadley: IMO it's simpler to check the arguments in the order they appear, with no special cases.

PR #959, opened 2022-09-22
Tweak ... advice

hadley:


lionel- (approved):

R/map.R
```diff
@@ -33,8 +33,19 @@
 #'
 #' @param ... Additional arguments passed on to the mapped function.
 #'
-#'   Note that the arguments that differ in each call come before `.f`,
-#'   and the arguments that are the same come after `.f`.
+#'   We now generally recommend against using `...` to pass additional
```
DavisVaughan, line 36:
```suggestion
#'   We now generally don't recommend using `...` to pass additional
```

Feels clearer to me

R/map.R
```diff
@@ -33,8 +33,19 @@
...
+#'   (constant) arguments to `f`. Instead use a shorthand anonymous function:
```
DavisVaughan, line 37:
```suggestion
#'   (constant) arguments to `.f`. Instead use a shorthand anonymous function:
```

DavisVaughan (approved):

R/map.R
hadley: I think this changes the meaning — my wording implies that we recommend you don't do it; your wording implies that we don't recommend you do it 😄

hadley: It's like the difference between "might could" and "could might" 😝

PR #961, opened 2022-09-23
Use vctrs to perform input validation

hadley:


R/map2.R
```diff
@@ -57,6 +57,13 @@ map2_chr <- function(.x, .y, .f, ..., .progress = FALSE) {
 map2_ <- function(.type, .x, .y, .f, ..., .progress = FALSE, ..error_call = caller_env()) {
   .f <- as_mapper(.f, ...)
   i <- 0L
+
+  .x <- vctrs_list_compat(.x)
```
DavisVaughan, line 61: I feel like `map()` should do this just for consistency

src/map.c
```diff
@@ -107,30 +107,19 @@ SEXP map2_impl(SEXP env, SEXP type_, SEXP progress, SEXP error_call) {
   SEXPTYPE type = Rf_str2type(CHAR(Rf_asChar(type_)));
 
   SEXP x_val = PROTECT(Rf_eval(x, env));
-  check_vector(x_val, ".x", error_call);
   SEXP y_val = PROTECT(Rf_eval(y, env));
```
DavisVaughan, line 110: Local compilation is giving me

```
map.c:110:8: warning: unused variable 'y_val' [-Wunused-variable]
     SEXP y_val = PROTECT(Rf_eval(y, env));
          ^
   1 warning generated.
```

R/utils.R
```diff
@@ -98,3 +98,12 @@ is_quantity <- function(x) {
   }
   x
 }
+
+# When we want to use vctrs, but treat lists like purrr does
+vctrs_list_compat <- function(x) {
+  if (is.list(x) && !vec_is_list(x)) {
+    unclass(x)
```
DavisVaughan, line 105: I don't think this is quite the right heuristic. It means that rcrd vectors will also be unclassed, and we don't want that.

I guess it can't be just this

```r
if (is.data.frame(x)) {
  unclass(x)
} else {
  x
}
```

because we want scalar lists like `lm` objects to be recyclable and work with `map2()` and friends for backwards compat.

So maybe it is like this?

```r
if (is.data.frame(x) || (is.list(x) && !vec_is(x))) {
  # Treat data frames and S3 scalar lists like bare lists.
  # Ensures rcrd vctrs retain their class.
  unclass(x)
} else {
  x
}
```

DavisVaughan: After we fix #963, we should definitely add some tests for rcrd vectors with `map()`, `map2()`, and `pmap()` to make sure the 3 important cases work as expected

R/pmap.R
```diff
@@ -101,11 +101,14 @@ pmap_chr <- function(.l, .f, ..., .progress = FALSE) {
...
-  if (is.data.frame(.l)) {
-    .l <- as.list(.l)
-  }
-  i <- 0L
 
+  .l <- vctrs_list_compat(.l)
+  vec_check_list(.l, call = ..error_call)
+
+  .l <- map(.l, vctrs_list_compat)
+  .l <- vec_recycle_common(!!!.l, .arg = ".l", .call = ..error_call)
```
DavisVaughan, line 109: At this point should `list_check_all_vectors()` also be used? It seems like there were some C checks to ensure that each element is a vector

DavisVaughan: Hmm, I guess it errors on `NULL` list elements and it seems like we don't want that?

DavisVaughan: Oh actually `vec_recycle_common()` will take care of this for us, and handles `NULL` right too

DavisVaughan, line 109: I'm going to make the assumption that recycling in the map2/pmap case really isn't very expensive, because most people won't be doing `map2(1:1000, 1L)` - i.e. where one of the inputs actually needs recycling.

If they are doing that, they probably should have just supplied `.y` as a fixed argument

(This is compared to what we currently do, which is _don't_ recycle but instead use `x[[1L]]` repeatedly)

R/map2.R
DavisVaughan: And in `map_()` you could do:

```r
.x <- vctrs_list_compat(.x)
vec_assert(.x, arg = ".x", call = ..error_call)
```

Since there is no `vec_recycle_common()` call, that would ensure you get the vector check.

This would allow you to remove the usage of `check_vector()` in `map_impl()`, then you can remove that `check_vector()` helper altogether

DavisVaughan: `map_()` probably _has_ to call `vctrs_list_compat()` to ensure that the length is computed correctly when we fix #963

DavisVaughan (changes requested):

DavisVaughan:
Also FYI you'll need https://github.com/tidyverse/purrr/commit/33444112a4c1edae6d4c7338a7eddefe6c967641 for checks to pass again

R/pmap.R
hadley: Yeah, I'm also assuming vctrs will use ALTREP for recycling one day, making the cost even lower.

R/utils.R
```diff
@@ -103,7 +103,9 @@ is_quantity <- function(x) {
 # Treat data frames and S3 scalar lists like bare lists.
 # But ensure rcrd vctrs retain their class.
 vctrs_list_compat <- function(x) {
-  if (is.data.frame(x) || (is.list(x) && !vec_is(x))) {
+  if (is.null(x)) {
+    list()
```
DavisVaughan, line 107: Nice idea!

R/utils.R
```diff
@@ -99,12 +99,31 @@ is_quantity <- function(x) {
   x
 }
 
+vctrs_list_compat <- function(x, error_call = caller_env(), error_arg = caller_arg(x)) {
+  force(error_arg)
```
DavisVaughan, line 103: Is this required? I thought part of the point of passing `error_call` and `error_arg` through in the specific way that we do is to avoid forcing the promises until they are actually required

hadley: I had to do it to get a good error.

lionel-: Maybe preserve `x` by assigning the result of `vctrs_vec_compat(x)` to `out` instead of `x`?

I agree that the default pattern should be to evaluate these things lazily.

DavisVaughan: Oh right, I couldn't figure out what the problem was but now I see that what Lionel suggests is a nice simple fix that maintains the standard pattern

```r
vctrs_list_compat <- function(x, error_call = caller_env(), error_arg = caller_arg(x)) {
  out <- vctrs_vec_compat(x)
  vec_check_list(out, call = error_call, arg = error_arg)
  out
}
```

hadley: Oh yeah, that's better. I pushed a fix to main.

PR #966, opened 2022-09-27
Use is_node instead of is_leaf

hadley:


R/modify-tree.R
```diff
@@ -6,10 +6,10 @@
 #' @param x A list.
 #' @param ... Reserved for future use. Must be empty
 #' @param leaf A function applied to each leaf.
-#' @param is_leaf A predicate function that determines whether an element is
-#'   a leaf (by returning `FALSE`) or a node (by returning `FALSE`). The
+#' @param is_node A predicate function that determines whether an element is
+#'   a node (by returning `TRUE`) or a leaf (by returning `FALSE`). The
 #'   default value, `NULL`, treats bare lists as nodes and everything else
-#'   (including data frames) as leaves.
+#'   (including data frames) as leaves, using [vctrs::vec_is_list()].
```
lionel-, line 12: - including rich objects stored as lists like data frames and linear models
- Use `is.list()` to traverse all objects stored as lists

lionel- (approved):

PR #967, opened 2022-09-28
Refactor `map.c` to do more in R/vctrs

### LINKED ISSUE #963: Use `length()` dispatch internally, not `Rf_xlength()`
We have an internal rlang function that does this (you wouldn't get access even if you import the whole rlang lib)
https://github.com/r-lib/rlang/blob/b3327104bac9c9fec90fa081fdd9cb321c3f29bb/src/internal/attr.c#L172

Used like this with `r_is_object()`
https://github.com/r-lib/rlang/blob/b3327104bac9c9fec90fa081fdd9cb321c3f29bb/src/internal/attr.c#L96-L101

``` r
library(clock)
x <- year_month_day(2019, 1:3)

purrr::map(x, identity)
#> $year
#> <year_month_day<month>[1]>
#> [1] "2019-01"
#> 
#> $month
#> <year_month_day<month>[1]>
#> [1] "2019-02"

lapply(x, identity)
#> [[1]]
#> <year_month_day<month>[1]>
#> [1] "2019-01"
#> 
#> [[2]]
#> <year_month_day<month>[1]>
#> [1] "2019-02"
#> 
#> [[3]]
#> <year_month_day<month>[1]>
#> [1] "2019-03"
```

<sup>Created on 2022-09-26 with [reprex v2.0.2.9000](https://reprex.tidyverse.org)</sup>

hadley:
Now fixes #963

### DIFF
--- a/R/map.R
+++ b/R/map.R
@@ -161,8 +161,10 @@ map_ <- function(.type, .x, .f, ..., .progress = FALSE, ..error_call = caller_en
   with_indexed_errors(
     i = i,
     error_call = ..error_call,
-    .Call(map_impl, environment(), .type, .progress, ..error_call)
+    out <- .Call(map_impl, environment(), .type, .progress, ..error_call)
   )
+
+  vec_set_names(out, vec_names(.x))
 }
 
 
--- a/R/map.R
+++ b/R/map.R
@@ -156,15 +156,14 @@ map_ <- function(.type, .x, .f, ..., .progress = FALSE, ..error_call = caller_en
   n <- vec_size(.x)
 
   .f <- as_mapper(.f, ...)
+  names <- vec_names(.x)
 
   i <- 0L
   with_indexed_errors(
     i = i,
     error_call = ..error_call,
-    out <- .Call(map_impl, environment(), .type, .progress, ..error_call)
+    .Call(map_impl, environment(), .type, .progress, ..error_call)
   )
-
-  vec_set_names(out, vec_names(.x))
 }
 
 
--- a/R/map.R
+++ b/R/map.R
@@ -153,6 +153,7 @@ map_chr <- function(.x, .f, ..., .progress = FALSE) {
 map_ <- function(.type, .x, .f, ..., .progress = FALSE, ..error_call = caller_env()) {
   .x <- vctrs_vec_compat(.x)
   vec_assert(.x, arg = ".x", call = ..error_call)
+  n <- vec_size(.x)
 
   .f <- as_mapper(.f, ...)
 
--- a/src/map.c
+++ b/src/map.c
@@ -41,46 +41,53 @@ SEXP call_loop(SEXP env, SEXP call, int n, SEXPTYPE type, int force_args,
 }
 
 SEXP map_impl(SEXP env, SEXP type_, SEXP progress, SEXP error_call) {
-  SEXP x = Rf_install(".x");
-  SEXP f = Rf_install(".f");
-  SEXP i = Rf_install("i");
-  SEXPTYPE type = Rf_str2type(CHAR(Rf_asChar(type_)));
+  static SEXP f_call = NULL;
+  if (f_call == NULL) {
+    SEXP x = Rf_install(".x");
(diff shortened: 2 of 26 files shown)


src/map.c
```diff
@@ -64,14 +64,7 @@ SEXP map_impl(SEXP env, SEXP type_, SEXP progress, SEXP error_call) {
...
 
   SEXP x_val = PROTECT(Rf_eval(x, env));
-
   int n = Rf_length(x_val);
-  if (n == 0) {
-    SEXP out = PROTECT(Rf_allocVector(type, 0));
-    copy_names(x_val, out);
-    UNPROTECT(2);
-    return out;
-  }
```
DavisVaughan, line 74: I was hoping we'd be able to do this!

tests/testthat/test-pmap.R
```diff
@@ -32,6 +32,16 @@ test_that("verifies result types and length", {
...
+  expect_equal(pmap(list(), identity), list())
+  expect_equal(pmap(NULL, identity), list())
+
+  expect_equal(pmap_lgl(NULL, identity), logical())
+})
+
+
 test_that("requires list of vectors", {
   expect_snapshot(error = TRUE, {
     pmap(environment(), identity)
```
DavisVaughan, line 47: Do you have any tests for simpler vectors also not being allowed? Like `pmap(character(), identity)`?

src/map.c
```diff
@@ -64,14 +64,7 @@ SEXP map_impl(SEXP env, SEXP type_, SEXP progress, SEXP error_call) {
...
 
   // Constructs a call like f(x[[i]], ...) - don't want to substitute
   // actual values for f or x, because they may be long, which creates
```
DavisVaughan, line 70: Note that line 110 has an outdated comment
https://github.com/tidyverse/purrr/pull/967/files#diff-67280b99316e68e37d45166adb44ae5ae41cf479e477b8582239d82cbc13070eR110

```
// Check all elements are lists and find recycled length
```

DavisVaughan, line 70: Unrelated comment

I had a thought that in line 112 of `pmap()`, rather than do this dance:

```c
int n = m == 0 ? 0 : Rf_length(VECTOR_ELT(l_val, 0));
```

you could just call this in `pmap_()` to get the common size, then pass it on to C

```r
size <- vec_size_common(!!!.l, .arg = ".l", .call = ..error_call)
.l <- vec_recycle_common(!!!.l, .arg = ".l", .call = ..error_call, .size = size)
```

DavisVaughan: Could also do a similar thing in `map2_()`

DavisVaughan: Oh hey if you also passed the size through in `map_()` then you'd avoid needing length-dispatch altogether!

(Probably still need names-dispatch)

DavisVaughan (approved):

src/map.c
hadley: Oh but we could also just set the names in `map_` etc.

tests/testthat/test-pmap.R
hadley: Yup — we test `pmap(environment(), identity)` (which is not exactly a simpler vector, but it's not a list)

src/map.c
hadley: Oh that's so much simpler 😆

DavisVaughan (approved):

R/map.R
```diff
@@ -153,15 +153,18 @@ map_chr <- function(.x, .f, ..., .progress = FALSE) {
 map_ <- function(.type, .x, .f, ..., .progress = FALSE, ..error_call = caller_env()) {
   .x <- vctrs_vec_compat(.x)
   vec_assert(.x, arg = ".x", call = ..error_call)
+  n <- vec_size(.x)
```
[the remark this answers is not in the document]
hadley: Isn't it better to be consistent with the naming convention within the function? (IMO it's also a bit misleading to call it size, because it is more like length due to `vctrs_vec_compat`)

PR #971, opened 2022-09-29
Cleanup C level `map()` code

DavisVaughan:
- Biggest change is no more `Rf_findVarInFrame()`, passing through explicit arguments instead, which feels less magical
- We also don't need to pass `error_call` through at all because we didn't actually use it in the C code (this is also nice because it avoids forcing the `error_call` promise unless it is actually needed)
- Unlinked from vctrs C API
- Modernized naming conventions and some syntax

Otherwise no change in behavior at all

### DIFF
--- a/src/map.c
+++ b/src/map.c
@@ -16,7 +16,6 @@ SEXP call_loop(SEXP env,
                int n,
                SEXP names,
                int* p_i,
-               SEXP error_call,
                int force) {
   SEXP bar = PROTECT(cli_progress_bar(n, progress));
 
@@ -55,8 +54,7 @@ SEXP map_impl(SEXP env,
               SEXP progress,
               SEXP ffi_n,
               SEXP names,
-              SEXP i,
-              SEXP error_call) {
+              SEXP i) {
   static SEXP call = NULL;
   if (call == NULL) {
     SEXP x_sym = Rf_install(".x");
@@ -87,7 +85,6 @@ SEXP map_impl(SEXP env,
     n,
     names,
     p_i,
-    error_call,
     force
   );
 }
@@ -97,8 +94,7 @@ SEXP map2_impl(SEXP env,
                SEXP progress,
                SEXP ffi_n,
                SEXP names,
-               SEXP i,
-               SEXP error_call) {
+               SEXP i) {
   static SEXP call = NULL;
   if (call == NULL) {
     SEXP x_sym = Rf_install(".x");
@@ -129,7 +125,6 @@ SEXP map2_impl(SEXP env,
     n,
     names,
     p_i,
-    error_call,
     force
   );
 }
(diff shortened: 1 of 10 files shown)


src/map.c
```diff
@@ -9,25 +9,30 @@
...
+               int* p_i,
+               int force) {
   SEXP bar = PROTECT(cli_progress_bar(n, progress));
 
   SEXP out = PROTECT(Rf_allocVector(type, n));
-  SEXP names_val = Rf_findVarInFrame(env, R_NamesSymbol);
-  Rf_setAttrib(out, R_NamesSymbol, names_val);
+  Rf_setAttrib(out, R_NamesSymbol, names);
 
   for (int i = 0; i < n; ++i) {
```
DavisVaughan, line 25: The previous overriding of `SEXP i` with the `int i = 0` counter in the loop was killing me 😬

src/map.c
```diff
@@ -37,77 +42,101 @@ SEXP call_loop(SEXP env, SEXP call, int n, SEXPTYPE type, int force_args,
...
-SEXP map_impl(SEXP env, SEXP type_, SEXP progress, SEXP error_call) {
-  static SEXP f_call = NULL;
-  if (f_call == NULL) {
-    SEXP x = Rf_install(".x");
-    SEXP f = Rf_install(".f");
-    SEXP i = Rf_install("i");
+SEXP map_impl(SEXP env,
+              SEXP ffi_type,
+              SEXP progress,
+              SEXP ffi_n,
```
DavisVaughan, line 55: We use this `ffi_` prefix in vctrs/rlang anytime we have an argument that comes from R that we need to massage into the right form for use in C. We prefer to give the C side the "better" name by prefixing the R side object with `ffi_`

src/map.h
```diff
@@ -2,8 +2,21 @@
...
+                SEXP i);
+
+  SEXP pmap_impl(SEXP env,
+                 SEXP ffi_type,
+                 SEXP progress,
+                 SEXP ffi_n,
+                 SEXP names,
+                 SEXP i,
+                 SEXP call_names,
+                 SEXP ffi_call_n);
```
DavisVaughan, line 19: I do not know what these do, I feel like we should remove them

src/map.c
jennybc: I just read encountered this `ffi_` stuff in vctrs. Does it stand for something?

DavisVaughan: Foreign function interface! https://en.wikipedia.org/wiki/Foreign_function_interface

In vctrs and rlang we go a step further by using `ffi_` in the bridge function's name too (i.e. then function itself on the C side would be `ffi_map()`). This works very well if you have a function that you'd like to give the same name on the R and C side, but you have to deal with some fiddly bits when going from the R function to the C function. For instance, in vctrs we have:
- R level `vec_as_names()`
- C level `ffi_vec_as_names(r_obj* names, r_obj* repair, r_obj* ffi_quiet, r_obj* frame)`. Deals with turning R level objects into C friendly `opts` struct
- C level `vec_as_names(r_obj* names, const struct name_repair_opts* opts)` which we call elsewhere in our C code

lionel- (approved):

hadley (approved):

PR #981, opened 2022-10-27
Add compat fallback for S4 objects

hadley:


R/utils.R
```diff
@@ -120,7 +120,11 @@ vctrs_vec_compat <- function(x) {
     )
     as.list(x)
   } else if (is.data.frame(x) || (is.list(x) && !vec_is(x))) {
-    unclass(x)
+    if (!isS4(x)) {
+      unclass(x)
+    } else {
+      set_names(x@.Data, names(x))
+    }
```
DavisVaughan, line 127: What is our stance on S4 objects in purrr in general? I assume it is more lax than our stance on them in vctrs.

My guess is that if the S4 object has a `length()` and `[[` method then it should be usable in purrr - the base R stance seems to be that if the S4 object has an `as.list()` method then it should work (see the `lapply()` definition), but lubridate types are a bit broken there so idk if we want to use that https://github.com/tidyverse/lubridate/issues/1080.

In that case, this wouldn't work, suggesting we probably need an S4 specific branch in `vctrs_vec_compat()`:

```r
library(purrr)

foo <- methods::setClass("foo", slots = c(x = "numeric", y = "numeric"))

methods::setMethod("length", "foo", function(x) {
  length(x@x)
})
methods::setMethod("[[", "foo", function(x, i, j, ..., exact = TRUE) {
  foo(x = x@x[[i]], y = x@y[[i]])
})

x <- foo(x = 1:5, y = 6:10)
x
#> An object of class "foo"
#> Slot "x":
#> [1] 1 2 3 4 5
#> 
#> Slot "y":
#> [1]  6  7  8  9 10

purrr::map(x, identity)
#> Error in `purrr::map()`:
#> ! `.x` must be a vector, not a <foo> object.
```

We could have a branch that looked like:

```r
else if (isS4(x)) {
  if (vec_is(x)) {
    x # like lubridate S4 types that define vec_proxy() and "just work"
  } else {
    vctrs_compat_chop(x)
  }
}
```

With:

``` r
vctrs_compat_chop <- function(x) {
  n <- length(x)
  out <- vector("list", length = n)
  
  for (i in seq_len(n)) {
    out[[i]] <- x[[i]]
  }
  
  names <- names(x)
  if (!is.null(names)) {
    names(out) <- names
  }
  
  out
}

x <- foo(x = 1:5, y = 6:10)
x
#> An object of class "foo"
#> Slot "x":
#> [1] 1 2 3 4 5
#> 
#> Slot "y":
#> [1]  6  7  8  9 10

vctrs_compat_chop(x)
#> [[1]]
#> An object of class "foo"
#> Slot "x":
#> [1] 1
#> 
#> Slot "y":
#> [1] 6
#> 
#> 
#> [[2]]
#> An object of class "foo"
#> Slot "x":
#> [1] 2
#> 
#> Slot "y":
#> [1] 7
#> 
#> 
#> [[3]]
#> An object of class "foo"
#> Slot "x":
#> [1] 3
#> 
#> Slot "y":
#> [1] 8
#> 
#> 
#> [[4]]
#> An object of class "foo"
#> Slot "x":
#> [1] 4
#> 
#> Slot "y":
#> [1] 9
#> 
#> 
#> [[5]]
#> An object of class "foo"
#> Slot "x":
#> [1] 5
#> 
#> Slot "y":
#> [1] 10
```

<sup>Created on 2022-10-31 with [reprex v2.0.2.9000](https://reprex.tidyverse.org)</sup>

hadley: Ah yeah, I can do that — I originally had `set_names(lapply(seq_along(x), function(i) x[[i]]), names(x))`.

This PR was mostly to fix a failing revdep, but I think it's better to take the more general approach.

DavisVaughan: It is actually worth noting that my `foo` example actually doesn't work with CRAN purrr either, but does work with `lapply()` so this will be a nice "fix"

hadley: Hmmm, so maybe it should just be `as.list(x)` to match what `lapply()` does?

hadley: Oooooh but that doesn't work if it doesn't have a `as.list` method  because it leaves it as an S4 object. 😬

PR #986, opened 2022-11-01
Move this script

jennybc:
Inspired by my contemplation of the purpose of `tools/` vs. `inst/tools/` for R Packages


hadley:
Thanks!

PR #988, opened 2022-11-01
When modifying, strip class from proxy

### LINKED ISSUE #987: Work around issue with sf subsetting
``` r
sf <- sf::st_read(system.file("shape/nc.shp", package="sf"))
#> Reading layer `nc' from data source 
#>   `/Users/hadleywickham/Library/R/arm64/4.2/library/sf/shape/nc.shp' 
#>   using driver `ESRI Shapefile'
#> Simple feature collection with 100 features and 14 fields
#> Geometry type: MULTIPOLYGON
#> Dimension:     XY
#> Bounding box:  xmin: -84.32385 ymin: 33.88199 xmax: -75.45698 ymax: 36.58965
#> Geodetic CRS:  NAD27
sf2 <- purrr::modify_at(sf, "NAME", toupper)
#> Warning in `[<-.data.frame`(`*tmp*`, .where, value = list(NAME = c("ASHE", :
#> provided 2 variables to replace 1 variables
```

<sup>Created on 2022-11-01 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

Can `unclass()` `vec_proxy()` since we're just using it for its "list"-ing properties.

hadley:
Fixes #987

### DIFF
--- a/tests/testthat/test-modify.R
+++ b/tests/testthat/test-modify.R
@@ -77,6 +77,20 @@ test_that("can modify non-vector lists", {
   expect_equal(modify2(notlist(1), list(3, 4), ~ .y), notlist(3, 4))
 })
 
+test_that("modifying data frame ignores [<- methods", {
+  df <- function(...) structure(data.frame(...), class = c("df", "data.frame"))
+  local_bindings(
+    "[<-.df" = function(...) stop("Forbidden"),
+    .env = globalenv()
+  )
+
+  x <- df(list(x = 1, y = "x"))
+  expect_equal(modify(x, ~ 2), df(x = 2, y = 2))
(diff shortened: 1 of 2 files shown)


tests/testthat/test-modify.R
```diff
@@ -77,6 +77,25 @@ test_that("can modify non-vector lists", {
   expect_equal(modify2(notlist(1), list(3, 4), ~ .y), notlist(3, 4))
 })
 
+test_that("modifying data frame ignores [<- methods", {
+  df <- function(...) {
+    structure(
+      data.frame(..., stringsAsFactors = FALSE),
```
DavisVaughan, line 83:
```suggestion
      data_frame(...),
```

You have access to the vctrs version too, if you want it

DavisVaughan (approved):

PR #989, opened 2022-11-01
Coercion deprecation improvements

### LINKED ISSUE #976: Expected behaviour for `map_*()` functions when working with factors
Hi,

I am playing around with the dev version of `purrr` and the new `map_vec()` function and the `map_*()` variants. Is the following expected behaviour?
``` r
library(purrr)
packageVersion('purrr')
#> [1] '0.9000.0.9000'

map_vec(1:10, factor)
#>  [1] 1  2  3  4  5  6  7  8  9  10
#> Levels: 1 2 3 4 5 6 7 8 9 10
map_lgl(1:10, factor)
#>  [1] TRUE TRUE TRUE TRUE TRUE TRUE TRUE TRUE TRUE TRUE
map_chr(1:10, factor)
#> Error in `map_chr()`:
#> ℹ In index: 1.
#> Caused by error in `deprecate_to_char()`:
#> ! could not find function "deprecate_to_char"

#> Backtrace:
#>     ▆
#>  1. ├─purrr::map_chr(1:10, factor)
#>  2. │ └─purrr:::map_("character", .x, .f, ..., .progress = .progress)
#>  3. │   └─purrr:::with_indexed_errors(...)
#>  4. │     └─base::withCallingHandlers(...)
#>  5. └─base::.handleSimpleError(...)
#>  6.   └─purrr (local) h(simpleError(msg, call))
#>  7.     └─cli::cli_abort(c(i = "In index: {i}."), parent = cnd, call = error_call)
#>  8.       └─rlang::abort(...)
map_int(1:10, factor)
#>  [1] 1 1 1 1 1 1 1 1 1 1
map_dbl(1:10, factor)
#>  [1] 1 1 1 1 1 1 1 1 1 1
```

hadley:
* Use `type` not `"type"`
* Call `purrr:::deprecate_to_char()`
* Capture correct environment

Fixes #976

### DIFF
--- a/R/coerce.R
+++ b/R/coerce.R
@@ -23,9 +23,9 @@ deprecate_to_char <- function(type) {
   )
 }
 
-local_deprecation_user_env <- function(frame = caller_env()) {
+local_deprecation_user_env <- function(user_env = caller_env(2), frame = caller_env()) {
   local_bindings(
-    deprecation_user_env = frame,
+    deprecation_user_env = user_env,
     .env = the,
     .frame = frame
   )
--- a/R/coerce.R
+++ b/R/coerce.R
@@ -7,15 +7,26 @@ coerce <- function(x, type) {
 coerce_lgl <- function(x) coerce(x, "logical")
 coerce_int <- function(x) coerce(x, "integer")
 coerce_dbl <- function(x) coerce(x, "double")
-coerce_chr <- function(x) coerce(x, "character")
+coerce_chr <- function(x) {
+  local_deprecation_user_env()
+  coerce(x, "character")
+}
 
 
-deprecate_to_char <- function(type, call = caller_env()) {
+deprecate_to_char <- function(type) {
   lifecycle::deprecate_warn(
     "1.0.0",
     I(paste0("Automatic coercion from ", type, " to character")),
     I("an explicit call to as.character() within map_chr()"),
     always = TRUE,
-    user_env = call
+    user_env = the$deprecation_user_env
+  )
+}
+
+local_deprecation_user_env <- function(frame = caller_env()) {
+  local_bindings(
+    deprecation_user_env = frame,
+    .env = the,
+    .frame = frame
   )
 }
--- a/R/coerce.R
+++ b/R/coerce.R
@@ -17,7 +17,7 @@ deprecate_to_char <- function(type) {
   lifecycle::deprecate_warn(
     "1.0.0",
     I(paste0("Automatic coercion from ", type, " to character")),
-    I("an explicit call to as.character() within map_chr()"),
+    I("an explicit call to `as.character()` within `map_chr()`"),
     always = TRUE,
     user_env = the$deprecation_user_env
   )
--- a/src/coerce.c
+++ b/src/coerce.c
@@ -62,14 +62,13 @@ double integer_to_real(int x) {
 }
 
 void deprecate_to_char(const char* type_char) {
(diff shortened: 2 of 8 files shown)


tests/testthat/_snaps/coerce.md
```diff
@@ -4,18 +4,25 @@
...
     Condition
       Warning:
-      Automatic coercion from type to character was deprecated in purrr 1.0.0.
+      Automatic coercion from double to character was deprecated in purrr 1.0.0.
       i Please use an explicit call to as.character() within map_chr() instead.
 
+# error captures correct env
+
+    Code
+      map_chr(1:4, identity)
```
hadley, line 25: I'm clearly not passing on the correct environments in `deprecate_to_char()` — what am I doing wrong?

src/coerce.c
```diff
@@ -61,15 +61,15 @@ double integer_to_real(int x) {
...
+void deprecate_to_char(const char* type_char) {
   SEXP env = PROTECT(caller_env());
+  SEXP type = PROTECT(Rf_mkString(type_char));
 
-  Rf_eval(call, env);
-  UNPROTECT(3);
+  SEXP fun = PROTECT(Rf_lang3(Rf_install(":::"), Rf_install("purrr"), Rf_install("deprecate_to_char")));
+  SEXP call = PROTECT(Rf_lang3(fun, type, env));
+
+  Rf_eval(call, R_GetCurrentEnv());
```
lionel-, line 71: If `R_GetCurrentEnv()` worked (it is not functional), it would return the same thing as `caller_env()` above.

lionel-: By the way in rlang it's called `r_peek_frame()`. It's returning the current frame on the stack so maybe rename `caller_env()` to `current_env()` to avoid confusion?

tests/testthat/_snaps/coerce.md
lionel-: The C function `caller_env()` implemented in purrr returns the current env. We could instead take the caller env from R (possibly with lazy assign) and pass that on through the C layer.

R/coerce.R
```diff
@@ -7,15 +7,26 @@ coerce <- function(x, type) {
...
     I(paste0("Automatic coercion from ", type, " to character")),
-    I("an explicit call to as.character() within map_chr()"),
+    I("an explicit call to `as.character()` within `map_chr()`"),
     always = TRUE,
-    user_env = call
+    user_env = the$deprecation_user_env
+  )
+}
+
+local_deprecation_user_env <- function(frame = caller_env()) {
```
lionel-, line 26: The user env is the caller of the `map_` function. You can take the grand-parent frame here or pass it through by argument:

```suggestion
local_deprecation_user_env <- function(user_env = caller_env(2), frame = caller_env()) {
```

tests/testthat/test-coerce.R
```diff
@@ -43,6 +43,10 @@ test_that("can coerce to character vectors", {
   expect_equal(coerce_chr("x"), "x")
 })
 
+test_that("error captures correct env", {
```
lionel-, line 46: To really test that the correct env is captured, I think this needs to simulate an indirect usage. Then the snapshot would show specific advice.

I think we should make it easier to test. Something like:

```r
expect_deprecated(foo(), type = c("any", "direct", "indirect"))
```

lionel- (approved):

PR #1011, opened 2022-12-18
Use .purrr_ prefix for internal helpers

### LINKED ISSUE #1002: `reduce_impl()` needs to name `error_call` something like `purrr_error_call`
Like with `map_(..error_call)`, we need to rename the `error_call` arg of `reduce_impl()` to something less liable to conflict with user arguments passed through `...`

In tidyr, with dev purrr installed some snapshot tests change and I see this incorrect error call of `fn()`

``` r
library(tidyr)

df <- tibble(x = list(1:2), y = list(1:3))

# Used to say `unchop()` in the call
unchop(df, c(x, y))
#> Error in `fn()`:
#> ! In row 1, can't recycle input of size 2 to size 3.

#> Backtrace:
#>     ▆
#>  1. └─tidyr::unchop(df, c(x, y))
#>  2.   └─tidyr:::df_unchop(cols, ptype = ptype, keep_empty = keep_empty)
#>  3.     └─purrr::reduce(x_sizes, unchop_sizes2, error_call = error_call)
#>  4.       └─purrr:::reduce_impl(.x, .f, ..., .init = .init, .dir = .dir)
#>  5.         └─tidyr (local) fn(out, elt, ...)
#>  6.           └─cli::cli_abort(...)
#>  7.             └─rlang::abort(...)
```

This is due to `df_unchop()` calling `reduce()` and passing `error_call` through, but it gets captured early by `reduce_impl()` and isn't passed through to `unchop_sizes2()`.

I suggest that we rename all of these cases to `purrr_error_call` and `.purrr_error_call` (in cases when we dot the args, for consistency) to eliminate potential conflicts and have a consistent way to handle this

We should probably see if there are other places this can come up

hadley:
To avoid inadvertent matching to user supplied arguments. Fixes #1002.

### DIFF
--- a/R/lmap.R
+++ b/R/lmap.R
@@ -61,7 +61,12 @@ lmap_at <- function(.x, .at, .f, ...) {
   lmap_helper(.x, where, .f, ...)
 }
 
-lmap_helper <- function(.x, .ind, .f, ..., .else = NULL, .purrr_error_call = caller_env()) {
+lmap_helper <- function(.x,
+                        .ind,
+                        .f,
+                        ...,
+                        .else = NULL,
+                        .purrr_error_call = caller_env()) {
   .f <- rlang::as_function(.f, call = .purrr_error_call)
   if (!is.null(.else)) {
     .else <- rlang::as_function(.else, call = .purrr_error_call)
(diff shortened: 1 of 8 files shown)


DavisVaughan (approved):
FWIW, I also did this in slider as needed https://github.com/r-lib/slider/commit/dccf974f4bbf4dd93952686f86cdcdab2b6a6405#diff-20a6ad8855fe7c996810d840857217f6fa537af586c0cc72be8f655db5125bcbR152

PR #1013, opened 2022-12-18
Check modify() gives good error with zap()

### LINKED ISSUE #1005: `zap()` and `modify_at`
``` r
library(purrr)
x3 <- list(a = 1, b = 2)
x3 |> 
  modify_at("b", \(x) zap()) |> 
  str()
#> List of 2
#>  $ a: num 1
#>  $ b: list()
#>   ..- attr(*, "class")= chr "rlang_zap"
```

<sup>Created on 2022-11-15 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>
should error?

hadley:
Fixes #1005

### DIFF
--- a/R/modify.R
+++ b/R/modify.R
@@ -174,7 +174,7 @@ imodify <- function(.x, .f, ...) {
 modify_where <- function(.x, .where, .f, ..., .error_call = caller_env()) {
   if (vec_is_list(.x)) {
     out <- vec_proxy(.x)
-    out[.where] <- map(out[.where], .f, ...)
+    out[.where] <- no_zap(map(out[.where], .f, ...), .error_call)
     vec_restore(out, .x)
   } else if (is.data.frame(.x)) {
     size <- vec_size(.x)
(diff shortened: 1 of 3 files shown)


R/modify.R
```diff
@@ -174,7 +174,7 @@ imodify <- function(.x, .f, ...) {
 modify_where <- function(.x, .where, .f, ..., .error_call = caller_env()) {
   if (vec_is_list(.x)) {
     out <- vec_proxy(.x)
-    out[.where] <- map(out[.where], .f, ...)
+    out[.where] <- no_zap(map(out[.where], .f, ...), .error_call)
```
DavisVaughan, line 177: Was this one of the places where the `.error_call` got prefixed? Just noting that this will need to be updated depending on the order you merge the PRs

R/modify.R
```diff
@@ -174,7 +174,7 @@ imodify <- function(.x, .f, ...) {
...
     vec_restore(out, .x)
   } else if (is.data.frame(.x)) {
```
DavisVaughan, line 179: Do you want to use `no_zap()` in this branch too for consistent errors? Right now the snapshot is telling me it looks like it errors in `vec_size()`, which is fine, at least it errors, but this might be clearer?

R/modify.R
```diff
@@ -187,7 +187,7 @@ modify_where <- function(.x, .where, .f, ..., .error_call = caller_env()) {
     .x[.where] <- map_vec(.x[.where], .f, ..., .ptype = .x)
```
DavisVaughan, line 187: Can't really wrap this one in `no_zap()`, but I think that is fine. I wouldn't want to slow down `map_vec()` by trying to push `no_zap()` into it. The snapshot error is also pretty good

DavisVaughan (approved):

R/modify.R
hadley: No, but it probably should've been 😬

PR #1014, opened 2022-12-18
Add .is_node to list_modify() and friends

### LINKED ISSUE #1003: Add `.is_node` argument to `list_modify()` and friends

hadley:
Fixes #1003

### DIFF
--- a/R/list-modify.R
+++ b/R/list-modify.R
@@ -17,6 +17,7 @@
 #'   [Dynamic dots][rlang::dyn-dots] are supported. In particular, if your
 #'   replacement values are stored in a list, you can splice that in with
 #'   `!!!`.
+#' @inheritParams map_depth
 #' @export
 #' @examples
 #' x <- list(x = 1:10, y = 4, z = list(a = 1, b = 2))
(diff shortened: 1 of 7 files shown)


R/list-modify.R
```diff
@@ -42,32 +43,34 @@
...
-
   y <- dots_list(..., .named = NULL, .homonyms = "error")
-  list_recurse(.x, y, c)
+
+  list_recurse(.x, y, c, is_node = .is_node)
 }
 
-list_recurse <- function(x, y, base_f, recurse = TRUE, error_call = caller_env()) {
+list_recurse <- function(x, y, base_f, recurse = TRUE, error_call = caller_env(), is_node = NULL) {
+  is_node <- as_is_node(is_node, error_call, ".is_node")
```
DavisVaughan, line 72: Maybe name the optional arguments here? `".is_node"` is the main one that took me a second to understand

DavisVaughan, line 72: Is there a reason the implementation of `as_is_node` wraps `vec_is_list()` in an anonymous function rather than just returning `vec_is_list`?

tests/testthat/test-list-modify.R
```diff
@@ -80,6 +80,10 @@ test_that("but data.frames are not", {
   x2 <- list(x = data.frame(y = 2))
   out <- list_modify(x1, !!!x2)
   expect_equal(out, x2)
+
+  # unless you really want it
+  out <- list_modify(x1, !!!x2, .is_node = is.list)
+  expect_equal(out, list(x = data.frame(x = 1, y = 2)))
```
DavisVaughan, line 86: Can also add tests for `list_merge()` and `list_update()` if you want to

DavisVaughan (approved):

R/list-modify.R
hadley: Hmm, I think that's just a remnant of an older approach where I was worried about bundling it in the package. I think it can be removed now.

tests/testthat/test-list-modify.R
hadley: What if I don't want? 😛

PR #1015, opened 2022-12-19
Pass `user_env` through helpers to `deprecate_soft()`

### LINKED ISSUE #1004: Soft deprecation warnings of `vctrs_vec_compat()` don't bubble up
i.e. should see a warning here

``` r
purrr::map(pairlist(1, 2), identity)
#> [[1]]
#> [1] 1
#> 
#> [[2]]
#> [1] 2
```

I think we need to pass the `user_env` down to `vctrs_vec_compat()`, which may be annoying

grep for any usage of `deprecate_soft()` to make sure we get them all

hadley:
Fixes #1004

### DIFF
--- a/R/deprec-rerun.R
+++ b/R/deprec-rerun.R
@@ -35,7 +35,7 @@
 #'   map(\(i) list(x = rnorm(5), y = rnorm(5))) |>
 #'   map_dbl(\(l) cor(l$x, l$y))
 rerun <- function(.n, ...) {
-  deprec_rerun(.n, ...)
+  deprec_rerun(.n, ..., .purrr_user_env = caller_env())
 
   dots <- quos(...)
 
@@ -55,7 +55,7 @@ rerun <- function(.n, ...) {
   out
 }
 
-deprec_rerun <- function(.n, ...) {
+deprec_rerun <- function(.n, ..., .purrr_user_env) {
   n <- .n
   old <- substitute(rerun(n, ...))
   if (dots_n(...) == 1) {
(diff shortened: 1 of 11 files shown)


R/utils.R
```diff
@@ -1,4 +1,8 @@
-where_at <- function(x, at, error_arg = caller_arg(at), error_call = caller_env()) {
+where_at <- function(x,
+                     at,
+                     user_env,
```
DavisVaughan, line 3: It is a bit of a tough call, but when I add `user_env` to a helper that I know I'm going to be testing directly, I do `user_env = caller_env()` so that the user environment is "right" when `where_at()` is called directly at the top level (i.e. in a test). That way you don't have to touch the tests. This would also apply to `vctrs_vec_compat()`, which you also tweaked tests for, and then you may as well do it for `vctrs_list_compat()` too (if you decide to do it at all)

R/deprec-rerun.R
```diff
@@ -55,7 +55,7 @@ rerun <- function(.n, ...) {
   out
 }
 
-deprec_rerun <- function(.n, ...) {
+deprec_rerun <- function(.n, ..., .purrr_user_env) {
```
DavisVaughan, line 58:
```suggestion
deprec_rerun <- function(.n, ..., .purrr_user_env = caller_env(2)) {
```

Probably reasonable here since it is only called in one place? Then you don't need to set it above

R/utils.R
DavisVaughan: This would _not_ fix the problem of having to set `user_env = current_env()` at the call sites of `where_at()` in the source code itself, but that seemed ok by me vs having to make the tests obscure looking

R/map2.R
```diff
@@ -60,9 +60,10 @@ map2_ <- function(.type,
                   .f,
                   ...,
                   .progress = FALSE,
+                  .purrr_user_env = caller_env(2),
                   .purrr_error_call = caller_env()) {
-  .x <- vctrs_vec_compat(.x)
-  .y <- vctrs_vec_compat(.y)
+  .x <- vctrs_vec_compat(.x, .purrr_user_env)
+  .y <- vctrs_vec_compat(.y, .purrr_user_env)
```
DavisVaughan, line 66: If you do make `user_env` an optional arg here by giving it a default, I would then lean towards naming it explicitly here

R/utils.R
DavisVaughan: Maybe using `user_env = caller_env()` is the answer to your question about how to test soft-deprecation 😅 , because that makes it work without any further tweaks

DavisVaughan (approved):
Approving but expecting that you'll probably have a comment or two about my `user_env = caller_env()` suggestion below

PR #1017, opened 2022-12-20
Set `Biarch: true` for old R versions on Windows

DavisVaughan:
Now that we have a `configure.win` script, we are forced to also set `Biarch: true` so that when purrr is installed from source on R < 4.2, it installs for both 32 and 64 bit Windows.

Otherwise, we get failures like this one in tidyr on our Windows 3.6 builds if we request `purrr (>= 1.0.0)`:
https://github.com/tidyverse/tidyr/actions/runs/3743952288/jobs/6356732834

```
*** arch - i386
Error: Error: package or namespace load failed for 'tidyr' in library.dynam(lib, package, package.lib):
 DLL 'purrr' not found: maybe not installed for this architecture?
```

This doesn't happen on newer versions of R because R 4.2 dropped support for 32 bit Windows. This is why windows-latest passes on tidyr.

See 6.3.4 of https://cran.r-project.org/doc/manuals/r-patched/R-admin.html for more details.

```
Windows
where there is a non-empty configure.win script, or a file src/Makefile.win (with some exceptions where the package is known to have an architecture-independent configure.win, or if --force-biarch or field ‘Biarch’ in the DESCRIPTION file is used to assert so).

In those cases only the current architecture is installed.
```

This might be important enough for a patch purrr release.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -34,6 +34,7 @@ LinkingTo:
     cli
 VignetteBuilder: 
     knitr
+Biarch: true
 Config/Needs/website: tidyverse/tidytemplate
 Config/testthat/edition: 3
 Encoding: UTF-8


hadley:
I'm slightly tempted to leave this for a bit to see if anyone complains about it.

hadley (approved):

DavisVaughan:
Me and my Windows 3.6 tidyr builds will complain loudly 🤣 

More seriously I imagine there are some corporate businesses stuck on older versions of R that are also stuck on 32-bit R that might come out and complain. I remember 32 bit R was needed to connect to an Access database (all too common)

lorenzwalthert:
[CRAN checks](https://cloud.r-project.org/web/checks/check_results_styler.html) of packages like {styler} also complain. And [CI/CD](https://github.com/r-lib/styler/actions/runs/3771984629/jobs/6412780732#step:6:142).

dieghernan:
Hi, this is causing errors in CRAN for a bunch of packages:
- tidyterra: https://cloud.r-project.org/web/checks/check_results_tidyterra.html
- arrow: https://cran.r-project.org/web/checks/check_results_arrow.html https://github.com/apache/arrow/issues/15080
- tune: https://cran.r-project.org/web/checks/check_results_tune.html https://github.com/tidymodels/tune/pull/595

A quick search on GitHub: https://github.com/search?q=%22DLL+purrr+not+found%3A+maybe+not+installed+for+this+architecture%3F%22&type=issues

hadley:
We will submit a patch to CRAN in early January when we are all back at work.

PR #1026, opened 2022-12-22
Add cleancall support to close progress bars on early exits

### LINKED ISSUE #1024: Persistent progress_bar after interrupting `map` loop in the middle
I'm using it in a personal function, but will try to reproduce it asap.
I run the function, progress bars appears and everything works fine (it even disappears as expected).
But if I interrupt this map loop with `Esc`, the progress bar keeping appearing after every code I ran

DavisVaughan:
Closes #1024 

Fairly certain this is correct. It does seem to fix the problem from #1024, both with hitting `Esc` and my `stop()` example.

Had to use the `cleancall.c/h` from {cli} to get access to `R_CLEANCALL_SUPPORT` because I think `r-lib/cleancall` is a little out of date.

Otherwise, the instructions at https://github.com/r-lib/cleancall#embedding-cleancall worked well.

As @gaborcsardi mentioned here https://github.com/tidyverse/purrr/issues/1024#issuecomment-1363028007, this doesn't require any changes to the purrr code itself as long as we include `cleancall.h` before `cli/progress.h`, because `cli/progress.h` knows how to set `cli_progress_bar()` up with an early exit handler using cleancall if it detects that the cleancall header is included.
https://github.com/r-lib/cli/blob/bc503509cddb65d0007aeb301936e27de76bc7d7/inst/include/cli/progress.h#L298-L301

### DIFF
--- a/src/map.c
+++ b/src/map.c
@@ -2,12 +2,20 @@
 #include <R.h>
 #include <Rversion.h>
 #include <Rinternals.h>
-#include "cleancall.h"
 #include "coerce.h"
 #include "conditions.h"
 #include "utils.h"
 
-#include "cli/progress.h"
+// Including <cli/progress.h> before "cleancall.h" because we want to register
+// exiting handlers ourselves, rather than letting cli register them for us.
+#include <cli/progress.h>
+#include "cleancall.h"
+
+static
+void cb_progress_done(SEXP bar) {
+  cli_progress_done(bar);
+  R_ReleaseObject(bar);
+}
 
 // call must involve i
 SEXP call_loop(SEXP env,
@@ -18,7 +26,9 @@ SEXP call_loop(SEXP env,
                SEXP names,
                int* p_i,
                int force) {
-  SEXP bar = PROTECT(cli_progress_bar(n, progress));
+  SEXP bar = cli_progress_bar(n, progress);
+  R_PreserveObject(bar);
+  r_call_on_exit((void (*)(void*)) cb_progress_done, (void*) bar);
 
   SEXP out = PROTECT(Rf_allocVector(type, n));
   Rf_setAttrib(out, R_NamesSymbol, names);
@@ -44,9 +54,8 @@ SEXP call_loop(SEXP env,
   }
 
   *p_i = 0;
-  cli_progress_done(bar);
 
-  UNPROTECT(2);
+  UNPROTECT(1);
   return out;
 }
 
--- a/src/map.c
+++ b/src/map.c
@@ -2,6 +2,7 @@
 #include <R.h>
 #include <Rversion.h>
 #include <Rinternals.h>
+#include "cleancall.h"
 #include "coerce.h"
 #include "conditions.h"
 #include "utils.h"
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # purrr (development version)
 
+* Fixed an issue where progress bars weren't being closed when user interrupts
+  or errors were encountered during a `map()` call (#1024).
+
 * Fixed an invalid C signature for `pluck()` (#1018).
 
 * Set `Biarch: true` to build purrr on 32-bit Windows on R < 4.2.0 (#1017).
--- a/R/cleancall.R
+++ b/R/cleancall.R
@@ -1,0 +1,3 @@
+call_with_cleanup <- function(ptr, ...) {
+  .Call(cleancall_call, pairlist(ptr, ...), parent.frame())
+}
--- a/R/map.R
+++ b/R/map.R
@@ -171,7 +171,7 @@ map_ <- function(.type,
   with_indexed_errors(
     i = i,
     error_call = .purrr_error_call,
-    .Call(map_impl, environment(), .type, .progress, n, names, i)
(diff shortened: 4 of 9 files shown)


gaborcsardi (approved):
Thanks much! Indeed it looks correct, assuming this is the direction we are going. 

I have used cleancall virtually unchanged in smaller packages for years, but obviously in purrr it will be exposed to a lot more...

Nevertheless @lionel- wrote the cleancall internals, which helps with the trust. :)

DavisVaughan:
@lionel- do you feel comfortable using cleancall in vctrs? I'm also going to ask you for a review here.

DavisVaughan:
I'm a bit worried about the fact that `<cli/progress.h>` requires that purrr needs to use a very specific version of `cleancall.{c,h}` that matches whatever version cli was using.

I'm worried that if `cleancall.{c,h}` evolve at all, then if purrr tries to update to the latest version of those files, it won't be able to if cli isn't updated first. Or, even worse, cli might update its `progress.h` API to expect the latest version of the `cleancall.{c,h}` files, which might not be compatible with whatever versions are already on CRAN with purrr.

For example, cli expects that `cleancall.h` exposes `r_cleancall_is_active()`, which cli added to its custom version of `cleancall.{c,h}`. I don't _think_ that function is needed, so I was going to remove it from the cli and purrr versions of `cleancall.{c,h}` and from where cli tries to use it in `<cli/progress.h>`, but since CRAN cli expects that function to be available, if we tried to send purrr in with this updated version of cleancall then it wouldn't compile. 
https://github.com/r-lib/cli/blob/bc503509cddb65d0007aeb301936e27de76bc7d7/inst/include/cli/progress.h#L299

I wonder if cli should somehow vendor its own version of `cleancall.h` in `inst/include/cli/cleancall.h` that purrr would include before `cli/progress.h`. That way cli would be in charge of keeping them in sync. It would also mean purrr would have to use the cli version of `cleancall_call()` that would get called from R in place of `.Call()`, I'm not sure how I feel about that part.

gaborcsardi:
I think as long as the API of cleancall does not change, this should work.

But we could also implement the exit handlers in purrr, and then there is no need for cleancall.

OTOH it would be great to have a general solution to this, so we can use exit handlers in C/C++ APIs.

DavisVaughan:
It might be cleaner if we all collectively decide to use `LinkingTo` to access cleancall rather than embedding it.

That way we make any changes directly in `r-lib/cleancall`, which will probably force us to be more careful about making API changes, and probably gives us an easier to way revdep test any API changes against "reverse LinkingTo" deps

gaborcsardi:
I think embedding is actually more stable, as long as the API is the same, and it is unlikely that we would change the API.

The check for whether cleancall is active or not is needed in case a plain .Call is used instead of call_with_cleanup().

gaborcsardi:
> In this case I think a better solution would be to add a function to the cli API that also takes two pointers to a callback address and its data. If not NULL, cli would set these pointers to a cleanup handler. Then the caller is in charge of calling that handler, with or without cleancall, vendored or not.

I don't understand how this would work. What does cli do with the callback address and its data? cli does not have any cleanup handlers. The cleanup handlers must be set by the package that uses `.Call()`. Who is "the caller"? purrr or the user?

In any case, I agree that it is certainly cleaner if purrr itself takes care of the cleanup. Then purrr should not use cleancall, but implement its own cleanup functions, and register an exit handler every time it creates a progress bar. I.e. something like:

```r
  SEXP bar = PROTECT(cli_progress_bar(n, progress));
  r_call_on_early_exit((void (*)(void *)) cli_progress_done2, (void*) bar);
  ...
```

lionel-:
oh I had misunderstood how this is set up.

If all that is needed is to call a cleanup function that is statically available, with data that is in hands of the caller, then we should do that.

Gabor do I understand correctly that we can do this without any changes to cli by:

- Including `cli/progress.h` _before_ including `cleancall.h`
- Using cleancall to call `cli_progress_done()` on exit

gaborcsardi:
Yes, cli cannot call a cleanup function because there is no cli function on the stack when the cleanup is needed.

> Gabor do I understand correctly that we can do this without any changes to cli by:
> 
> - Including `cli/progress.h` _before_ including `cleancall.h`
> - Using cleancall to call `cli_progress_done()` on exit

Yes. Maybe it is better to explicitly unset `R_CLEANCALL_SUPPORT` before you include `progress.h` and then cli will not add any exit handlers for sure. And you can add them manually, with cleancall embedded into purrr.

(This is exactly how I meant this to work, `R_CLEANCALL_SUPPORT` is opt-in, basically.)

src/map.c
```diff
@@ -17,7 +26,9 @@ SEXP call_loop(SEXP env,
                SEXP names,
                int* p_i,
                int force) {
-  SEXP bar = PROTECT(cli_progress_bar(n, progress));
+  SEXP bar = cli_progress_bar(n, progress);
+  R_PreserveObject(bar);
```
gaborcsardi, line 30: I don't think this is needed.

DavisVaughan:
@hadley we have agreed upon the current implementation (with some room for improvement in cleancall in the future to avoid the `R_PreserveObject()` call).

If you approve, I will rebase and merge

hadley (approved):

PR #1030, opened 2022-12-27
Generate man files via Makefile, not configure

jeroen:
Just a suggestion: it is more robust and portable to run this as a `make` step rather than introduce a configure script.

Configure should only be used to query and set platform specific things (which is why you ran into the biarch problem). What you are doing here does not require configure, it is the same on any OS. There is no need to depend on a shell script for it, and you can avoid a lot of issues by not introducing one (many windows users don't have a shell on the PATH, etc).


gaborcsardi:
The reason for `./configure` was that I wanted something that works for packages without compiled code.

> There is no need to depend on a shell script for it, and you can avoid a lot of issues by not introducing one (many windows users don't have a shell on the PATH, etc).

Rtools has a shell and afair puts it on the path, and we already require Rtools, at least for packages that have compiled code.

src/Makevars
```diff
@@ -1 +1,6 @@
 PKG_CFLAGS = $(C_VISIBILITY)
+
+all: ../man/macros/examples.Rd
```
gaborcsardi, line 3: Don't you need to include the DLL here, i.e. `$(SHLIB)`?

jeroen:
OK makes sense, was just a suggestion to avoid shell trouble. I did notice that `biach` is now [deprecated](https://github.com/r-devel/r-svn/commit/d76957f58069c3a5260863200b73a1cb4aff90c2), but I think we're still allowed to use it in packages to support older R versions. Either way I'll close the suggestion for now.

PR #1038, opened 2023-01-04
Use names in indexed error

hadley:


tests/testthat/_snaps/map.md
```diff
@@ -38,6 +38,23 @@
       Caused by error in `fail_at_3()`:
       ! Doesn't work
 
+# error location uses name if present
+
+    Code
+      map_int(c(a = 1, b = 2, c = 3), ~ fail_at_3(.x, stop("Error")))
+    Condition
+      Error in `map_int()`:
+      i In index: c.
```
hadley, line 47: Having written this, it feels to me like this should really be `In index: c [3]`, i.e. you want to show both name and position, so I wonder if we actually want to include both `position` and `name` in the error object.

lionel-: oh good point. That makes sense, only locations are unambiguous.

`c [3]` doesn't feel very clear though. Other ideas:

```
i In index: 3 (with name `c`).
```

```
i In index: 3.
i With name: "c".
```

hadley: I like the first option. Maybe `In index 3, named "c"`?

hadley:
Closing; to be integrated into #1034.

PR #1117, opened 2024-02-01
Update list-modify.R

asadow:
Consistent formatting of line breaks and more clear use of list_modify() - it was easy to miss the change from list_assign() previously


asadow:
Ooph, did not mean to set off a whole bunch of GHA's with this small adjustment.

hadley:
Thank you! (And no need to worry about the number of GHA jobs; that's usual 😄)

PR #1123, opened 2024-03-22
Update reduce.R example using stochastic simulation section with corr…

hathawayj:
`list_rbind()` has the old argument of `id`.  Fixed it to have `list_rbind(names_to = "simulation")`


hadley:
Thank you!

PR #1140, opened 2024-08-15
Add example to map_dfr()/map_dfc()

### LINKED ISSUE #1074: map_dfr
`map_dfr` has been deprecated and the help file suggests using `map` with `list_rbind` or `list_cbind` but that means that we need two commands where we used to get away with one. `map_dfr` was probably the most frequently used command in purrr at least for me.  The reason to deprecate it cited is inconsistency of the name but if that is the only problem we could have `map_dfr` under another name.  Also `map(...) %>% list_cbind` and `map(...) %>% list_cbind` do not even work below whereas `map_dfr` does -- maybe I am doing this wrong but it seems that this is what the help file is suggesting.

```R
# https://stackoverflow.com/questions/76094233/remove-all-non-number-characters-from-every-cell-in-a-df/76094508#76094508
library(purrr)

DF <- structure(list(`Column A` = "dfd:13", `Column B` = "hhh:34", 
    `Column C` = "dsd:15", `Column D` = "ffd:67", `Column E` = "hdf:89", 
    `Column F` = "hhj:43"), class = "data.frame", row.names = c(NA, 
-1L))

DF %>%
  map(trimws, whitespace = "\\D") %>%
  list_cbind
## Error in `list_cbind()`:
## ! Each element of `x` must be either a data frame or `NULL`.
## ℹ Elements 1, 2, 3, 4, 5, and 6 are not.
## Run `rlang::last_trace()` to see where the error occurred.

DF %>%
  map(trimws, whitespace = "\\D") %>%
  list_rbind
## Error in `list_rbind()`:
## ! Each element of `x` must be either a data frame or `NULL`.
## ℹ Elements 1, 2, 3, 4, 5, and 6 are not.
## Run `rlang::last_trace()` to see where the error occurred.

# desired result
DF %>%
  map_dfr(trimws, whitespace = "\\D")
## # A tibble: 1 × 6
##   `Column A` `Column B` `Column C` `Column D` `Column E` `Column F`
##   <chr>      <chr>      <chr>      <chr>      <chr>      <chr>     
## 1 13         34         15         67         89         43   
```

maike2011:
Example to demonstrate an edge case, where map_dfr() works, but can not be replaced by list_rbind(). Fixes #1074

### DIFF
--- a/R/superseded-map-df.R
+++ b/R/superseded-map-df.R
@@ -41,6 +41,32 @@
 #'   map(\(mod) as.data.frame(t(as.matrix(coef(mod))))) |>
 #'   list_rbind()
 #'
+#' # map: combining vectors ---------------------------
+#' DF <- structure(
+#' list(
+#'   `Column A` = c(" 13", "  15 "),
+#'   `Column B` = c("  34",  " 67 ")
+#' ),
+#' class = c("data.frame"),
+#' row.names = c(NA, -2L)
+#' )
+#'
+#' # Was:
+#' # map_dfr() and map_dfc() actually both combine the list by column
+#' map_dfr(DF, trimws)
+#' map_dfc(DF, trimws)
+#'
+#' # Now:
+#' # to combine a list of vectors use as_tibble()
+#' map(DF, trimws) |> tibble::as_tibble()
+#'
+#' # list_rbind()/list_cbind() require list of data.frames or NULL to work and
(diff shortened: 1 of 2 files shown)


hadley:
Thanks for working on this! While looking at this with fresh eyes, I realised that an even simpler approach is to use `modify()`.

PR #1142, opened 2024-08-15
expect_error -> expect_snapshot

### LINKED ISSUE #1132: Convert `expect_error()` to `expect_snapshot()`
Since snapshots give higher fidelity and also report on calls.

As part of this issue you could also fix any problems that you find, or you could just file a new issue.

catalamarti:
fixes #1132

### DIFF
--- a/tests/testthat/test-adverb-auto-browse.R
+++ b/tests/testthat/test-adverb-auto-browse.R
@@ -1,4 +1,4 @@
 test_that("auto_browse() not intended for primitive functions", {
   expect_snapshot(auto_browse(log)(NULL), error = TRUE)
-  expect_error(auto_browse(identity)(NULL), NA)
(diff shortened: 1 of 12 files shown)


tests/testthat/test-adverb-auto-browse.R
```diff
@@ -1,4 +1,4 @@
 test_that("auto_browse() not intended for primitive functions", {
   expect_snapshot(auto_browse(log)(NULL), error = TRUE)
-  expect_error(auto_browse(identity)(NULL), NA)
+  expect_snapshot(auto_browse(identity)(NULL))
```
lionel-, line 3: Here we are not testing an error message but the absence of error so this is clearer about the intention:

```suggestion
  expect_no_error(auto_browse(identity)(NULL))
```

lionel- (approved):
Looks great!

PR #1143, opened 2024-08-15
Correct documentation for pmap

### LINKED ISSUE #1124: pmap output names help file
The help file for `pmap`  says 

> The  output names are determined by the input names. 

but the names of BOD are `c("Time", "demand")` and the output has no names.  There were some SO posts guessing that it means the names of the first column of the input  (https://stackoverflow.com/questions/78310377/name-output-of-pmap-on-tibble/78310769#78310769) or the first column with names in the input (https://stackoverflow.com/questions/78314849/naming-of-elements-in-purrrmap).  Please clarify.

```
str(pmap(BOD, ~ .x + .y))
## List of 6
## $ : num 9.3
## $ : num 12.3
## $ : num 22
## $ : num 20
## $ : num 20.6
## $ : num 26.8
```

SoyAndrea:
Fixes #1124

change help file to pmap because the behaviour of map is slightly different from pmap

### DIFF
--- a/R/pmap.R
+++ b/R/pmap.R
@@ -23,7 +23,25 @@
 #'   * A formula, e.g. `~ ..1 + ..2 / ..3`. This syntax is not recommended as
 #'     you can only refer to arguments by position.
 #' @inheritParams map
-#' @inherit map return
+#' @returns
+#' The output length is determined by the length of the input.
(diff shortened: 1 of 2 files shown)


hadley:
Thank you! I made a couple more tweaks based on my priviledged knowledge of `pmap()` 😄

PR #1144, opened 2024-08-15
Adding keep_empty argument to list_c, list_cbind, and list_rbind. 

### LINKED ISSUE #1096: FR: New `keep_empty = FALSE` argument to `list_c()` and `list_rbind()`
I'm looking for the following behavior so that I can use `pivot_wider() %>% mutate(across(..., ~ list_c(.x, keep_empty = TRUE)))` as a more robust replacement for `tidyr::spread(convert = TRUE)` .

I can achieve it today with the lower-level `list_unchop()`, in a very clumsy way.

``` r
library(purrr)

# Existing behavior:
list_c(list("a", NULL, "b"))
#> [1] "a" "b"
vctrs::list_unchop(list("a", NULL, "b"))
#> [1] "a" "b"
vctrs::list_unchop(list("a", NULL, "b"), indices = as.list(1:3))
#> [1] "a" NA  "b"

# I'd like to write this:
list_c(list("a", NULL, "b"), keep_empty = TRUE)
#> [1] "a" NA  "b"
```

<sup>Created on 2023-08-20 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

SokolovAnatoliy:
Fixes #1096.


NEWS.md
```diff
@@ -1,5 +1,7 @@
 # purrr (development version)
 
+* Added a new keep_empty argument to list_c(), list_cbind() and list_rbind() which will keep empty elements as NA in the returned dataframe. (#1096)
```
hadley, line 3:
```suggestion
* Added a new keep_empty argument to `list_c()`, `list_cbind()`, and `list_rbind()` which will keep empty elements as `NA` in the returned data frame. (#1096)
```

R/list-combine.R
```diff
@@ -22,6 +22,7 @@
 #'   same size (i.e. number of rows).
 #' @param name_repair One of `"unique"`, `"universal"`, or `"check_unique"`.
 #'   See [vctrs::vec_as_names()] for the meaning of these options.
+#' @param keep_empty An optional Logical to keep empty elements of a list as NA.
```
hadley, line 25: Could you try rewriting this using the formula "If `FALSE` (the default), then ...; if `TRUE`, then ...`.

R/list-combine.R
```diff
@@ -30,17 +31,21 @@
 #'
 #' x2 <- list(
 #'   a = data.frame(x = 1:2),
-#'   b = data.frame(y = "a")
+#'   b = data.frame(y = "a"),
+#'   c = data.frame(z = NULL)
```
hadley, line 35: I think this is equivalent to just `data.frame()`?

R/list-combine.R
```diff
@@ -58,19 +63,22 @@ list_cbind <- function(
     x,
     ...,
     name_repair = c("unique", "universal", "check_unique"),
-    size = NULL
+    size = NULL,
+    keep_empty = FALSE
   ) {
   check_list_of_data_frames(x)
   check_dots_empty()
+  if(keep_empty) x <- convert_empty_element_to_NA(x)
```
hadley, line 71: Are you sure this applies to `list_cbind()`? I don't know why but I think it shouldn't?

R/list-combine.R
```diff
@@ -95,3 +103,8 @@ check_list_of_data_frames <- function(x, error_call = caller_env()) {
     call = error_call
   )
 }
+
+## used to convert empty elements into NA for list_binding functions
+convert_empty_element_to_NA = function(x) {
+  map(x, \(x) if(vctrs::vec_is_empty(x)) NA else x)
```
hadley, line 109: I think this anonymous function syntax requires R 4.1? Otherwise, this implementation looks good to me!

tests/testthat/test-list-combine.R
```diff
@@ -70,6 +70,15 @@ test_that("NULLs are ignored", {
   expect_equal(list_cbind(list(df1, NULL, df2)), vec_cbind(df1, df2))
 })
 
+test_that("NULLs are converted to NA when keep_empty = TRUE", {
+  df1 <- data.frame(x = 1)
+  df2 <- data.frame(y = 1)
+
+  expect_equal(list_c(list(1, NULL, 2), keep_empty = TRUE), c(1, NA, 2))
+  expect_equal(list_rbind(list(df1, NULL, df1), keep_empty = TRUE), vec_rbind(df1, NA, df1))
```
hadley, line 78: I think I'd find it a little easier to verify that this is correct if the expected value looked like `data.frame(x = c(1, NA, 1)`

hadley (commented):
Thanks for working on this!

R/list-combine.R
```diff
@@ -95,3 +105,8 @@ check_list_of_data_frames <- function(x, error_call = caller_env()) {
     call = error_call
   )
 }
+
+## used to convert empty elements into NA for list_binding functions
+convert_empty_element_to_NA = function(x) {
+  map(x, function(x) if(vctrs::vec_is_empty(x)) NA else x)
```
krlmlr, line 111: Should that be something like

```suggestion
  map(x, function(x) if(is.null) NA else x)
```

?

R/list-combine.R
krlmlr: To me, it's unclear what type the empty column should have.

NEWS.md
krlmlr: If you like, you can add your GitHub handle after the issue: `(#1096, @SokolovAnatoliy)` .

krlmlr (commented):
Thanks for taking this on!

R/list-combine.R
mgirlich, line 111: This might have quite a performance impact if `x` is big. A faster alternative would be something like

```r
empty <- vctrs::list_sizes(x) == 0
x[empty] <- NA
x
```

R/list-combine.R
hadley: It's also weird that for `list_cbind()` the name of empty columns comes from the outer name rather than the inner name.

R/list-combine.R
```diff
@@ -58,19 +66,26 @@ list_cbind <- function(
     x,
     ...,
     name_repair = c("unique", "universal", "check_unique"),
-    size = NULL
+    size = NULL,
+    keep_empty = FALSE
```
hadley, line 70: @DavisVaughan my intuition is that `keep_empty` doesn't make sense for `list_cbind()`, but I don't have anything concrete to back that up. Do you have any thoughts?

R/list-combine.R
```diff
@@ -22,6 +22,9 @@
 #'   same size (i.e. number of rows).
 #' @param name_repair One of `"unique"`, `"universal"`, or `"check_unique"`.
 #'   See [vctrs::vec_as_names()] for the meaning of these options.
+#' @param keep_empty  An optional logical. If `FALSE` (the default), then
```
DavisVaughan, line 25: I think `keep_empty` is the wrong name if `double()` isn't promoted to `NA_real_` and kept. It's more like `keep_null`.

```r
> list_c(list(1, double(), 2))
[1] 1 2
> list_c(list(1, double(), 2), keep_empty = T)
[1] 1 2
```

DavisVaughan: Compare with

```r
> tidyr::unchop(tibble::tibble(x = list(1, double(), 2)), x)
# A tibble: 2 × 1
      x
  <dbl>
1     1
2     2
> tidyr::unchop(tibble::tibble(x = list(1, double(), 2)), x, keep_empty = TRUE)
# A tibble: 3 × 1
      x
  <dbl>
1     1
2    NA
3     2
```

R/list-combine.R
```diff
@@ -30,16 +33,21 @@
...
+#' list_rbind(x2, names_to = "id", keep_empty = TRUE)
 #' list_rbind(unname(x2), names_to = "id")
-#'
 #' list_cbind(x2)
-list_c <- function(x, ..., ptype = NULL) {
+#' list_cbind(x2, keep_empty = TRUE)
+list_c <- function(x, ..., ptype = NULL, keep_empty = FALSE) {
   vec_check_list(x)
   check_dots_empty()
+  if (keep_empty) {
```
DavisVaughan, line 48: We'd want to use `check_bool(keep_empty)` everywhere

R/list-combine.R
```diff
@@ -95,3 +110,9 @@ check_list_of_data_frames <- function(x, error_call = caller_env()) {
     call = error_call
   )
 }
+
+convert_null_to_NA <- function(x) {
+  is_null <- map_lgl(x, is.null)
+  x[is_null] <- list(NA)
```
DavisVaughan, line 116: `vctrs::vec_detect_missing()` would be much faster at detecting `NULL`s since we know `x` is a list.

See `tidyr:::list_replace_null()` for a robust and very fast version of this operation

tests/testthat/test-list-combine.R
```diff
@@ -70,6 +70,24 @@ test_that("NULLs are ignored", {
...
+  )
+  expect_equal(
+    list_rbind(list(df1, NULL, df1), keep_empty = TRUE),
+    data.frame(x = c(1, NA, 1))
+  )
+  expect_equal(
+    list_cbind(list(df1, z = NULL, df2), keep_empty = TRUE),
+    data.frame(df1, z = NA, df2)
+  )
+})
```
DavisVaughan, line 89: If we decide to implement this then I'd like the chance to come through and write _a lot_ of tests for the 3 cases individually. The `keep_empty` logic in tidyr is quite hard to get 100% right, and required a lot of edge case tests.

DavisVaughan (commented):
Dropping in some initial comments while looking at this. No need to make any changes yet though.

But also see https://github.com/tidyverse/purrr/issues/1096#issuecomment-2315466918. I think it is possible we made a small mistake by marking this as a TDD issue, since it can already be solved through `tidyr::unchop()` in a way that I feel is much more robust for the motivating example given there (i.e. the size invariants are nicer when you use `unchop()`).

DavisVaughan:
We have decided not to implement this feature for now since:
- `tidyr::unchop()` solves the problem already in a very robust way https://github.com/tidyverse/purrr/issues/1096#issuecomment-2315466918
- We can't seem to justify the argument for `list_cbind()`, suggesting something feels off with putting it here in purrr

@SokolovAnatoliy we apologize about that! We hope you had fun at TDD anyways, and it seems like you got to do some other PRs in other repos, which is awesome! Thanks again!

PR #1146, opened 2024-08-15
Updated reduce documentation

### LINKED ISSUE #1103: Documentation of `reduce` refers to `map`
The documentation of the `...` attribute refers to the `map()` function. It would be a bit nicer if the reference for `reduce` would point to `reduce()` instead of `map()` ;)

ccani007:
Fixed  #1103.  Eliminated accumulate2() and document  ...  itself.


hadley:
Thanks for working on this! I hope you enjoyed TDD 😄

PR #1141, opened 2024-08-15
allow list_transpose() to work on data frames

KimLopezGuell:
Fixed issue #1109

PR #1149, opened 2024-08-21
Revert allowing data frames in `list_transpose()`

DavisVaughan:
Reverts #1141

@KimLopezGuell we have looked at this a bit more in detail and decided that we actually should not have labeled this an issue for TDD, that is totally our fault! We do actually want to be strict about `list_transpose()` only allowing lists, not data frames, just like almost every other `list_*()` function in both purrr and vctrs. But your PR was great, and we hope you had a lot of fun at TDD! I've reused your test as an additional check for this going forward, so not all was lost!

### DIFF
--- a/tests/testthat/test-list-transpose.R
+++ b/tests/testthat/test-list-transpose.R
@@ -8,9 +8,7 @@ test_that("can't transpose data frames", {
   df <- data.frame(x = 1:2, y = 4:5)
 
   # i.e. be consistent with other `list_*()` functions from purrr/vctrs
-  expect_snapshot(error = TRUE, {
-    list_transpose(df)
-  })
+  expect_snapshot(error = TRUE, list_transpose(df))
 })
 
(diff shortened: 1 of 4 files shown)


tests/testthat/test-list-transpose.R
```diff
@@ -4,10 +4,13 @@ test_that("can transpose homogenous list", {
...
-test_that("can transpose data frames", {
+test_that("can't transpose data frames", {
   df <- data.frame(x = 1:2, y = 4:5)
-  out <- list_transpose(df)
-  expect_equal(out, list(c(x = 1, y = 4), c(x = 2, y = 5)))
+
+  # i.e. be consistent with other `list_*()` functions from purrr/vctrs
+  expect_snapshot(error = TRUE, {
+    list_transpose(df)
+  })
```
hadley, line 13: You could put this in one line.

hadley (approved):

PR #1159, opened 2025-01-11
Fix CRAN ASAN error

### LINKED ISSUE #1157: CRAN clang-san error from purrr during package submission pre-check
We are updating a package on CRAN and this error is thrown during the pre-check, from `cleancall.c` in purrr:

```
File: valr-Ex.Rout
cleancall.c:110:46: runtime error: call to function cb_progress_done through pointer to incorrect function type 'void (*)(void *)'

File: tests/testthat.Rout
cleancall.c:110:46: runtime error: call to function cb_progress_done through pointer to incorrect function type 'void (*)(void *)'
```

https://github.com/rnabioco/valr/issues/423#issuecomment-2577973592

CRAN wants us to fix. We don't import purrr directly but it's an eventual dependency (haven't tracked the specific culprit though).

``` r
db <- tools::CRAN_package_db()    
deps <- tools::package_dependencies("valr", recursive=TRUE, db=db)

"purrr" %in% deps$valr
#> [1] TRUE
```

<sup>Created on 2025-01-09 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

This error doesn't appear to have come up before in the purrr issues. Any ideas for a fix?

jayhesselberth:
Fixes #1157

I confirmed that clang-asan fails on current purrr with a similar error to what we're seeing in valr. This is in purrr-Ex.Rout from [this rhub check](https://github.com/jayhesselberth/purrr/actions/runs/12720310041/job/35461775454).

```
cleancall.c:110:46: runtime error: call to function cb_progress_done through pointer to incorrect function type 'void (*)(void *)'
/__w/purrr/purrr/check/purrr.Rcheck/00_pkg_src/purrr/src/map.c:15: note: cb_progress_done defined here
    #0 0x7fd287f0d48b in call_exits /__w/purrr/purrr/check/purrr.Rcheck/00_pkg_src/purrr/src/cleancall.c:110:46
    #1 0x7fd28ddf1131 in R_ExecWithCleanup /tmp/R-devel/src/main/context.c:913:5
    #2 0x7fd287f0cdf2 in r_with_cleanup_context /__w/purrr/purrr/check/purrr.Rcheck/00_pkg_src/purrr/src/cleancall.c:133:14
```

This simple patch proposed on the R-devel mailing list fixes that clang-asan error ([clang asan run](https://github.com/jayhesselberth/purrr/actions/runs/12718786947/job/35457848128)).

Also updated the rhub action because it was useful to debug this issue.

### DIFF
--- a/.github/workflows/rhub.yaml
+++ b/.github/workflows/rhub.yaml
@@ -1,13 +1,13 @@
-# R-hub's genetic GitHub Actions workflow file. It's canonical location is at
-# https://github.com/r-hub/rhub2/blob/v1/inst/workflow/rhub.yaml
+# R-hub's generic GitHub Actions workflow file. It's canonical location is at
+# https://github.com/r-hub/actions/blob/v1/workflows/rhub.yaml
 # You can update this file to a newer version using the rhub2 package:
 #
-# rhub2::rhub_setup()
+# rhub::rhub_setup()
 #
 # It is unlikely that you need to modify this file manually.
 
 name: R-hub
-run-name: ${{ github.event.inputs.name || format('Manually run by {0}', github.triggering_actor) }} (${{ github.event.inputs.id }})
+run-name: "${{ github.event.inputs.id }}: ${{ github.event.inputs.name || format('Manually run by {0}', github.triggering_actor) }}"
 
 on:
   workflow_dispatch:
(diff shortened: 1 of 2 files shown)


hadley:
Thanks!

PR #1163, opened 2025-01-22
Parallelized map using mirai

### LINKED ISSUE #1162: Parallelized Map
As proposed by @hadley at Posit::Conf 2024 in Seattle, it is possible to use the mirai package as an efficient, lightweight backend to provide parallel / distributed map for purrr.

shikokuchuo:
@hadley this builds on the proof of concept you reviewed last week at https://github.com/shikokuchuo/purrr/pull/1.

Adds a `.parallel` argument to every function that currently accepts a `.progress` argument, i.e.:
- `map()` and variants
- `map2()` and variants
- `pmap()` and variants

There is a full test suite added at `test-parallel.R`, which runs parallel versions of __all__ the tests from the map/map2/pmap test files, so I'm fairly confident that the behaviour is consistent.

This version relies on current dev `mirai`. I made certain upstream changes to align, in particular, the error messages so that they are consistent and identical to the normal map cases. I'm ready to make another mirai release once this PR is merged.

I've added a couple of simple examples at the bottom of `map()` so it's hopefully clear to users what they should be doing.

The full documentation is available as an Rd page at `?parallelization`. I've added it this way, as opposed to a vignette, to mirror how `.progress` is documented. It also has the advantage of linking easily from function documentation. This contains:
- An explanation of when to provide a list to `.parallel`
- How to set `daemons()`
- The `with()` method for daemons
---
Conceptually, this has been mostly an interface-mapping exercise, with purrr directly using the public exported functions `mirai_map()` and `collect_mirai()` from the mirai package. I've focused almost completely on consistency with the normal, non-parallel, versions thus far.

I've resisted the urge to do any specific optimization for efficiency, as it's likely premature to do so. Just noting here that this is likely to yield results in future.

Welcome your review!

cc. @jcheng5 FYI.

closes #1162.


R/map.R
```diff
@@ -125,50 +130,70 @@
...
 #'   map(summary) |>
 #'   map_dbl("r.squared")
-map <- function(.x, .f, ..., .progress = FALSE) {
-  map_("list", .x, .f, ..., .progress = .progress)
+#'
+#' @examplesIf interactive()
+#' # run in interactive sessions only as spawns additional processes
+#'
+#' # To use parallelized map, set daemons (number of parallel processes) first:
+#' daemons(2)
```
hadley, line 138: What happens if you don't set daemons?

....

Oh I see you describe the later.

Here it might be worth a comment that in user code you'd normally leave the daemons running throughout your session? (I assume that's the case?)

R/map.R
```diff
@@ -179,21 +204,60 @@ map_ <- function(.type,
   )
 }
 
+mmap_ <- function(.x, .f, .args, .parallel, .progress, .type, error_call) {
```
hadley, line 207: Maybe this should live in the parallelisation file?

R/parallelization.R
```diff
@@ -0,0 +1,101 @@
...
+#'
+#' ## with() method
+#'
+#' The `with()` method for daemons provides a convenient way of evaluating a
+#' statement with daemons set up for the duration of the statement.
+#'
+#' As an example, the below is evaluated in parallel using 7 daemons:
+#'
+#' \preformatted{
+#' with(daemons(7), {
```
hadley, line 87: The slightly more standard tidyverse way would be to provide `with_daemons()` and `local_daemons()`. That might be worth doing here. I'm not sure.

R/parallelization.R
```diff
@@ -0,0 +1,101 @@
+#' Parallelization in purrr
+#'
+#' @description
+#' purrr's map functions have a `.parallel` argument to parallelize a map using
+#' the \CRANpkg{mirai} package. This allows you to run computations in parallel
+#' using more cores on your machine, or distributed over the network.
```
hadley, line 6: I think it would be worth including a little advice here about when parallelisation is likely to save you time — i.e. talk a little about roughly what the overhead associated with parallelisation is.

hadley (commented):
This is looking great. Thanks so much for working on this!

R/parallelization.R
```diff
@@ -0,0 +1,101 @@
...
+#' If `.f` is a user-defined function, then ideally everything required by the
+#' function is passed in as arguments (a pure function). Any constant arguments
+#' in this case can be supplied through `...` in the usual way.
+#'
+#' However, if the function references an object that is not supplied as an
+#' argument or defined in the function itself, i.e. a free variable, these
+#' should be supplied as a list to `.parallel`. This makes them available to
+#' `.f` in the parallel processes. `purrr` requires these to be explicitly
+#' supplied, rather than try to determine them from analysing `.f`, as this
+#' provides for consistent and reliable behaviour that is hence easy to debug.
```
lionel-, line 29: This API feel a bit unfriendly.

Any thoughts on using `carrier::crate()` to create serialisable functions instead?
https://github.com/r-lib/carrier

R/parallelization.R
lionel-: I thought having it work with `with()` is a nice touch :)

lionel- (commented):
Great to see progress on this front!

I'm not familiar with mirai so I mainly have a high level comment about potentially using carrier for the treatment of functions.

R/map.R
shikokuchuo: Yes, setting daemons once per session is usually the idea. Added in [84d6b63](https://github.com/tidyverse/purrr/pull/1163/commits/84d6b6356a8ba7712bf6e54a23aeb3bd3bbc9506).

R/map.R
shikokuchuo: Yes, I guess it's used by several of the files. Moved in [d348652](https://github.com/tidyverse/purrr/pull/1163/commits/d348652b4e54a6897a50e486d98f46f945d514df).

R/parallelization.R
shikokuchuo: I was about to suggest just removing mention of `with()` in the purrr docs for now, but Lionel seems to like it!

It's not at all essential and it's probably more idiomatic just to set daemons at the top of your script for interactive work.

It was designed in the context of Shiny as people tend to not know where to put the `daemons()` call - at the top level or within `server()` etc., so I introduced it so it's easy to do:
```r
app <- shinyApp(server, ui, session)
with(daemons(4), runApp(app))
```

R/parallelization.R
```diff
@@ -0,0 +1,144 @@
...
+      "'.parallel' must be TRUE/FALSE or a list, not a {.obj_type_friendly {(.parallel)}}.",
+      call = error_call
+    )
+  }
+
+  options <- c(".stop", if (isTRUE(.progress)) ".progress")
+  x <- withCallingHandlers(
+    mirai::collect_mirai(m, options = options),
+    error = function(cnd) {
+      location <- cnd$location
```
DavisVaughan, line 123: Maybe extract out `with_parallel_indexed_errors()` to be put near `with_indexed_errors()` so it is easier for us to keep them in sync in the future?

R/parallelization.R
```diff
@@ -0,0 +1,144 @@
...
+  } else if (is.list(.parallel)) {
+    mirai::mirai_map(.x, .f, as.environment(.parallel), .args = .args)
+  } else {
+    cli::cli_abort(
+      "'.parallel' must be TRUE/FALSE or a list, not a {.obj_type_friendly {(.parallel)}}.",
+      call = error_call
+    )
+  }
+
+  options <- c(".stop", if (isTRUE(.progress)) ".progress")
```
DavisVaughan, line 119: I'm not entirely sure how you'd expose this as an option, but I would not be surprised if we eventually got requests for a way to opt out of "early stopping", instead running everything to completion/failure.

I think what you have here is probably fine for now, just thinking out loud about this one

shikokuchuo: That would require purr to actually return something other than error. 

It's of course better to stop early as soon as we know we're going to error if that's all we're going to do anyway.

I think this would be another `.` argument as it would apply to purrr in general (it wouldn't be useful just for the parallel case?)  When that happens, it can then be handled the same way as `.progress`.

R/parallelization.R
shikokuchuo: I had a go, but it seems cleaner to keep as is - this is as the interrupt handler requires a reference to the mirai object to be passed in, which would then be inconsistent with having a general `expr` in the fashion of `with_indexed_errors()`.

R/parallelization.R
```diff
@@ -0,0 +1,104 @@
...
+#' it is launching one local daemon - this ensures that the function proceeds,
+#' but offers no actual parallelization and is hence probably not what you would
+#' want.
+#'
+#' `daemons()`arguments:
+#'
+#' * `n`: the number of daemons to launch on your local machine, e.g.
+#'   `daemons(7)`. As a rule of thumb, for maximum efficiency this should be (at
+#'   most) one less than the number of cores on your machine, leaving one core
+#'   for the main R process.
```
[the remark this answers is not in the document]
shikokuchuo: This is indeed the reason why the number of daemons is user-supplied. It's not clear that any automated function is going to get this right, precisely as we can't know what else is going on in the compute environment at the same time.

R/parallelization.R
```diff
@@ -0,0 +1,104 @@
...
+#'
+#' Examples:
+#' \preformatted{
+#' # can just specify TRUE:
+#' map(1:3, stats::runif, .parallel = TRUE)
+#'
+#' # requires 'fun1' to be supplied in a list:
+#' fun1 <- function(x) \{x + x \%\% 2 \}
+#' fun2 <- function(x) \{ x + fun1(x) \}
+#' map(1:3, fun2, .parallel = list(fun1 = fun1))
```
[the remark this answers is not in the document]
shikokuchuo: A function will get serialized as a closure, but the global environment is never serialized by R. This is just R serialization btw. not something bespoke.

Key here is that `fun2` is defined in the global environment, but has the free variable `fun1` (it's a function in this case, but it could be any object). As `fun2`'s closure _is_ the global environment, then we need to ship `fun1` and put it in the global environment of the remote process.

So if `fun2` is a closure that contains `fun1` then we wouldn't need to worry about separately shipping `fun1`.

At the end of the day, I do think most of the time it's going to be a package function here and then we're safe.

`carrier::crate()` also seems to provide a safe way to construct these closures, so we could recommend that.

R/parallelization.R
```diff
@@ -0,0 +1,104 @@
...
+#' parallel processes.
+#'
+#' ## Supplying objects to parallel processes
+#'
+#' In the majority of cases, such as when `.f` is a base or package function,
+#' simply set `.parallel` to `TRUE`.
+#'
+#' If `.f` is a user-defined function, then ideally everything required by the
+#' function is passed in as arguments (a pure function). Any constant arguments
+#' in this case can be supplied through `...` in the usual way.
```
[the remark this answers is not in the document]
shikokuchuo: The NSE for the `...` is funky! I see the logic encouraging the use of anonymous functions. I think documenting the difference is probably safest for now. If we do anything stronger for the normal case, I think that's when we do the same here.

R/parallelization.R
shikokuchuo: This looks great! I'm certainly happy to recommend it in the docs. Were you thinking of forcing people to use this and not have the list API at all?

R/parallelization.R
```diff
@@ -0,0 +1,104 @@
+#' Parallelization in purrr
```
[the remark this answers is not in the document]
shikokuchuo: Let me take a look at the `furrr` docs to refresh my understanding of why it's an issue there. From what I remember it's the way that chunking is done + the lack of queuing in the old `parallel` backend. mirai should be completely flexible in just working in this respect. But I'll add something to the docs on this point.

R/parallelization.R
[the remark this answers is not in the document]
DavisVaughan: This is what I am worried about, which I get questions about in furrr a lot

"Why is this furrr thing so slow??" - because it isn't obvious that serializing `helper` also captured its enclosing environment, so `data` came along for the ride and is shipped to each worker, and often times `data` itself is _huuuuuge_ and you never intended for it to get there.

``` r
library(purrr)
library(mirai)
library(tibble)

daemons(2)
#> [1] 2

do_the_thing <- function(data) {
  # Imagine this is a function that doesn't use `data` directly at all,
  # so it should not need access to it. I expect that `env$data`
  # should return `NULL`.
  helper <- function(x) {
    env <- parent.env(environment())
    env$data
  }
  
  # `helper` should be serialized in a special way where its
  # environment is forcibly set to the global environment,
  # preventing anything else from "accidentally" getting
  # serialized along with it
  map(data$x, helper, .parallel = TRUE)
}

data <- tibble(
  x = list(1, 2, 3),
  y = 1:3
)

do_the_thing(data)
#> [[1]]
#> # A tibble: 3 × 2
#>   x             y
#>   <list>    <int>
#> 1 <dbl [1]>     1
#> 2 <dbl [1]>     2
#> 3 <dbl [1]>     3
#> 
#> [[2]]
#> # A tibble: 3 × 2
#>   x             y
#>   <list>    <int>
#> 1 <dbl [1]>     1
#> 2 <dbl [1]>     2
#> 3 <dbl [1]>     3
#> 
#> [[3]]
#> # A tibble: 3 × 2
#>   x             y
#>   <list>    <int>
#> 1 <dbl [1]>     1
#> 2 <dbl [1]>     2
#> 3 <dbl [1]>     3
```

<sup>Created on 2025-01-23 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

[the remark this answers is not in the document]
DavisVaughan: And before you say "well defining a named helper function inside another function isn't that common", note that the same behavior also applies to anonymous functions, and something like this is _very_ common

```r
do_the_thing <- function(data) {
  map(data$x, .parallel = TRUE, function(x) {
    env <- parent.env(environment())
    env$data
  })
}
```

R/parallelization.R
shikokuchuo: In [f349d71](https://github.com/tidyverse/purrr/pull/1163/commits/f349d71ca90609ccb17ef378d7606b3a63efab9a) I remove the option of supplying a list to `.parallel` to make the interface less unwieldy.

We now rely on `carrier::crate()` to create self-contained functions where necessary and this is documented in `?parallelization`.

This provides for a more robust and maintainable solution to have the `carrier` package bear this responsibility.

R/parallelization.R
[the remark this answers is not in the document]
shikokuchuo: Actually as we're trying to discourage the use of `...`, Hadley's idea is to go straight for the error, thus moving faster to the 'ideal state'.

This is implemented in [db60ece](https://github.com/tidyverse/purrr/pull/1163/commits/db60ece748d1d0948574ec1aabbd6363bdc81a2d). I've also added a section in the docs based on Hadley's drafting from elsewhere.

R/parallelization.R
shikokuchuo: It's possible to implement `with_daemons()` and `local_daemons()`, but I prefer not to do it in the initial implementation so as not to overcomplicate things.

I think there's no harm leaving the `with()` method in the docs, but if you think it diverts attention etc. I have no issue with taking the section out.

R/parallelization.R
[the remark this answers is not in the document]
shikokuchuo: Thanks I can see that `data` is being passed to the daemons along with `.f`.

As far as I know, stripping of function environments isn't done in any existing implementation? I'd have to consider the consequences. Also there would need to be exceptions for package namespaces and 'crate' objects at least, now that we rely on `carrier::crate()` to 'crate' our functions :)

R/parallelization.R
shikokuchuo: Further, in [60dadd1](https://github.com/tidyverse/purrr/pull/1163/commits/60dadd150f27eec120e7a828902582d4f400ce14) I import carrier and re-export `crate()` so it can be easily used within purrr.

There are now clear recommendations to `crate()` functions for use in a parallel map, both upfront in the '.parallel' argument definition, and also in `?parallelization`.

The parallelization docs have been streamlined to focus on just 2 things: (i) daemons and (ii) crating.

I think this is the best we can do to make things user-friendly at this stage, as `carrier::crate()` only works with 'fresh' functions / formulae.

It may be the scope of an update to `carrier` to have it automatically crate existing functions. If that happens then we could possibly look to inline it in purrr and drop the recommendation in the docs. However already-crated functions would presumably still work, so any upgrade would be backwards compatible and not disruptive.

R/parallelization.R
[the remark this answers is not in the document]
shikokuchuo: So in [60dadd1](https://github.com/tidyverse/purrr/pull/1163/commits/60dadd150f27eec120e7a828902582d4f400ce14) I update both the `.parallel` argument docs and `?parallelization` to recommend the use of `crate()`, which is now re-exported by purrr for ease of use.

This does effectively strip the environment of a function and prevents the situation above.

``` r
library(purrr)
library(mirai)
library(tibble)

daemons(2)
#> [1] 2

data <- tibble(
  x = list(1, 2, 3),
  y = 1:3
)

do_the_thing <- function(data) {
  map(data$x, .parallel = TRUE, function(x) {
    env <- parent.env(environment())
    env$data
  })
}

do_the_thing(data)
#> [[1]]
#> # A tibble: 3 × 2
#>   x             y
#>   <list>    <int>
#> 1 <dbl [1]>     1
#> 2 <dbl [1]>     2
#> 3 <dbl [1]>     3
#> 
#> [[2]]
#> # A tibble: 3 × 2
#>   x             y
#>   <list>    <int>
#> 1 <dbl [1]>     1
#> 2 <dbl [1]>     2
#> 3 <dbl [1]>     3
#> 
#> [[3]]
#> # A tibble: 3 × 2
#>   x             y
#>   <list>    <int>
#> 1 <dbl [1]>     1
#> 2 <dbl [1]>     2
#> 3 <dbl [1]>     3

do_crated <- function(data) {
  map(data$x, .parallel = TRUE, crate(function(x) {
    env <- parent.env(environment())
    env$data
  }))
}

do_crated(data)
#> [[1]]
#> NULL
#> 
#> [[2]]
#> NULL
#> 
#> [[3]]
#> NULL
```

<sup>Created on 2025-01-24 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

R/parallelization.R
shikokuchuo: As part of [60dadd1](https://github.com/tidyverse/purrr/pull/1163/commits/60dadd150f27eec120e7a828902582d4f400ce14) I've taken out mention of the `with()` method in the purrr docs as I want it to just focus on 2 things: (i) daemons and (ii) crating.

It's still available of course, and users may find it in the `daemons()` docs.

R/parallelization.R
shikokuchuo: Actually I've taken your suggestion in [fda5abb](https://github.com/tidyverse/purrr/pull/1163/commits/fda5abb85901d18be98c8d715b42dd5afe3e70fa). I pass in an `interrupt_expr` as well as the main `expr`.
It does help to separate out the main logic. Thanks.

R/map.R
shikokuchuo: Actually due to https://github.com/tidyverse/purrr/pull/1163#discussion_r1927130312 I've moved it back so that the parallel function versions are just directly below the normal ones. It's easier so that everything is in one place.

R/map.R
shikokuchuo: Also I've updated the behaviour so that there is now an error with a suggestion if daemons are not set. I think this is what you had in mind.

R/parallelization.R
shikokuchuo: I've added a point in [b5824c1](https://github.com/tidyverse/purrr/pull/1163/commits/b5824c1af366edb660220d415ce60e3e7ca10854), trying to be as succinct as possible.

The backup to the 100microsecond - 1millisecond guide is below:

The first set measures the pure comms overhead, which shows parallel winning at around 50 microseconds.

The second set measures comms + serialization overhead, shipping the large diamonds data each time (3.5MB), and there parallel starts winning at around 500 microseconds.

``` r
library(purrr)
daemons(8)
#> [1] 8
bench::mark(
  map(1:20, function(x) Sys.sleep(0.0001)),
  map(1:20, function(x) Sys.sleep(0.0001), .parallel = TRUE)
)
#> # A tibble: 2 × 6
#>   expression                             min median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>                          <bch:> <bch:>     <dbl> <bch:byt>    <dbl>
#> 1 map(1:20, function(x) Sys.sleep(1e… 3.04ms 3.18ms      315.    4.04MB     0   
#> 2 map(1:20, function(x) Sys.sleep(1e… 1.54ms 1.75ms      504.  515.27KB     2.02
bench::mark(
  map(1:20, function(x) Sys.sleep(0.00005)),
  map(1:20, function(x) Sys.sleep(0.00005), .parallel = TRUE)
)
#> # A tibble: 2 × 6
#>   expression                             min median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>                          <bch:> <bch:>     <dbl> <bch:byt>    <dbl>
#> 1 map(1:20, function(x) Sys.sleep(5e…    2ms 2.09ms      475.    4.88KB     0   
#> 2 map(1:20, function(x) Sys.sleep(5e… 1.53ms 1.68ms      535.   194.2KB     4.08
bench::mark(
  map(1:20, function(x) Sys.sleep(0.00001)),
  map(1:20, function(x) Sys.sleep(0.00001), .parallel = TRUE)
)
#> # A tibble: 2 × 6
#>   expression                             min median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>                          <bch:> <bch:>     <dbl> <bch:byt>    <dbl>
#> 1 map(1:20, function(x) Sys.sleep(1e… 1.23ms 1.25ms      767.    4.88KB     2.02
#> 2 map(1:20, function(x) Sys.sleep(1e… 1.45ms 1.62ms      562.   194.2KB     4.13

library(ggplot2)
bench::mark(
  map(diamonds, function(x) Sys.sleep(0.001)),
  map(diamonds, function(x) Sys.sleep(0.001), .parallel = TRUE)
)
#> # A tibble: 2 × 6
#>   expression                            min  median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>                        <bch:t> <bch:t>     <dbl> <bch:byt>    <dbl>
#> 1 map(diamonds, function(x) Sys.sl… 10.98ms 11.36ms      88.0    10.4MB     0   
#> 2 map(diamonds, function(x) Sys.sl…  3.63ms  3.96ms     228.     96.8KB     2.02
bench::mark(
  map(diamonds, function(x) Sys.sleep(0.0005)),
  map(diamonds, function(x) Sys.sleep(0.0005), .parallel = TRUE)
)
#> # A tibble: 2 × 6
#>   expression                             min median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>                          <bch:> <bch:>     <dbl> <bch:byt>    <dbl>
#> 1 map(diamonds, function(x) Sys.slee… 5.73ms  5.8ms      172.    4.67KB     0   
#> 2 map(diamonds, function(x) Sys.slee… 2.66ms 3.08ms      299.    96.8KB     2.02
bench::mark(
  map(diamonds, function(x) Sys.sleep(0.0001)),
  map(diamonds, function(x) Sys.sleep(0.0001), .parallel = TRUE)
)
#> # A tibble: 2 × 6
#>   expression                             min median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>                          <bch:> <bch:>     <dbl> <bch:byt>    <dbl>
#> 1 map(diamonds, function(x) Sys.slee…  1.5ms 1.59ms      626.        0B     0   
#> 2 map(diamonds, function(x) Sys.slee… 2.31ms 2.64ms      347.    96.8KB     2.03
```

<sup>Created on 2025-01-24 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

R/parallelization.R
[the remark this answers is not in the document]
shikokuchuo: I didn't see anything specific at `furrr`, but I did read the nested topologies at https://future.futureverse.org/index.html?q=nested#nested-futures-and-evaluation-topologies. In conclusion, I think this is specific to the design of future and it's syntax that requires you to define 2 layers up front.

mirai can spawn other mirai and so on, and isn't limited by the initial `daemons()` setting. So you can easily have daemons launched over the network, and then each of those launch local daemons for example.

For mirai/purrr it works just like the below. The only thing you really need to remember is to set daemons within the function. I haven't added anything specifically to the docs as I really want to keep that focused on (i) daemons and (ii) crating.

I'd be happy to add to a vignette down the line along with whatever other issues people start reporting.

``` r
purrr::daemons(3)
#> [1] 3
f <- function(x) {purrr::daemons(2); purrr::map(1:2, function(x) Sys.getpid(), .parallel = TRUE)}
purrr::map(1:3, f, .parallel = TRUE)
#> [[1]]
#> [[1]][[1]]
#> [1] 68371
#> 
#> [[1]][[2]]
#> [1] 68366
#> 
#> 
#> [[2]]
#> [[2]][[1]]
#> [1] 68295
#> 
#> [[2]][[2]]
#> [1] 68297
#> 
#> 
#> [[3]]
#> [[3]][[1]]
#> [1] 68331
#> 
#> [[3]][[2]]
#> [1] 68334
```

<sup>Created on 2025-01-24 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

lionel- (commented):
It would be nice to avoid implicit deletion of the function environment as I think it is surprising behaviour.

To that end when a non-package function is supplied, we could:

- Fail and recommend calling `crate()`
- Auto-crate and print a message about it (so it's not implicit) with the crate size (suggested by @hadley on Slack).

Regarding auto-crating, it possibly should be restricted to the case of calling from the global env as determined by `identical(topenv(caller_env()), globalenv())`. In that case it would be regarded as a script user convenience whereas package authors would have to explicitly crate.

Should we set the default argument of `.parallel` to `TRUE` when the input is a crate?

DavisVaughan:
> Should we set the default argument of .parallel to TRUE when the input is a crate

I think that would be a bit too magical. I would want to see an explicit `parallel = TRUE` call in my code when I come back to it in 6 months

shikokuchuo:
> To that end when a non-package function is supplied, we could:
>     * Auto-crate and print a message about it (so it's not implicit) with the crate size (suggested by @hadley on Slack).

Auto-crating for non-package functions is now implemented in [93eaa2c](https://github.com/tidyverse/purrr/pull/1163/commits/93eaa2c9270f0606040d21696788a659d70fa268). @hadley I believe this now rounds out the interface you had in mind:
- error on no daemons set
- auto-crating
- no use of `...`

> Regarding auto-crating, it possibly should be restricted to the case of calling from the global env as determined by `identical(topenv(caller_env()), globalenv())`. In that case it would be regarded as a script user convenience whereas package authors would have to explicitly crate.

I think it could be confusing having different behaviours here depending on whether it's called from a package.

More importantly, if the goal is to encourage package authors to use `crate()`, then I think this is achieved by the current behaviour, as auto-crating doesn't happen when a crate is passed to `.f`. So they would avoid this by explicitly crating themselves - I think that makes sense.

DESCRIPTION
```diff
@@ -15,9 +15,11 @@ BugReports: https://github.com/tidyverse/purrr/issues
 Depends: 
     R (>= 4.0)
 Imports:
+    carrier (>= 0.1.1),
```
hadley, line 18: For now, lets keep carrier and mirai as suggests. That will make the syntax a little less appealing, but purrr is such a widely used package I want to be extra careful before adding any new packages.

R/map.R
```diff
@@ -46,6 +46,14 @@
...
 #'   function and will tend to yield better error messages.
 #'
+#' @param .parallel `r lifecycle::badge("experimental")` Whether to map in
+#'   parallel. Use `TRUE` to parallelize using the \CRANpkg{mirai} package. Set
+#'   up parallelization in your session beforehand using
+#'   [`daemons()`][mirai::daemons]. We recommended you [crate][carrier::crate]
+#'   your function for sharing with parallel processes. Non-package functions
+#'   are auto-crated. Use of `...` is not permitted in this context,
+#'   [crate][carrier::crate] an anonymous function instead. See
+#'   [parallelization] for more details.
```
hadley, line 56:
```suggestion
#'   parallel. Use `TRUE` to parallelize using the \CRANpkg{mirai} package. 
#'   * Set up parallelization in your session beforehand using [`daemons()`][mirai::daemons].
#'   * We recommended you [crate][carrier::crate] your function for sharing with 
#'      parallel processes. Non-package functions are auto-crated. 
#    * Use of `...` is not permitted in this context, [crate][carrier::crate] an anonymous 
#'     function instead. 
#'  See [parallelization] for more details.
```

Still needs a bit more work, but I think a bulleted list is easier to skim.

R/map.R
```diff
@@ -225,6 +289,28 @@ with_indexed_errors <- function(expr, i, names = NULL, error_call = caller_env()
...
 
+with_parallel_indexed_errors <- function(expr, interrupt_expr = NULL, error_call = caller_env()) {
+  withCallingHandlers(
+    expr,
+    error = function(cnd) {
+      location <- cnd$location
+      iname <- cnd$name
+      cli::cli_abort(
+        c(i = "In index: {location}.",
+          i = if (length(iname) && nzchar(iname)) "With name: {iname}."),
```
hadley, line 300:
```suggestion
        c(
          i = "In index: {location}.",
          i = if (length(iname) && nzchar(iname)) "With name: {iname}."
        ),
```

R/map.R
hadley: If we auto-create functions, why do we recommend that you do it yourself?

R/map.R
```diff
@@ -179,21 +207,57 @@ map_ <- function(.type,
...
+      "Don't use `...` with `.parallel = TRUE`.",
+      call = error_call
+    )
+  }
+
+  if (!carrier::is_crate(.f) && !isNamespace(topenv(environment(.f)))) {
+    .f <- crate(rlang::set_env(.f))
+    cli::cli_alert_success(
+      "Automatically crated `.f`: {format(lobstr::obj_size(.f))}"
+    )
```
hadley, line 229:
```suggestion
    cli::cli_inform(c(
      v = "Automatically crated `.f`: {format(lobstr::obj_size(.f))}"
    ))
```
```suggestion
    cli::cli_alert_success(
      "Automatically crated `.f`: {format(lobstr::obj_size(.f))}"
    )
```

R/parallelization.R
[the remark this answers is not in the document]
hadley: I think the primary concern is that if you accidentally use `parallel = TRUE` with a function that already uses `parallel = TRUE`, you can end up with a multiplicative explosion of processes and this sucks up a bunch of system memory etc.

I just did `mirai::daemons(1e4)` and hard crashed my computer 😬

R/parallelization.R
hadley: Yes, that's perfect, thanks!

R/parallelization.R
```diff
@@ -0,0 +1,107 @@
...
+#'
+#' All daemons automatically terminate when your session ends and the connection
+#' drops. Hence you do not need to explicitly terminate daemons in this instance,
+#' although it is still good practice to do so.
+#'
+#' # Crating a function
+#'
+#' As the function `.f` needs to be serialized and shared with other processes,
+#' we recommend that you [crate()][carrier::crate] your function for use in a
+#' parallel map.
```
hadley, line 61: Mention something about it being self-contained?

R/parallelization.R
```diff
@@ -0,0 +1,107 @@
...
+#'
+#' The only exception is if `.f` is a function from a package or base R. Do not
+#' [crate][carrier::crate] these functions - although you can
```
hadley, line 64:
```suggestion
#' The only exception is if `.f` is a function from a package or base R, since these are already self-contained. Do not
#' [crate][carrier::crate] these functions - although you can
```
```suggestion
#' The only exception is if `.f` is a function from a package or base R. Do not
#' [crate][carrier::crate] these functions - although you can
```

hadley (approved):
This looks like a great first pass at this feature, and I think we're on track to merge it after this last round of (mostly minor) feedback.

R/parallelization.R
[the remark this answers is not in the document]
shikokuchuo: Right, so I think that means we're protected as just using `parallel = TRUE` never creates any daemons, so you can't 'accidentally' do this. If we supply a function that itself contains a `map(parallel = TRUE)`, that would simply error in the daemon.

Note my example above only works as we explicitly call `daemons(n)` as part of the function. This is also a reason why we require the user to explicitly supply this number rather than leave it to a function that e.g. automatically detects the number of cores. There is also long-standing [guidance](https://shikokuchuo.net/mirai/articles/mirai.html#using-mirai-in-a-package) for package authors to leave daemons settings to the user.

Btw. `mirai::daemons(1e4)` is just a simple OOM. It _will_ handle that many connections i.e. over the network, but trying to launch that many R processes on one machine is just too much...

R/map.R
shikokuchuo: True, shouldn't be a recommendation. Have re-worded in [300c16e](https://github.com/tidyverse/purrr/pull/1163/commits/300c16ea947a4c1122c730fe8e20fbc79b637555) - explicitly crate when you need to crate your function with other objects.

R/parallelization.R
shikokuchuo: Added in [300c16e](https://github.com/tidyverse/purrr/pull/1163/commits/300c16ea947a4c1122c730fe8e20fbc79b637555). The rationale is also explained [a bit further down](https://github.com/tidyverse/purrr/pull/1163/files#diff-446777212a6795a48f6f10afd28f3789e7adb07fe34ed819a21f853ae890e1b1R74-R78).

R/parallelization.R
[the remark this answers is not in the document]
shikokuchuo: Just for the record - we solved this one through auto-crating.

DESCRIPTION
shikokuchuo: Yes, it does complicate the syntax a little bit as we can't re-export `daemons()` and `crate()`, but it's fine to start there.

R/parallelization.R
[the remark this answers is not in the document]
hadley: This isn't a problem right now, but we need something to protect you calling `purrr::map(1:100, pkg::foo, .parallel = TRUE)` where unbeknownst to you, `pkg:foo()` is creating 10 daemons.

`mirai::daemons(1e4)` felt more than a regular OOM problem because I got the spinning beach ball of death.

R/parallelization.R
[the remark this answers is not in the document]
hadley: For now, you might want to reinforce that you should never call `daemons()` inside of a package.

R/map.R
```diff
@@ -225,6 +293,60 @@ with_indexed_errors <- function(expr, i, names = NULL, error_call = caller_env()
...
+      )
+    },
+    interrupt = function(cnd) {
+      interrupt_expr
+    }
+  )
+}
+
+ensure_parallel_dependencies <- function(error_call = NULL) {
+
```
hadley, line 321: Is the a reason to not use `rlang::check_installed()` here?

tests/testthat/test-parallel.R
```diff
@@ -0,0 +1,381 @@
...
+
+test_that("negative .at omits locations", {
+  x <- c(1, 2, 3)
+  out <- map_at(x, -1, ~ .x * 2, .parallel = TRUE)
+  expect_equal(out, list(1, 4, 6))
+})
+
+# ---------------------------------------------------------------------------
+
+mirai::daemons(0)
```
hadley, line 381: This shouldn't be necessary, right? (Because of the `on.exit()` call above. Or is just just belt-and-braces for someone running the file line-by-line?

shikokuchuo:
@hadley I've been through your review comments. Notably, I've moved mirai and carrier to suggests in [e433035](https://github.com/tidyverse/purrr/pull/1163/commits/e4330353259fd6942f14914351cdfd46169e5a61) and used cli links to make them easier to install in [ac32e66](https://github.com/tidyverse/purrr/pull/1163/commits/ac32e6623a960ce28553df0c6fb76483126b45b1).

I found that due to limitations in cli, there's seemingly no way to merge the 2 installation commands into one. But I guess that's fairly minor!

I've also re-worded the 'crating' section of the parallelization docs in light of auto-crating.

There are further mirai features that purrr can take advantage of down the line, but I think it makes for a good starting point what we have now, without over-burdening users.

tests/testthat/test-parallel.R
shikokuchuo: I just want to be extra careful in case the tests are run in parallel etc. `daemons(0)` always returns zero - it's just a no-op if there's nothing set up.

R/map.R
shikokuchuo: Hmm good point, let me switch.

shikokuchuo: Done in [304af59](https://github.com/tidyverse/purrr/pull/1163/commits/304af592b4696a2669d22cd696ea52a9b62b0aad). Saves a lot of boilerplate!

R/parallelization.R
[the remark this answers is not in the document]
shikokuchuo: > For now, you might want to reinforce that you should never call `daemons()` inside of a package.

I've added a point in [ea5fc35](https://github.com/tidyverse/purrr/pull/1163/commits/ea5fc356bdccade10b0195ea886bd6a9deab8ec5) to this section of the docs.

R/parallelization.R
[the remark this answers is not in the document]
shikokuchuo: > This isn't a problem right now, but we need something to protect you calling `purrr::map(1:100, pkg::foo, .parallel = TRUE)` where unbeknownst to you, `pkg:foo()` is creating 10 daemons.

Btw. your hypothetical would only work if the daemons were being launched from different machines - so it wouldn't cause a situation where you end up with 1,000 daemons on the _same_ machine.

If you try to revise `daemons()` settings on the same machine without an explicit reset, you'd just error on a subsequent attempt.

R/map.R
shikokuchuo: Changed to still cache the result in [4b4066a](https://github.com/tidyverse/purrr/pull/1163/commits/4b4066ac67eb0116eb0b802f2f83abb5aa82a69c) as `rlang::check_installed()` is rather expensive.

shikokuchuo:
Final changes my side, I'll make a cut of mirai once this is merged.

R/parallelization.R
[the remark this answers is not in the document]
hadley: I think we're good for now if we tell people not to set daemons in a package, but it's something we'll need to come back to again in the future.

PR #1170, opened 2025-02-07
Use CRAN mirai

shikokuchuo:
mirai 2.1.0 which fully supports the new purrr parallelization features is already on CRAN.

Will leave this as a draft until the binary builds propagate.


shikokuchuo:
As mirai itself doesn't contain any compiled code, it's actually fine to merge now. Thanks @hadley!

hadley:
Thank you!

PR #1181, opened 2025-05-24
Updates post-mirai 2.3.0

shikokuchuo:
Minor updates:
- Uses the new `mirai::daemons_set()`
- Updates mirai URLs in docs to 'r-lib'
- No great need to cache installed packages after we solved https://github.com/r-lib/rlang/issues/1776

### DIFF
--- a/R/map.R
+++ b/R/map.R
@@ -211,7 +211,7 @@ map_ <- function(.type,
 
 mmap_ <- function(.x, .f, .progress, .type, error_call, ...) {
 
-  rlang::check_installed(c("mirai", "carrier"), reason = "for parallel map.")
+  check_installed(c("mirai", "carrier"), reason = "for parallel map.")
 
   if (!mirai::daemons_set()) {
     cli::cli_abort(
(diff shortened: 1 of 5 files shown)


R/map.R
```diff
@@ -211,12 +211,9 @@ map_ <- function(.type,
 
 mmap_ <- function(.x, .f, .progress, .type, error_call, ...) {
 
-  if (is.null(the$packages_installed)) {
-    rlang::check_installed(c("mirai", "carrier"), reason = "for parallel map.")
-    the$packages_installed <- TRUE
-  }
+  check_installed(c("mirai", "carrier"), reason = "for parallel map.")
```
hadley, line 214: Do you want to check the versions here too? The version dependencies in suggests are not enforced by `install.packages()`.

hadley (approved):

R/map.R
shikokuchuo: Good point - added version checks in [afccb43](https://github.com/tidyverse/purrr/pull/1181/commits/afccb439873a2d333f7e7cc86a074580b7d4c766). And because of that moved back to caching the result as we're no longer on the fast path for `rlang::check_installed()`.

PR #1184, opened 2025-06-04
Auto-crate anonymous functions

shikokuchuo:
Proof of concept, so that the below 'just works':

```r
library(purrr)
mirai::daemons(3)

slow_lm <- function(formula, data) {
  Sys.sleep(1)
  lm(formula, data = data)
}

models <- mtcars |>
  split(mtcars$cyl) |>
  map(\(df) slow_lm(mpg ~ disp, data = df), .parallel = TRUE)

mirai::daemons(0)
```


shikokuchuo:
Superseded by #1185.

PR #1185, opened 2025-06-09
Update Parallelization to eliminate the `.parallel` argument

shikokuchuo:
This PR updates the existing parallel implementation by replacing the `.parallel` argument of `map()` functions with the use of `parallelize()` as an alias of `carrier::crate()`.

By requiring `parallelize()` as _the_ way to enable parallel operation, we no longer need to worry about which functions need to be crated or not, or when to perform auto-crating. This also has the added benefit of forcing the use of lambda functions (or formulae), i.e.
```r
map(1:3, stats::rnorm, .parallel = TRUE)
map(1:3,\(x) stats::rnorm(x), .parallel = TRUE)

# Now all become:
map(1:3, parallelize(\(x) stats::rnorm(x)))
```

- The test suite has been updated, and the only changes recorded in the snapshot tests are due to the changes in the calls themselves.
- The main documentation is now included under the definition for `parallelize()`.

Also as discussed, perhaps we want to use the adverb `in_parallel()` rather than `parallelize()` to be semantically clearer. So the above example would become:

```r
map(1:3, in_parallel(\(x) stats::rnorm(x)))
```
I'm happy to make this change if we're in agreeement - just let me know.

cc. @hadley @DavisVaughan @lionel-

### DIFF
--- a/R/parallelization.R
+++ b/R/parallelization.R
@@ -14,8 +14,8 @@
 #'  * It helps you create self-contained functions that are isolated from your
 #'    workspace. This is important because the function is packaged up
 #'    (serialized) to be sent across to parallel processes. Isolation is
-#'    critical for performance because it prevents sending very large objects
-#'    without the intention.
+#'    critical for performance because it prevents accidentally sending very large
+#'    objects between processes.
 #'
 #' @param .f A fresh formula or function. "Fresh" here means that they should be
 #'   declared in the call to [in_parallel()].


R/parallelization.R
```diff
@@ -52,58 +102,19 @@
...
 #' \pkg{purrr}'s parallelization is powered by \CRANpkg{mirai}. See the
 #' [mirai website](https://mirai.r-lib.org/) for more details.
 #'
-#' Crating is provided by the \CRANpkg{carrier} package. See the
-#' [carrier readme](https://github.com/r-lib/carrier) for more details.
-#'
-#' @name parallelization
-NULL
+#' @export
+parallelize <- carrier::crate
```
hadley, line 120:
```suggestion
parallelize <- fucntion(.fn, ...) {
   check_installed(c("carrier", "mirai"))   
   carrier::crate(.fn. ...)
}
```

There are a couple of advantages: most importantly this avoids literally copying the crate function in at build-time (which 99% of the time is fine, but when it's not causes extremely confusing bugs), but also allows you to move carrier back to suggests (which I think make sense because carrier and mirai should have the same status). 

It's not strictly necessary to check for mirai here but I think it gives a better user experience.

R/map-if-at.R
```diff
@@ -20,8 +20,9 @@
 #'   `TRUE` will be modified.
 #' @param .else A function applied to elements of `.x` for which `.p`
 #' returns `FALSE`.
-#' @export
+#' @inheritSection map Parallelization
```
hadley, line 23: Maybe this could now just be a link in the documentation for `.f`?

R/map.R
```diff
@@ -137,54 +139,53 @@
...
 #'
-#' mtcars |> map_dbl(sum, .parallel = TRUE)
+#' mtcars |> map_dbl(parallelize(\(x) sum(x)))
 #'
 #' 1:10 |>
-#'   map(function(x) stats::rnorm(10, mean = x), .parallel = TRUE) |>
-#'   map_dbl(mean, .parallel = TRUE)
+#'   map(parallelize(\(x) stats::rnorm(10, mean = x))) |>
+#'   map_dbl(parallelize(\(x) mean(x)))
 #'
```
hadley, line 153: Can you please include an example where you include some other state from the current environment?

R/map.R
```diff
@@ -193,7 +194,7 @@ map_ <- function(.type,
 
   .f <- as_mapper(.f, ...)
 
-  if (isTRUE(.parallel)) {
+  if (is_crate(.f)) {
```
hadley, line 197: Maybe move above `as_mapper` now?

R/map.R
```diff
@@ -212,11 +213,7 @@ map_ <- function(.type,
 mmap_ <- function(.x, .f, .progress, .type, error_call, ...) {
 
   if (is.null(the$packages_installed)) {
```
hadley, line 215: Maybe all this logic could move to `in_parallel()`?

R/map.R
```diff
@@ -228,18 +225,11 @@ mmap_ <- function(.x, .f, .progress, .type, error_call, ...) {
   }
   if (...length()) {
     cli::cli_abort(
-      "Don't use `...` with `.parallel = TRUE`.",
+      "Don't use `...` with parallelized functions.",
```
hadley, line 228:
```suggestion
      "Can't use `...` with parallelized functions.",
```

https://style.tidyverse.org/errors.html#problem-statement

R/parallelization.R
hadley: I also wonder if it's still worth printing the size of the crated function?

R/map.R
shikokuchuo: Yes, that should be moved up, done in [626fa80](https://github.com/tidyverse/purrr/pull/1185/commits/626fa801c7e8836bf310395910b21a743a6f7283).

R/map.R
shikokuchuo: Have added our infamous `slow_lm()` example in [5f023e6](https://github.com/tidyverse/purrr/pull/1185/commits/5f023e6dcc7f046080e6c1c4fdbb162acca510ea). 😄

R/map-if-at.R
shikokuchuo: Agree that this is clearer, and amended in [c47a161](https://github.com/tidyverse/purrr/pull/1185/commits/c47a16146f5ba981aa8b5c3900e3d2364206c9fa).

R/parallelization.R
shikokuchuo: > There are a couple of advantages: most importantly this avoids literally copying the crate function in at build-time (which 99% of the time is fine, but when it's not causes extremely confusing bugs), but also allows you to move carrier back to suggests (which I think make sense because carrier and mirai should have the same status).

I'm aware of this, and I did initially try this way.

However `carrier::crate()` is a bit special in this case as it requires a freshly-defined function. All my attempts at getting around this have caused at least one failure in the tests.

My suggestion - either leave it as is, or we vendor the function into purrr. I've tested that vendoring works, and the definition's fairly minimal. The advantage of this is that we can also tailor the error messages to mention `in_parallel()` rather than `crate()`.

Inviting @lionel-'s view. I think either way works, as it's a very limited scope package where we wouldn't be making changes frequently, if at all.

lionel-: I guess we would want `enexpr()` instead of `substitute()` so callers can inject stuff in the crated function? Though implicit injection will cause surprising behaviour if the parallel function uses tidyeval. It's likely better to require explicit injection.

shikokuchuo: > This should work:
> 
> ```r
> inject(carrier::crate(!!substitute(fn), ...)
> ```
> 
> I've opened [r-lib/carrier#14](https://github.com/r-lib/carrier/issues/14) regarding error messages.

This still causes one test to fail: https://github.com/tidyverse/purrr/blob/main/tests/testthat/test-parallel.R#L101-L114. Same using `enexpr()`.

With the error:
```r
Error (test-parallel.R:105:3): primitive dispatch correctly
Error: object 'method' not found
Backtrace:
     ▆
  1. ├─testthat::expect_identical(...) at test-parallel.R:105:3
  2. │ └─testthat::quasi_label(enquo(object), label, arg = "object")
  3. │   └─rlang::eval_bare(expr, quo_get_env(quo))
  4. ├─purrr::map(...)
  5. │ └─purrr:::map_("list", .x, .f, ..., .progress = .progress) at purrr/R/map.R:164:3
  6. │   └─carrier::is_crate(.f) at purrr/R/map.R:202:3
  7. └─purrr::in_parallel(function(x) as.character(x), as.character.test_class = method)
  8.   ├─rlang::inject(carrier::crate(!!substitute(.f), ...), env = environment()) at purrr/R/parallelization.R:121:3
  9.   └─carrier::crate(function(x) as.character(x), ...)
 10.     └─carrier:::locally(!!!dots, .env = env)
 11.       └─rlang::eval_bare(dots[[i]], .env)
```

shikokuchuo: Oh I think I found the solution:
```r
inject(carrier::crate(!!substitute(.f), !!!list(...)))
```
Thanks again!

R/map.R
shikokuchuo: Moved in [8b48804](https://github.com/tidyverse/purrr/pull/1185/commits/8b4880480ba87fb05963229bfff2754b2d05fa4a). Note there's an outside chance someone could pass a crate created by `carrier::crate()` outside of purrr, and not have mirai installed. But I don't think it's worth worrying about.

R/parallelization.R
shikokuchuo: @hadley's suggestions have now been implemented in [8b48804](https://github.com/tidyverse/purrr/pull/1185/commits/8b4880480ba87fb05963229bfff2754b2d05fa4a).

shikokuchuo:
Thanks all for making this implementation all that much better!

Especially:
@lionel- for persisting with your instinct that crating should be the signal for parallel execution.
@DavisVaughan for pinpointing that we needed to 'skin' `carrier::crate()`.
@hadley for being on top of everything!

I'll merge this in the next couple of days unless I receive further comments.

R/parallelization.R
```diff
@@ -1,9 +1,62 @@
...
+#' mtcars |> map_dbl(in_parallel(sum))
+#' # Use an anonymous function:
+#' mtcars |> map_dbl(in_parallel(\(x) sum(x)))
+#'
+#' # Package functions need to be explicitly namespaced, so instead of:
+#' map(1:3, in_parallel(\(x) runif(x)))
+#' # Use :: to namespace all packages, even those on the default search path:
+#' map(1:3, in_parallel(\(x) stats::runif(x)))
+#'
+#' fun <- \(x) \{x + x \%\% 2 \}
```
DavisVaughan:
```suggestion
#' fun <- function(x) \{x + x \%\% 2 \}
```

R/parallelization.R
```diff
@@ -1,9 +1,62 @@
...
+#'   with its namespace prefix: `stats::var(x)`.
+#'
+#' * They should declare any data they depend on. You can declare data by
+#'   supplying additional named arguments to `...`⁠.
+#'
+#' [in_parallel()] is a simple wrapper of [carrier::crate()] and you may refer
+#' to that package for more details.
+#'
+#' Example usage:
+#' \preformatted{
```
DavisVaughan: You can also use an R markdown block like

``````
```r
``````

And then you don't need to escape the braces, like `\{`

**Review by DavisVaughan — APPROVED** [review]

File: R/parallelization.R [hunk 9]
```diff
@@ -1,9 +1,64 @@
...
+#' @param .f A fresh formula or function. "Fresh" here means that they should be
+#'   declared in the call to [in_parallel()].
+#' @param ... Arguments to declare in the environment of the function.
+#'
+#' @section How to Use:
+#'
+#' Wrapping a function that is passed to the `.f` argument in [map()] or any of
+#' its variants with [in_parallel()] declares that the map should be performed
+#' in parallel.
+#'
```
lionel-, line 21: Suggestion: Structure the doc along these lines:


@description
`in_parallel()` is a purrr adverb that plays two roles:

- It is a _signal_ to purrr verbs like [map()] to go ahead and perform computations in parallel.
- It helps you create _self-contained_ functions that are isolated from your workspace. This is important because the function is packaged up (in technical terms serialized) to be sent over to worker processes. Isolation is critical for performance because it prevents sending very large objects unexpectedly.

@section Creating self-contained functions:

R/parallelization.R
```diff
@@ -1,9 +1,64 @@
...
+#'   with its namespace prefix: `stats::var(x)`.
+#'
+#' * They should declare any data they depend on. You can declare data by
+#'   supplying additional named arguments to `...`⁠.
+#'
+#' [in_parallel()] is a simple wrapper of [carrier::crate()] and you may refer
+#' to that package for more details.
+#'
+#' Example usage:
+#' ```r
```
lionel-, line 43: Maybe show the return value of `in_parallel()` and how the size varies as a function of the data that is passed along?

And how the function can be tested on its own to check that every dependency has been packaged up?

R/parallelization.R
```diff
@@ -1,9 +1,64 @@
...
+#' in parallel.
+#'
+#' Under the hood, [in_parallel()] provides a systematic way of making `.f`
+#' self-contained so that it can be readily shared with other parallel
+#' processes. It ensures that everything needed by the function is serialized
+#' along with it, but not other objects which happen to be in the function's
+#' enclosing environment. This helps to prevent inadvertently shipping large
+#' data objects when they may not be needed.
+#'
+#' To create self-contained functions:
```
lionel-, line 29: This requires a little work because you'll need to:

- Pass along dependencies, such as a list of parameters or metadata, to be packaged up in your self-contained function.
- Fully qualify calls to non-base functions with their namespaces.

R/parallelization.R
```diff
@@ -13,97 +68,106 @@
...
-#'
-#' Crating is provided by the \CRANpkg{carrier} package. See the
-#' [carrier readme](https://github.com/r-lib/carrier) for more details.
+#' # Tear down daemons when no longer in use:
+#' mirai::daemons(0)
 #'
-#' @name parallelization
-NULL
+in_parallel <- function(.f, ...) {
+  if (is.null(the$packages_installed)) {
```
lionel-, line 160: `the$in_parallel_checked`? `packages_installed` makes it sound like the scope of that variable is bigger than just `in_parallel()`.

R/parallelization.R
```diff
@@ -13,97 +68,106 @@
...
+    check_installed(
+      c("carrier", "mirai"),
```
lionel-, line 162: Do you also check for mirai down the line? This might be necessary if the user saved a parallel function in an rds and loads it back in a fresh session.

R/parallelization.R
```diff
@@ -13,97 +68,106 @@
...
+      version = c("0.1.1", "2.3.0"),
+      reason = "for parallel map."
+    )
+    the$packages_installed <- TRUE
+  }
+  inject(carrier::crate(!!substitute(.f), !!!list(...)))
+}
+
+is_crate <- function(x) {
+  inherits(x, "crate")
```
lionel-, line 172: I must have created carrier before we started to systematically namespace S3 classes with package names...

lionel- (approved):

R/map.R
```diff
@@ -31,6 +31,12 @@
 #'     `\(x) pluck(x, "idx", 1)` respectively. Optionally supply `.default` to
 #'     set a default value if the indexed element is `NULL` or does not exist.
 #'
+#'   `r lifecycle::badge("experimental")`
+#'
+#'   Wrap a function with [in_parallel()] to declare that it should proceed
+#'   in parallel. See [in_parallel()] for more details.
+#'   * Use of `...` is not permitted in this context.
```
hadley, line 38:
```suggestion
#'   Use of `...` is not permitted in this context.
```

R/map.R
```diff
@@ -140,63 +135,77 @@
...
-#' # To use parallelized map, set daemons (number of parallel processes) first:
+#' # To use parallelized map:
+#' # 1. Set daemons (number of parallel processes) first:
 #' mirai::daemons(2)
 #'
-#' mtcars |> map_dbl(sum, .parallel = TRUE)
+#' # 2. Wrap .f with in_parallel():
+#' mtcars |> map_dbl(in_parallel(\(x) sum(x)))
 #'
+#' # Note that functions from packages should be fully qualified with `pkg::`
```
hadley, line 145: Or load with `library()`? (Or is that not ready yet?)

R/parallelization.R
```diff
@@ -13,97 +68,99 @@
 #' according to your individual setup and task, but a rough guide would be in
 #' the order of 100 microseconds to 1 millisecond for each map iteration.
 #'
-#' # Daemons settings
+#' @section Daemons Settings:
```
hadley, line 66: Out of interest, why did you change these?

R/parallelization.R
```diff
@@ -3,30 +3,25 @@
...
-#' processes. It ensures that everything needed by the function is serialized
-#' along with it, but not other objects which happen to be in the function's
-#' enclosing environment. This helps to prevent inadvertently shipping large
-#' data objects when they may not be needed.
+#' @param .f A fresh formula or function. "Fresh" here means that they should be
+#'   declared in the call to [in_parallel()].
+#' @param ... Named arguments to declare in the environment of the function.
 #'
-#' To create self-contained functions:
+#' @section Creating Self-contained Functions:
```
hadley, line 24:
```suggestion
#' @section Creating self-contained functions:
```

We are team sentence case 😄

R/parallelization.R
shikokuchuo: Are you referring to the case of the section title? I wasn't aware that we were team sentence case 😄

R/map.R
shikokuchuo: Will update once PRs at carrier are merged.

R/parallelization.R
hadley: Here I meant from `#` to `@section`

shikokuchuo: Just to make it consistent with other the other usage in the same function. I probably copied the # over originally from the progress bars docs.

NAMESPACE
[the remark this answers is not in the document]
shikokuchuo: Now implemented in carrier.

R/map.R
shikokuchuo: Now updated as part of [ccf95f3](https://github.com/tidyverse/purrr/pull/1185/commits/ccf95f37a9abb8774f96ef5c2357bb2e5245a843).

shikokuchuo:
All updates from carrier now incorporated with https://github.com/r-lib/carrier/pull/21 merged. Thanks all!

PR #1194, opened 2025-06-27
Deprec advance 1.0.0

hadley:


R/coerce.R
```diff
@@ -7,22 +7,7 @@ coerce <- function(x, type) {
...
-  lifecycle::deprecate_warn(
-    "1.0.0",
-    I(paste0("Automatic coercion from ", type, " to character")),
-    I("an explicit call to `as.character()` within `map_chr()`"),
-    always = TRUE,
-    user_env = the$deprecation_user_env
-  )
-}
-
+coerce_chr <- function(x) coerce(x, "character")
```
lionel-, line 10: I think it would be worthwile using `depecrate_stop()` here as the error message is more friendly and provides a suggested fix. The warnings might have flown under the radar.

lionel- (approved):

R/coerce.R
hadley: I'll take a look at the revdep failures and see if it's causing much pain.

hadley: There were only four failures (out of ~2000 packages) so I think it's just as easy to patch those packages.

PR #1195, opened 2025-06-30
Require R 4.1

### LINKED ISSUE #1177: Drop base pipe hackery
Now that we can depend on R 4.1

### LINKED ISSUE #1180: Documentation: NEWS.md does not mention increased requirement (R >= 4.0)
https://github.com/tidyverse/purrr/blob/9c8beb4e648e64ec3e3ca816df0baeaaf693d1a4/DESCRIPTION#L17

hadley:
The only place it remains is in `test-deprec-prepend.R` since we use a bunch of the magittr pipe features there, and that code will soon go away.

Fixes #1177. Fixes #1180.


R/keep.R
```diff
@@ -16,8 +16,7 @@
 #'   * A named function, e.g. `is.character`.
 #'   * An anonymous function, e.g. `\(x) all(x < 0)` or `function(x) all(x < 0)`.
 #'   * A formula, e.g. `~ all(.x < 0)`. You must use `.x` to refer to the first
-#'     argument). Only recommended if you require backward compatibility with
-#'     older versions of R.
+#'     argument). No longer recommended.
```
lionel-, line 19: Here and below, I think it'd be helpful to be explicit about  why this is no longer recommended, e.g.:

```suggestion
#'     argument). No longer recommended as we now embrace the base R `\(x)` syntax added in R 4.1.
```

lionel- (approved):

PR #1217, opened 2025-09-19
modify error message to show .x argument name for one or more inputs

### LINKED ISSUE #1151: Names in indexed errors in `map2()` can be misleading
The recently introduced indexed error messages are super helpful, many thanks! However, I’ve spent some time head-scratching recently because I had some ill-sorted named lists that I wanted to merge using `dplyr`'s `bind_rows()` and the error message sort of navigated me to compare list elements by name, not by index.

Maybe in case of `map2()`, can we have a note that the name of list element refers only to the first list? Or even better, names of both lists could be provided. If both names are equal at the given index, it can be left as it is currently implemented.

What do you think about that?

``` r
purrr::map2(
  list(a = 1),
  list(b = 1),
  \(x, y) stop()
  )
#> Error in `purrr::map2()`:
#> ℹ In index: 1.
#> ℹ With name: a.
#> Caused by error in `.f()`:
```

<sup>Created on 2024-09-11 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

wahalulu:
fix #1151

### DIFF
--- a/R/map.R
+++ b/R/map.R
@@ -281,10 +281,10 @@ with_indexed_errors <- function(
       if (i == 0L) {
         # Error happened before or after loop
       } else {
-        message <- c(i = "In index: {i}.")
+        message <- c(i = "In .x[[{i}]].")
         if (!is.null(names) && !is.na(names[[i]]) && names[[i]] != "") {
           name <- names[[i]]
-          message <- c(message, i = "With name: {name}.")
+          message <- paste(message, "with element name: {name}.", sep = " ")
         } else {
           name <- NULL
         }
--- a/tests/testthat/_snaps/keep.md
+++ b/tests/testthat/_snaps/keep.md
@@ -4,14 +4,14 @@
       keep(1:3, ~NA)
     Condition
       Error in `keep()`:
-      i In index: 1.
+      i In .x[[1]].
       Caused by error:
       ! `.p()` must return a single `TRUE` or `FALSE`, not `NA`.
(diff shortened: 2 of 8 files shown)


R/map.R
```diff
@@ -281,10 +281,10 @@ with_indexed_errors <- function(
       if (i == 0L) {
         # Error happened before or after loop
       } else {
-        message <- c(i = "In index: {i}.")
+        message <- c(i = "In .x[[{i}]].")
         if (!is.null(names) && !is.na(names[[i]]) && names[[i]] != "") {
           name <- names[[i]]
-          message <- c(message, i = "With name: {name}.")
+          message <- paste(message, "with element name: {name}.", sep = " ")
```
lionel-, line 287: I think @hadley suggested this should be `/ x${name}`

lionel- (commented):
Good start but the name formatting needs a little more work.

R/map.R
wahalulu: got it. fixed in f2f2869

lionel-:
Thanks!

lionel-:
@wahalulu Sorry I missed in my review that this requires adding an `error_arg = ".x"` argument to all mapper functions hitting that code path, as indicated by Hadley in the original issue. Could you please take another look?

lionel-:
Just noting that this is sparking a discussion within the Tidyverse team. Could you come and find us please @wahalulu?

Issue #1205, opened 2025-08-21
Carefully examine all mentions of formula


hadley:
And either remove it, or make it clear that you should only use if you need backward compatibility with old R.

PR #1220, opened 2025-09-19
check mentions of formulas, specifiy backward compatibility

luisDVA:
addresses #1205 

Examples, vignettes, and readme no longer have mentions to formulas except for a more consistent message explaining that these may be used but only for backward compatibility with older versions of R.
* also fixed a typo or two

part of Tidyverse Dev Day 2025, ATL

### DIFF
--- a/R/detect.R
+++ b/R/detect.R
@@ -6,8 +6,8 @@
 #'
 #'   * A named function, e.g. `mean`.
 #'   * An anonymous function, e.g. `\(x) x + 1` or `function(x) x + 1`.
-#'   * A formula, e.g. `~ .x + 1`. You must use `.x` to refer to the first
-#'     argument. No longer recommended.
+#'   A formula, e.g. `~ .x + 1`. Only recommended if you require backward
+#'     compatibility with older versions of R.
 #'   * A string, integer, or list, e.g. `"idx"`, `1`, or `list("idx", 1)` which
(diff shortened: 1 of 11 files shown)


jennybc:
I think it might be good to replace "older versions of R" with something specific, e.g. "R <= 4.1".

hadley:
I think we could probably just stick with the previous language and say "No longer recommended." (Sorry @luisDVA I think I gave you bad advice when we chatted)

R/detect.R
```diff
@@ -6,8 +6,8 @@
 #'
 #'   * A named function, e.g. `mean`.
 #'   * An anonymous function, e.g. `\(x) x + 1` or `function(x) x + 1`.
-#'   * A formula, e.g. `~ .x + 1`. You must use `.x` to refer to the first
-#'     argument. No longer recommended.
+#'   A formula, e.g. `~ .x + 1`. Only recommended if you require backward
+#'     compatibility with older versions of R.
 #'   * A string, integer, or list, e.g. `"idx"`, `1`, or `list("idx", 1)` which
```
hadley, line 11: There's also a mention of "lambda-formula" in the examples that should be removed.

luisDVA:
continuation of the work from dev day.
I reverted the comments about backward compatibility to just state "No longer recommended". The only mention of backward compatibility with R < 4.1 is now in a note at the end of the readme, as suggested by @hadley at the end of dev day. Also deleted a remaining mention of "lambda-formulas"

PR #1224, opened 2025-09-25
Don't attempt to name `NULL`

### LINKED ISSUE #1206: `map_vec()` fails on empty named lists()
Empty lists `map_vec()` to `NULL`, as expected:

```r
map_vec(list(), identity)
#> NULL
```

But empty *named* lists fail (looks like because they try to restore the names on `NULL`):

```r
map_vec(setNames(list(), character()), identity)
#> Error in names(x) <- names : attempt to set an attribute on NULL
```

Expected behavior: named lists should also just `map_vec()` to `NULL`, right?

hadley:
Fixes #1206

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # purrr (development version)
 
+* `map_vec()` no longer fails on empty named lists (#1206).
+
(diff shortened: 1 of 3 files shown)


shikokuchuo (commented):
I'd probably leave this out for now and fix at vctrs, perhaps by adding an early return here: https://github.com/r-lib/vctrs/blob/ca33f4d5c9afb6e44b5373c8f2b196602b6a67aa/src/names.c#L849
Of course, defer to Davis.

PR #1238, opened 2025-11-26
Fix typos in NEWS.md

rcannood:
Not sure how interesting fixing typos in old release notes is, but here we are. I created a separate PR so that #1237 could be merged and #1238, in case there is little interest in this PR.


hadley:
Thanks!

Issue #1245, opened 2026-01-25
Unexpected order of names in `accumulate(.dir = "backward")`


ErdaradunGaztea:
I believe the current name order of the output of `accumulate(.dir = "backward")` applied to a named vector is unexpected. To explain why I think so, let me demonstrate:

What `.dir = "backward"` essentially does in a `reduce()` call is to call `.f` from right to left of the input vector. We can translate that to `rev(.x)` _and_ reversed argument order to the `.f` function, giving us equivalent calls:

``` r
paste_dot <- \(x, y) { paste(x, x, y, sep = ".") }
x <- c("a", "b", "c")

purrr::reduce(x, paste_dot, .dir = "backward")
#> [1] "a.a.b.b.c"
purrr::reduce(rev(x), \(.x, .y) { paste_dot(.y, .x) })
#> [1] "a.a.b.b.c"
```

Since `accumulate()` stores all the intermediate outputs, we need to add one more `rev()` on its output (mathematically speaking, the same is true for `reduce()`, but since there's only one value returned from there, it doesn't make a difference):

```r
purrr::accumulate(x, paste_dot, .dir = "backward")
#> [1] "a.a.b.b.c" "b.b.c"     "c"
purrr::accumulate(rev(x), \(.x, .y) { paste_dot(.y, .x) }) |>
  rev()
#> [1] "a.a.b.b.c" "b.b.c"     "c"
```

Now, each name in the output of `accumulate()` should preferably correspond to the name of the corresponding input element. This works just fine in the `.dir = "forward"` case:

``` r
y <- c(A = "a", B = "b", C = "c")

purrr::accumulate(y, paste_dot)
#>               A               B               C 
#>             "a"         "a.a.b" "a.a.b.a.a.b.c"
```

However, this breaks for `.dir = "backward". If we compare it to the output of the reversed `accumulate()` (according to the equivalence established above), the order of the values is correct, but the names do not match:

```r
purrr::accumulate(y, paste_dot, .dir = "backward")
#>           C           B           A 
#> "a.a.b.b.c"     "b.b.c"         "c"
purrr::accumulate(rev(y), \(.x, .y) { paste_dot(.y, .x) }) |>
  rev()
#>           A           B           C 
#> "a.a.b.b.c"     "b.b.c"         "c"
```

[...]

PR #1246, opened 2026-01-28
Rewrite `reduce()` and family in C

### LINKED ISSUE #1168: Progress bar to reduce
Would it be possible to add `.progress` as an argument to `reduce` (and relatives, e.g., `accumulate` and `reduce2`)?
Some times you may use `reduce` in stead of `map_dfr` to achieve similar output, but the former doesn't provide options for progress bar.

### LINKED ISSUE #1243: `accumulate()` fails when input has names and using early return
`accumulate()` fails if `.x` has names and an early return is used through a `done()` call.

``` r
#" This works just fine
1:4 |>
  purrr::accumulate(\(x, y) { if (y < 3) x + y else rlang::done() }, .init = 0)
#> [1] 0 1 3
```

Adding names makes it fail

``` r
list(a = 1, b = 2, c = 3, d = 4) |>
  purrr::accumulate(\(x, y) { if (y < 3) x + y else rlang::done() }, .init = 0)
#> Error in `names(res) <- accumulate_names(names(.x), .init, .dir)`:
#> ! 'names' attribute [5] must be the same length as the vector [3]
```

<sup>Created on 2026-01-18 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

The reason for that is that
```
# line 482
names(res) <- accumulate_names(names(.x), .init, .dir)
```
doesn't take early return possibility into account and always tries to assign full-length names vector to the result.

I already work on another issue related to `reduce()` and will solve this one as well, but in case I don't follow with a pull request this month (January) I wanted to notify you of that.

ErdaradunGaztea:
Closes #1168 and closes #1243. Your acceptation of #1169 prompted me to tackle a bigger task 🦔 

This PR reimplements `reduce()` (and `reduce2()`, `accumulate()`, and `accumulate2()` as well) in C. The idea came from Hadley's comment on #1168 stating that it'd be hard to performantly add progress bars to `reduce()` as they are written in C, so I rewrote the function in C instead. It is more performant too, getting `reduce()` to be on par with base `Reduce()`:

``` r
x <- as.list(1:20000)
# Without .init or .progress
bench::mark(
  purrr::reduce_old(x, sum),
  purrr::reduce(x, sum),
  Reduce(sum, x),
  min_time = 1
)
#> # A tibble: 3 × 6
#>   expression                     min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>                <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 purrr::reduce_old(x, sum)   15.1ms  15.37ms      61.9     762KB     81.8
#> 2 purrr::reduce(x, sum)       8.08ms   8.35ms     118.      170KB     73.6
#> 3 Reduce(sum, x)              7.72ms   8.05ms     124.      349KB     51.6
```

<sup>Created on 2026-01-18 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

`accumulate()` isn't quite as fast as in base R, but there's improvement nonetheless:

``` r
x <- as.list(1:20000)
bench::mark(
  purrr::accumulate_old(x, sum, .init = 0),
  purrr::accumulate(x, sum, .init = 0),
  Reduce(sum, x, init = 0, accumulate = TRUE),
  min_time = 1
)
#> # A tibble: 2 × 6
#>   expression                            min  median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>                        <bch:t> <bch:t>     <dbl> <bch:byt>    <dbl>
#> 1 purrr::accumulate_old(x, sum, .i… 26.2ms  27.3ms       36.8    1.45MB    178. 
#> 2 purrr::accumulate(x, sum, .init … 12.92ms 14.14ms      69.7    1.47MB    103. 
#> 3 Reduce(sum, x, init = 0, accumul…  9.33ms  9.86ms     100.   505.53KB     73.1
```

<sup>Created on 2026-01-28 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

I tried not to change any behavior (except for #1243) and hopefully succeeded, though that meant adding a lot of tests in the end. That was also the reason I didn't tackle #1245, as it'd change non-erroring behavior.

As for the code, well... It's not pretty, but I couldn't think of any approach that would look cleaner with interacting `.init`, `.dir`, and `done()`, and trying to make one shared implementation for all four functions with only some differences in certain points of the algorithm. If you prefer, I can separate C loops for `reduce()` and `accumulate()` family, that would probably be the most influential one. If I were to say, it's probably 15-20% more complex than the current R implementation (numbers completely made-up).

Finally, the C implementation should make it very easy to add `.dir` parameter to `reduce2()` and `accumulate2()`, as well as make `accumulate2()` return a named vector if input is named, like `accumulate()` does. Then again, I didn't want to add more changes than necessary and I can always add it later if you so wish.

No worries, I'm quickly running out of purrr functions to optimize :)


ErdaradunGaztea:
~Note that my last commit was necessitated by the latest vctrs (I had v0.6.5 installed) release adding extra information to the errors. If you'd rather snatch these updated snaps out of this PR because they do not fit the theme here - you're welcome.~

PR #1255, opened 2026-03-20
Speed up `map()` and `map2()` somewhat (~20%?)

ErdaradunGaztea:
Hi, I'd like to propose a way to speed `map()` and `map2()` up by... well, that's difficult to tell, but 20% seems to be a solid guess.

So, I've noticed that `pluck()` has a very interesting implementation of `extract_vector()` where it'd first try to extract the vector element using C API and only dispatch to `[[` when the variable is an object (with classes). Since `map()` already drops classes from lists unless they are vectors as understood by {vctrs} (so very rarely), I figured it'd make sense to split `map()` implementation into two, keeping current behaviour for objects, but optimizing it akin to `pluck()` for everything else.

Currently, `map()` executes a call of the form `.f(.x[[i]], ...)` (or `.f(.x[[i]], .y[[i]], ...)`, but I'm gonna skip `map2()` for now) in each iteration. It'd be better if we could pass `i`-th element of `.x` to `.f()` instead, i.e. `.f(value, ...)`, where `value` was extracted in C. This would, however, require re-creating a call each time - though that was still faster overall - and makes the code stray further from the original implementation that we still keep for the classed object case. And I wanted to add as little code as possible.

So, I've decided to add an `.x_i` object on R side and make C assign to it upon calculating `value`. Now it's a simple `.f(.x_i, ...)` call that can be defined once, and the logic is very similar. There's an extra call to an `r_extractor` object that is also a new parameter to the mapping loop. Oh, and `x` and `y` are also passed to C now, as we need to extract from them now. That doesn't seem to be much of a cost, though.

What else? I've extracted call creation to a separate file, `calls.c`, because there are four calls now instead of two for `map()` and `map2()`. I've skipped one test case that tested internal implementation too much but added a new one too.

Also, a benchmark:

(Using base R `lapply()` and `mapply()` as benchmark, as `cross::bench_branches()` doesn't work with `reprex()`; their performance is pretty much equal to that of purrr v1.2.1, though)

``` r
x <- 0:1e5
bench::mark(
  purrr = purrr::map(x, \(x) sqrt(x)),
  base = lapply(x, \(x) sqrt(x)),
  iterations = 20
)
#> Warning: Some expressions had a GC in every iteration; so filtering is
#> disabled.
#> # A tibble: 2 × 6
#>   expression      min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr> <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 purrr        40.2ms   50.5ms      18.3     5.8MB     22.9
#> 2 base         53.2ms   78.2ms      11.3   781.3KB     17.5
```

<sup>Created on 2026-03-20 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

``` r
x <- 0:1e5
y <- 1e5:2e5
bench::mark(
  purrr = purrr::map2(x, y, \(x, y) sqrt(y) - sqrt(x)),
  base = mapply(
    \(x, y) sqrt(y) - sqrt(x),
    x, y,
    SIMPLIFY = FALSE, USE.NAMES = FALSE
  ),
  iterations = 15
)
#> Warning: Some expressions had a GC in every iteration; so filtering is
#> disabled.
#> # A tibble: 2 × 6
#>   expression      min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr> <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 purrr        64.2ms   85.1ms     11.1     6.21MB     25.8
#> 2 base         95.1ms  137.2ms      6.98   781.3KB     26.1
```

<sup>Created on 2026-03-20 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

This approach could very well be extended to other iterating functions, like `every()` or maybe `reduce()` from another of my PRs. Or even `pmap()`, because I've decided it looks too complicated to try it now, but it could save a lot of time potentially, as it takes more arguments. I also decided that `map2()` should use faster route only when both `x` and `y` are vectors, but it can be extended to cover mixed cases where only one is an object, too.


ErdaradunGaztea:
Superseded by #1263

PR #1259, opened 2026-04-17
Document with dev roxygen2

hadley:
Looks like all the changes are link related and look correct to me.


man/list_c.Rd
```diff
@@ -42,9 +42,9 @@ the elements will be used instead of the names.}
 \item \code{list_c()} combines elements into a vector by concatenating them together
 with \code{\link[vctrs:vec_c]{vctrs::vec_c()}}.
 \item \code{list_rbind()} combines elements into a data frame by row-binding them
-together with \code{\link[vctrs:vec_bind]{vctrs::vec_rbind()}}.
+together with \code{\link[vctrs:vec_rbind]{vctrs::vec_rbind()}}.
```
lionel-, line 45: This one looks more specific, pointing to the alias rather than the doc nam (still looks correct, just noting).
There's a few others like that.

man/purrr-package.Rd
```diff
@@ -24,6 +24,7 @@ Useful links:
 
 Authors:
 \itemize{
+  \item Hadley Wickham \email{hadley@posit.co} (\href{https://orcid.org/0000-0003-4757-117X}{ORCID})
```
lionel-, line 27: The maintainer now appears in both `\author{}` and the Authors list. Not sure of the implications.

lionel- (approved):
LG!

man/list_c.Rd
hadley: Yeah, that's what you're supposed to do now.

man/purrr-package.Rd
hadley: Yeah, someone asked for that and it seemed reasonable to me.

PR #1260, opened 2026-05-13
Fix #1243 and #1245: Fix accumulate() names

### LINKED ISSUE #1243: `accumulate()` fails when input has names and using early return

mrv0for0vandeta:


mrv0for0vandeta:
This PR fixes two issues with `accumulate()`:

- #1243: Fixed name handling when using early termination with `.done()` - names are now correctly preserved
- #1245: Fixed name order when using `.dir = "backward"` - names now match the correct elements
Both fixes ensure consistent name handling across all accumulate scenarios.

ErdaradunGaztea:
Hi, I don't think your PR fixes #1245. It doesn't change any relevant code and doesn't modify the existing tests that'd fail if you changed the behaviour. If you believe the current behaviour is the correct one (a valid opinion which I nevertheless don't share), I'd recommend commenting on the issue instead.

Speaking of tests, you might want to cover #1243 with some test cases to show that the fix indeed works. A great starting point can be the example provided in the issue description, if available. In this case, I'd add more tests to cover interactions between different parameters, because the issue description only shows one basic case.

PR #1264, opened 2026-07-01
Speed up the map family

hadley:
## Summary

This reworks the C implementation of the `map()`, `map2()`, and `pmap()` families to remove repeated R evaluation and other per-element overhead while preserving S3 extraction, coercion, progress, interrupt, and indexed-error behavior.

## Changes and motivation

- Pass input vectors directly to the C implementations. Bare vectors are read through cached read-only data pointers; objects still use `[[` so S3/vctrs dispatch is preserved. This removes repeated evaluation of `.x[[i]]` and `.y[[i]]` in R.
- Build one reusable call per map invocation and update its argument cells directly. Symbols and language objects are quoted before insertion so they continue to be passed as data. This avoids per-element environment rebinding and symbol lookup.
- Omit `...` from the reusable call when it is empty. Empty dots previously incurred expansion overhead on every iteration.
- Resolve `.f` once and place the function directly in the reusable call. This avoids resolving the callback for every element.
- Allow mapped arguments to remain lazy. Because calls now contain the actual extracted values rather than references to a changing loop binding, callbacks can safely force only the arguments they use. This particularly helps `map2()` and `pmap()` callbacks that ignore some inputs.
- Skip all cli progress setup and volatile timer polling when `.progress = FALSE`. Explicit progress bars retain the existing path.
- Add direct output paths. List results go straight to `SET_VECTOR_ELT()` without redundant protection, and exact-type atomic results write through a cached output pointer. Results needing coercion retain `set_vector_value()`.
- Construct `pmap()` argument and S3 extraction machinery once per invocation rather than rebuilding nested subset calls per element.
- Add the `VECTOR_PTR_RO()` compatibility definition needed before R 4.5.

The internal call shown by `sys.call()` now contains the resolved function and mapped values. That representation is an implementation detail; the test now checks the meaningful call properties rather than its exact syntax.

## Performance

`bench::mark()` medians for 10,000 elements, compared with purrr 1.2.2 on R 4.5 / Apple arm64:

| Workload | purrr 1.2.2 | This PR | Improvement |
|---|---:|---:|---:|
| `map(list, identity)` | 1.669 ms | 1.003 ms | 40% |
| `map_int(integer, identity)` | 1.810 ms | 1.096 ms | 39% |
| `map_dbl(double, identity)` | 1.813 ms | 1.096 ms | 40% |
| `map2_int(integer, integer, add)` | 3.192 ms | 1.814 ms | 43% |
| two-argument `pmap_int()` | 4.187 ms | 1.856 ms | 56% |
| five-argument `pmap_int()` | 9.293 ms | 3.907 ms | 58% |

The first direct-extraction/call changes account for most of the gain. In isolated comparisons against that faster intermediate implementation, the later changes improve callbacks using only a subset of arguments by roughly 10–17%; callbacks forcing every argument improve by roughly 3–8%.

## Verification

- 1,032 tests pass.
- Existing coverage exercises symbolic inputs, retained closures, S3/vctrs extraction, coercion, dots forwarding, progress bars, interrupts, and indexed errors.
- `git diff --check` passes.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # purrr (development version)
 
+* All `map()` functions now have less overhead, particularly on unclassed vectors. This means that they should run faster when `.f` itself is very fast, like with the common cases of `map(x, names)` or `map_lgl(x, is.null)` (#1263, with an assist from @ErdaradunGaztea).
+
 # purrr 1.2.2
 
 * Fixes for CRAN checks (@ErdaradunGaztea, #1256).
--- a/R/map.R
+++ b/R/map.R
@@ -221,7 +221,16 @@ map_ <- function(
     i = i,
     names = names,
     error_call = .purrr_error_call,
-    call_with_cleanup(map_impl, environment(), .type, .progress, n, names, i)
+    call_with_cleanup(
+      map_impl,
+      .x,
+      environment(),
+      .type,
+      .progress,
+      n,
+      names,
+      i
+    )
   )
 }
 
--- a/R/map2.R
+++ b/R/map2.R
@@ -105,7 +105,17 @@ map2_ <- function(
     i = i,
     names = names,
     error_call = .purrr_error_call,
-    call_with_cleanup(map2_impl, environment(), .type, .progress, n, names, i)
+    call_with_cleanup(
+      map2_impl,
+      .x,
+      .y,
+      environment(),
+      .type,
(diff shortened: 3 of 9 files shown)


hadley:
Superseded by the correctly staged PR against #1263: https://github.com/DavisVaughan/purrr/pull/1

